Issue #1, opened 2014-07-09
Missing some Dominican neumes


olivierberten:
See Salve Regina, for instance...
![salve-op](https://cloud.githubusercontent.com/assets/213117/3530702/3e55bfd6-07a7-11e4-9d56-114f022d145a.png)

henryso:
I'll admit to being confused by this discussion.  Which shapes, exactly are missing?  We have the epiphohus, so is it just the "u"-shaped note over "ten" and the inverted "u"-shaped noted over "Vir" that are missing?

eroux:
Maybe there should be an alternative epiphonus shape that looks more like this one, but I think the new thing is the weird neume above "ten" (we don't see the text because of the red circle)... At first I thought it was like an alternative porrectus deminutus but without the first and last note... but can these bars above the note get higher than this or are they always at the same height?

henryso:
What I'd like to understand is how much variety there is in the new shape.  In addition to your question about height, can there be a line only on one side?  A line going up and a line going down?  I don't have any Dominican liturgical books, so I cannot see for myself.  Are there any ones available on the Internet that I can look through?

eroux:
There are somes PDF sources in [this comment](https://github.com/gregorio-project/gregorio/issues/1#issuecomment-52437198) and [on gregobase](http://gregobase.selapa.net/scores.php) (in the *by source* section).

henryso:
From what I've been reading, it seems that the only liquescents Dominican chant uses are the epiphonus, the cephalicus, the epiphonus-line neume without the final note, and the cephalicus-like neume without the final note.  Should the latter two figures without the final notes just be alternate glyphs for the virga (reversa) auctus ascendens -- which would needed to be added -- and virga (reversa) auctus descendens -- which already exists? If this is the case, the notation would be `gV>` and `gV<`.  If the preference is to add these neumes as regular figures (as opposed to alternates), then how about `g&>` and `g&<` for the epiphonus-line and cephalicus-like figures, respectively?

eroux:
But doesn't the bar before the absent note vary in lenght?

henryso:
As far as I can tell, that line height and whether to vary it within a score is a stylistic decision.  Does anyone know for sure?

henryso:
[...]

PR #19, opened 2014-12-31
Fix [ocba:...] not to look like [ocb:...].

jakubjelinek:
Supposedly starting with the addition of 161 to glyph encodings [ocba:1;3mm] looks like [ocb:1;3mm] - the accentus is not printed.
Wonder if
\def\grelinea#1#2#3{%
  \greglyph{\char 87}{#1}{#2}{#3}{}{}%
  \relax %
}
doesn't need similar treatment, but there is nothing in gregorio-chars.tex for it and I don't know how to test it.


eroux:
Many thanx!

Issue #27, opened 2015-01-18
english centering scheme should be available as TeX option


eroux:
having the centering scheme set in gabc was, I believe, as mistake (though, now, it will stay). But it should be possible to set "English" centering scheme from TeX, for scores with a normal gabc centering scheme.

rpspringuel:
A quick look at the TeX and my idea for implementing this is as follows:
1) Create a new global count to function as a boolean, `\gre@englishcentering`
2) Create user functions which set the value of this count to 1 (use English centering scheme) or 0 (use what is specified in the gabc score)
3) Add variables `\gre@firstsyllablepart` `\gre@middlesyllablepart` and `\gre@endsyllablepart` and to `\gresyllable` and use these instead of references to arguments `#1`, `#2`, and `#3`.
4) Add a check at the beginning of `\gresyllable` which maps the arguments to the new variables based on the value of the new count.

Assuming I haven't missed anything, I should be able to implement this fairly quickly.

eroux:
I can't see any reason no to close it... Tell me if there is some!

rpspringuel:
No reason.  I just forgot to close it when I merged the commit.  Sorry.

PR #28, opened 2015-01-18
Adding option to force English centering scheme from TeX.  RE: #27

rpspringuel:
RE: #27

This fix will only affect scores in which the gabc header either doesn't specify a centering-scheme or "centering-scheme: latin;" (first glyph of the neume is centered over the vowel).  In those cases adding "\englishcentering" to the TeX will switch to using the English centering scheme (first glyph of the neume is centered over whole syllable).  "\defaultcentering" changes things back to normal.  These commands will have no effect on scores where the gabc header has "centering-scheme: english;".

### DIFF
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -502,6 +502,19 @@
 
 \let\grefixednexttextformat\gretextnormal %
 
+
+\newcount\gre@englishcentering
+
+\def\englishcentering{
+	\global\gre@englishcentering=1
+}
+
+\def\defaultcentering{
+	\global\gre@englishcentering=0
+}
+
+\defaultcentering
+
 %% general macro : it will typeset the syllable : arguments are :
 % #1 : the first letters of the syllable, that don't count for the alignment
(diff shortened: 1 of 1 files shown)


eschwab:
Has a standard been set for a naming scheme with gregoriotex variables? It seems preferable to me to keep using the `\gre*` format. I propose changing:

`\englishcentering` to `\greenglishcentering`
`\defaultcentering` to `\gredefaultcentering`

rpspringuel:
No standard has been set, yet, though many of the command are prefixed with `\gre`.  I'm working on establishing the following standard in my texrenaming branch (https://github.com/rpspringuel/gregorio/tree/texrenaming):

1) functions internal to gregoriotex are named `\gre@*`
2) functions which appear in the gtex output of gregorio, but which should not be accessed by users are named `\gre*`
3) functions which are meant for user access and use have no prefix

My changes in this commit follow that scheme.  Should I perhaps open an issue specifically for discussing the command naming scheme?

eroux:
Wonderful! Thank you very much @rpspringuel !

Issue #29, opened 2015-01-20
Documentation on macros callable by the user.


eschwab:
There is a lack of good documentation on macros that the user can call in his main TeX file. I am aware of the list here: http://home.gna.org/gregorio/gregoriotex/details but it is out of date.

It would be helpful to have a list of macros. A simple item/definition list in plain text or markdown would possibily suffice. I am willing to work on this.

eroux:
You're right, this will be very helpful! Markdown is enough yes.

rpspringuel:
This is part of what's taking me so long to finish establishing a macro naming convention; I'm documenting the macros as I go.  Take a look at the Command_Index_* files in the texrenaming branch in my fork.  I've created a pull request (#30) to update the base fork's texrenaming branch to make finding them easier. 

Since I've been documenting things as I go, the documentation reflects the naming convention I'm headed towards rather than the current working state (with a couple of exceptions I haven't yet decided how to deal with).  Still, it should be useful as a starting point if you want to make use of it.

eschwab:
@rpspringuel Your Command_Index_* files are mostly what I was looking for. After #33 is merged I may consider adding examples or something to help out a new user. Would need to find a way to keep it simple to not interfere with the cleanness of the Command_Index* files.

A doc/ section might be helpful and a more obvious location. But, I don't want to introduce multiple lists of macros for maintenance's sake. Just bouncing ideas around at the moment.

rpspringuel:
The only reason for them being in the tex/ folder is because that's where I started my work and I felt my time was better used at this point continuing with the work needed to establish the naming convention rather than cleaning up my working documents.  If you want to start a proper doc/ section in the repository, feel free.  I can always add to them documents in their new location.

It was always my intention to eventually spruce up the Command_Index_* files to make them fully typesettable (and/or converting them to html format for posting on the website).  I'd also hoped, one day, to add examples for the User commands, as I don't expect that the average user would necessarily understand the simplistic, code oriented documentation I've written thus far.  Said day, however, is still well in the future for me.  I'd actually be delighted if you got to (some of) it before me.

eschwab:
Fixed with #34.

PR #34, opened 2015-01-22
Documentation Improvement

eschwab:
WIP: addressing #29 

No need to merge right away. This seems like the best way to make this work easily accessible to all. Comments and suggestions are most welcome.


rpspringuel:
Looks good so far.  Couple of things I noticed (and which are my fault, not yours):

1. There's some inconsistency in how some of the listing of the argument descriptions.  Ones I added earlier in the process look more like sentences (see `\commentary`), while the ones I added later use a more standardized listing format (see `\settranslationcenteringscheme`).  Now that I see a typeset version, I think this should be consistent.
2. Some text indicating what the color definitions at the end of those entries are is probably in order.  Their confusing just tacked on like I left them.

Also, I've added your repository as a remote on my system so that I can pull your changes into my repository, even if this doesn't get merged into the master.  That way when I get a chance to go back to work (probably this weekend) I can add directly to these files, making it less likely that both of us will have to cover the same ground again as this heads towards prime time.

eschwab:
That should resolve point 1. Any comments before I take on the other two files?

rpspringuel:
Looking very nice.

With the other two files (and perhaps with this one, though I'm not as sure), you might want to think of a way to incorporate the file identifiers into the presentation (i.e. pointers to the file where a function is defined).  Right now they're comments, and thus don't show up in the typeset document, but that's mostly because I couldn't figure out what to do with them.  A regular user doesn't really need to know where particular functions are defined, but that information can be very useful to a developer looking to solve a bug or implement a new feature.  Recording the information as comments was my way of keeping track of it while I figured out how to present it.

eschwab:
Finished with Command_Index_gregorio. Would be fairly easy to add the source file pointers to Index_user. I think I will just for consistency.

rpspringuel:
With the first draft done, I'm going to merge this into the main repository if there are no objections.  That should keep the commit history a bit cleaner when I get back to my work.

eschwab:
None from me. I was just sitting on making the request pondering making some further changes/additions, but now it a good breaking point.

PR #37, opened 2015-02-04
Installation Issues

rpspringuel:
Added a script to the fonts directory for converting std to ttf; needed for gresym and greextra.
Added gresym.ttf and greextra.ttf to fonts directory.
Added some files which were generated on my system during install procedures to .gitignore


eroux:
Thanks! In fact I don't quite understand why it wasn't there already... Could you please just convert the fontforge script to python? I'm not sure if it's still valid, but fontforge had plans to remove fontforge scripts...

rpspringuel:
Looks like I'm not going to be able to solve my python/fontforge issues today (and probably not before the weekend).  If someone else has a working system already and could test the script, it will speed things up.

rpspringuel:
Turns out my problem was a question of having too many simultaneous Python installations on my machine.  The scripts weren't pointed at the installation which had the fontforge extension, resulting in my tests always failing when it tried to import it.  Having cleaned out some of the extraneous installations and reestablished all the appropriate symlinks, I was able to get things up and running and finish this up.

If someone would kindly check my work before I merge.

eschwab:
Sent pull request. Everything checks out for me.

eroux:
Thanx!

PR #45, opened 2015-02-10
Fix text formatting from leaking to previous syllable.

eschwab:
Fix as indicated by @rpspringuel in the Gregorio-users mailing list.

### DIFF
--- a/plugins/gregoriotex/gregoriotex-write.c
+++ b/plugins/gregoriotex/gregoriotex-write.c
@@ -34,7 +34,7 @@
 
 #include "gregoriotex.h"
 
-typedef enum { NEXT_SYL, THIS_SYL } syllable;
+enum syllable { NEXT_SYL, THIS_SYL };
 
 #if ALL_STATIC == 0
 DECLARE_PLUGIN (gregoriotex)
@@ -1236,7 +1236,7 @@ gtex_print_char (FILE *f, grewchar to_print)
 }
 
 void
-gregoriotex_write_text (FILE *f, gregorio_character *text, char *first_syllable, syllable next_syl)
+gregoriotex_write_text (FILE *f, gregorio_character *text, char *first_syllable, int next_syl)
 {
   char first_syllable_val = 0;
   if (text == NULL)
--- a/plugins/gregoriotex/gregoriotex-write.c
+++ b/plugins/gregoriotex/gregoriotex-write.c
@@ -34,7 +34,7 @@
 
 #include "gregoriotex.h"
 
-enum syllable { NEXT_SYL, THIS_SYL };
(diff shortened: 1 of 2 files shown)


eroux:
This most certainly will break things for current syllable. You need to pass an additional argument to the function, telling if it's writing next syllable or current syllable.

rpspringuel:
I'll admit, that it's been years since I programmed in any variant of C, so this suggestion is cobbled together from some quick internet searches and my sketchy memory:

1. Add `typedef enum { NEXT_SYL, THIS_SYL } syllable;` near the head of the file (probably around line 36).
2. Add the parameter `syllable next_syl` to the definition of `gregorio_write_text` (line1237)
3. Convert lines 1252 to 1254 to a `if (next_syl) {} else {}` block.
4. What is currently lines 1252 to 1254 become the `else` code for the above block.  The modified versions in @eschwab's commit become the true code
5. Find call sites of `gregorio_write_text` in the program and add `NEXT_SYL` or `THIS_SYL` as appropriate to the parameter list.

eroux:
precisely!

eschwab:
Upon further tesing (on `master`) I cannot reproduce this bug. I am using the supplied mwe:

    name: excerpt of Kyrie XII;
    office-part: Kyriale;
    mode: 8;
    commentary: XII. s.;

    %%
    (c4) KY(g')ri(h)e(igkvJI'HG.) *(,) e(hg)lé(hih')i(h)son.(g.) <i>bis</i>(::)
    Chri(g)ste(gegvFED.) e(f)lé(e')i(f)son.(g.) <i>bis</i>(::)

I just noticed that the bug reporter is using gregorio in scribus. This may be a scribus issue?

rpspringuel:
I can reproduce the bug and I'm not using scribus.  Did you reinstall gregorio?  Since the bug is in the c, you have to rebuild and reinstall gregorio after it was introduced in commit gregorio/master@f2d354fbb3e7446215a4a8b263744a15f92572e4

eschwab:
Bah, the problems of working from two different boxes. I have not rebuilt on this box for a couple of months.

eschwab:
That should do it. Builds and works for me.

eroux:
@rpspringuel : if it's ok for you, you can merge with the main repo (I don't have the opportunity to test right now). Thank you both!

rpspringuel:
Hmm... I just tested this and I'm getting backwards behavior.  @eschwab can you double check your fix and test the change I submitted.

eschwab:
Will do. Might be a bit, later this evening. I want to clean up the TeXLive and gregorio install on this box to make sure everything is running right.

eschwab:
I had quite the mess of an install. Anyways we should be good now. The behavior was indeed backwards. Ready to merge.

PR #46, opened 2015-02-11
Backwards behavior.

rpspringuel:
My testing had `\gresetfixedtextformat` and `\gresetfixednexttextformat` reversed.  This change fixed that.


rpspringuel:
Merge to wrong place.

Issue #41, opened 2015-02-08
template: simple way to specify static initial length


eroux:
The template should contain the following definition for the width of the initial: 2mm+1em+2mm (em, here would in the initial font), whatever the initial letter.

rpspringuel:
I can't get the em and ex units to work properly, but I've made an attempt to address this.

PR #47, opened 2015-02-16
Requested fix for #41

rpspringuel:
I've figured out that the problem with using em and ex is related to the scope, but haven't quite worked out how to fix it yet.  Probably need a `\global` somewhere, but my first few tries haven't worked.

Since this is useful as is, I'm putting it in place now, though I'll continue to work on the scope problem.


rpspringuel:
Cross-referencing #41

Issue #38, opened 2015-02-07
Centralized Version Number


rpspringuel:
Right now the version number is hard coded into several different files, making it difficult to update.  I think it would be nice if we could have a single text file which contains the current version number (and nothing else) and then have the various files which need the version number load it from there.  Is this possible?  A quick google search tells me that [Makefiles can import information from a file into a variable] (http://stackoverflow.com/questions/6767413/create-a-variable-in-a-makefile-by-reading-contents-of-another-file) (at least in theory, I haven't tested it) and InnoSetup (which creates the Windows installer) can do [something similar] (http://stackoverflow.com/questions/14530504/innosetup-how-to-declare-a-variable-by-reading-from-a-file).  I know from experience that python scripts can do so, and I assume that lua and C can as well.

Alternatively (or perhaps additionally if there are files which can't read the version number in themselves), a script could be written which would systematically replace the version number in all files where it appears.

rpspringuel:
Oh, and it would also be nice if commits to the main project repository would trigger a webhook which would automatically increment the version number.

eschwab:
Could pull the version number from the commit hash. Would not be a nice +1 increment though.

rpspringuel:
The issue I see with a simple count is that it doesn't have any reset possibilities.

If we are at version number 2.4.1, then ideally the hook should increment the last, and only the last number, changing the version number to 2.4.2.  A subsequent commit should increment to 2.4.3, etc.  However, at some point we're going to want to manually increment one (or both) of the first two numbers and reset the third: e.g. 2.5.0.  A simple count would continue to increment the third number resulting in something like 2.5.4.  This, I think, would be unduly confusing.

You are right, though, that only commits to the master branch should matter (as your suggestion implies).

eschwab:
Fixed by #48 and #51. Please close.

PR #48, opened 2015-02-17
Centralized Version Number

eschwab:
Addressing #38.

Still needs a way to automatically run the script to push the version number to the files. I think the makefile can be hacked to do this, but I am not sure right now.

Also I think the `VERSION` and the `.py` files can find a better place to reside at.

As far as bumping the version number, a webhook may not work to well. I am concerned about having an endless cycle of commits. It would be possible to add a git commit hook to +1, but that is not practical for a project with multiple contributers.

Not ready to merge. (also the version number is probably wrong)


eschwab:
I've improved the version changer script. Now we can change both the gregorio and gregoriotexapi versions.

Still need to resolve how and when to bump the version numbers in the `VERSION` file.

Secondly, when should the script run? I can think of two options:
* Execute during/prior to building. This option makes sense if we want to dynamically update the versions based upon the commit. This seems viable for changing the `gregorio` version.
* Execute periodically when it seems necessary. To me this option seems better, at least for the `gregoriotexapi` version to avoid needing to constantly re built the `gtex` files.

eroux:
I'm also in favor of the second version, thank you very much for the script!

eroux:
As the author, I leave you the honor to run it first (these version number definitely need updating!)

eschwab:
The `gregoriotexapi` version is fairly straight forward - set to current date when updated.

What is the versioning scheme for `gregorio`? Should we start with `2.4.1-git` and go from there?

eroux:
As you certainly have noticed, the version numbering is quite a mess... When I started Gregorio I made very few releases, not often at all, and I kept this bad habit because the Mac and Windows builds are very difficult to do (I depend on other people for the Mac build and I cannot really test the windows build). Now Windows build is much easier, but Mac is still problematic... If there was a simple and reactive way to make a Mac build, I think we should build much more often... So, finding an easy build to build for mac would be the first issue to fix before being able to fix the version numbering.

rpspringuel:
Building on Mac from source is as straight forward as it is for Linux.  I do it on a regular basis. The difficulty is putting together a installer package, MacPort package, or Homebrew package.

For the version number, I'd remove the `-git` tag and simply stick with incrementing the numbers.  I'm also in favor of updating the version number often.  Preferably with every change to the master branch, but at least once every week or two (assuming changes were made during that period).  It makes diagnosing and fixing problems easier if you can easily identify which version of the software someone is running.

Oh, and the ISS file (which builds the windows installer) need to be updated too.

eschwab:
Oh, I forgot about the windows build. I'll add that and drop the `-git` if @eroux does not object.

eschwab:
Been thinking more on how to increment the numbers. From the developer's perspective it makes sense. As a user though being required to constantly reprocess my gabc files would become annoying. If the `api version` could be updated sparingly, that would be great. Perhaps limit that number to important changes like Grefactor Initialization proposal and texrenaming.

That being said, I think incrementing the `gregorio version` would be helpful and would not put undue stress on the user. Minor changes to gregoriotex can still be referred to/tracked down this way. I am not opposed to a weekly or periodic update, but I wonder to whom will that responsibility fall and will that become an unnecessary burden? As an option we could have a script on a cron job to increment the version and build the Win/Mac binaries, if someone had a server to do this on.

Personally I think dynamically linking the `gregorio version` to the commit hash at build time may be a better option. That could give a version of the style `2.4.2-cb57e71`. The `2.4.2` can remain static in the VERSION file while the `HASH` is insertes at build time. Doing it in a semi-static, semi-volatile way avoids writing to any files in the repository, thus making it dirty and more difficult for the end user to update in the future.

rpspringuel:
The user would only be required to reprocess their gabc files when they updated, and chances are if they are updating it's because they want to make use of something that's been added to the code base.  Still, that might not apply to every project, so I see you point.

Would it be possible to make `\includegabcscore` (or rather the lua script that it calls) more intelligent so that it doesn't just check the relative dates of the gabc and gtex score, but also checks the api version of the gtex score (if it finds one) against the current api version?  In this fashion the recompiling can be hidden from the user, it just happens automatically when its needed.

eschwab:
Just this morning I needed to print another copy of a project, but there was a change of one or two words in the regular text. So I changed that, ran `lualatex` and it promptly died due to mismatched api versions. I had to reprocess ~20 gabc files. Granted they were pre `20130616`, but they did not need any of the new features added since then. (I should just write a shell function or alias that does the whole tree of gabc files.) That it why I hesitate to change the `api version` too frequently.

However, your idea just may work, and it would resolve my concern. As I look at it now, it may be possible to combine both `\includegabcscore` and `\includetexscore` into one.

eroux:
@eschwab the change in api version are not really about new features, but about incompatibilities. The most obvious example is the character numbering shift which caused so many troubles on the mailing list... api changes are mostly about that, rarely about new features. That's why it's been introduced: to enable users to see that there is an error and tell them to recompute their score, instead of just having a weird output which they think is a bug...

@rpspringuel it would require a few tricks, but I think it's possible yes... that would be a very good option!

rpspringuel:
@eschwab If you look in texrenaming, you'll see that I've already got a version of `\includescore` that combines the functionality of `\includegabcscore` and `\includetexscore`.  I wrote that as an experiment in leveraging the abilities of `xstring` but haven't really had a chance to do much testing on it (I made a mistake in doing it in texrenaming because of how many changes that requires).  Let me see if I extract those changes and create a development branch with that functionality working.  I'll open another issue when I've got something to share.

eschwab:
That is a great start on a unified `\includescore`. I would then extend the lua function `include_gabc_score` to add a cross check on the `api version` and recompile the gabc if necessary. We could also chage `\gre@includetexscore` to pass `#1` off to a new lua function to do the same.

rpspringuel:
Extracted and tested.  Seems to be working on my system.  See #53.

rpspringuel:
Thinking about this, we really sort of have 4 version numbers:

1. The gabc version number (0.9.3).  This indicates the syntax of the gabc file.
2. The Gregorio version number (2.4.1).  This indicates the capabilities of Gregorio for converting amongst gabc, GregorioXML, and GregorioTeX.
3. The gregoriotex api version (20150220).  This indicates what commands are available to appear inside gtex files.  These are the gregorio level commands are defined in gregoriotex (they are the `\gre*` commands under the naming convention I'm in the process of applying), but are written by Gregorio (via gregoriotex-write.c).
4. The gregoriotex version (not currently numbered).  This indicates the capabilities of gregoriotex for typesetting scores.  They consist of the User and internal level commands defined in gregoriotex.

As I see it, there's a considerable amount of overlap between these version numbers currently.  Changes in 1, for instance, necessitate changes in 2 (otherwise Gregorio won't be able to make sense of the new or changed syntax).  Perhaps, though, it makes sense to keep them all separate because it might be possible to have a bug or something that only requires changing one of them.

Most of the recent changes in the TeX haven't really affected the API as I defined it above.  What's been effected is how gregoriotex renders the content indicated by the API commands.  That would require a change to the gregoriotex version number, but not the API version number.  For instance, it's possible to fool gregoriotex into processing a file generated a few months ago (with API version 20130616) by changing the API version number in the gtex file to 20150220.  This doesn't raise any errors (though #55 does affect the visual results).

Am I making sense?  Should we perhaps be more specific about what are version numbers are and which parts of the code belong to which version numbers.

Am I making any sense here?

PR #49, opened 2015-02-20
Finished the fix for #41

### LINKED ISSUE #11: adjust lines beginning to annotation
Set up a minimal space between annotations and lines, and between initial and lines, so that spacing is fine by default. Annotation should never be behind lines.

rpspringuel:
em and ex units now used in the font of the initial when used to set the initial spacing (finishes #41)
Moved a few distances related to staff line spacing which were being defined in gregoriotex-spaces to gps-default with all the rest
Changed out some \ifnum\number\dimen constructions for \ifdim\dimen to make the code clearer.
Added check of initial width against the width of the annotation boxes.  Long annotations will no longer bleed behind staff (fixes #11)

### DIFF
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -132,12 +132,6 @@
 %% macros for the typesetting of spaces
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-%% All the following values correspond to a grefactor of 1,
-%% so in most cases it will be much smaller that what you will finally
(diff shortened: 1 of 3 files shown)


eroux:
I didn't test yet but it seems alright. Did you test a little with a grefactor other than 17?

rpspringuel:
I did.  That's actually what led me to realizing why my first attempt didn't work as expected.  The rescaling of distances when the `grefactor` changes was throwing things off (particularly the "hidden" rescaling at the beginning of the first score).  I'm opening a new issue about it: #50

PR #52, opened 2015-02-22
Update gregorio version for windows installer.

eschwab:
And drop the `-git` flag per @rpspringuel 's suggestion.


rpspringuel:
Probably should have done this before updating the development download.  Oh, well.

PR #53, opened 2015-02-22
includescore upgrade

rpspringuel:
These modifications allow `\includescore` to handle both gtex and gabc file.
	Still to do: there should be a check of the date of gtex files against their original gabc file to see if they need to be recompiled by date (a function currently handled by the greg-book engine), the lua script which handles gabc files does this already so that might be leveraged for this.  Also need to add api version checker to both sides
I've also changed the extension of the auto generated file.  This will prevent conflicts with LilyPond-Book (which doesn't like included files to have the extension tex).

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -191,7 +191,7 @@ local function include_gabc_score(gabc_file)
         return
     end
     local gabc_timestamp = lfs.attributes(gabc_file).modification
-    local tex_file = gabc_file:gsub("%.gabc+$","-auto.tex")
+    local tex_file = gabc_file:gsub("%.gabc+$","-auto.gtex")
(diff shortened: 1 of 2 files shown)


eroux:
Ok for me!

eroux:
By the way, if someone has time, it would be great to do a little changelog... the NEWS file could be a simpler for of it respecting [GNU changelog style](https://www.gnu.org/prep/standards/html_node/Change-Logs.html) (or even the output of [this script](http://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob_plain;f=build-aux/gitlog-to-changelog)), but it would be very useful to start a changelog.md file in [this format](http://keepachangelog.com/), even for us... If someone feels like starting it, I'll add the features done since 2.3 too (or I'll start it later on).

Issue #55, opened 2015-02-23
writemode and setfirstannotation conflict


rpspringuel:
If the user manually indicates an annotation in their TeX file, `\writemode` should not overwrite it with the mode from the gabc file.  This could be accomplished by having `\writemode` check to see if there are contents in the box containing the first annotation before entering the mode (and refraining from doing so if there is some thing there), or by having `\setfirstannotaion` change a flag which controls whether `\writemode` does anything.  In the latter case the flag would need to be reset at the end of the score (to allow for multiple score inclusions).

eroux:
That's right, indeed!

PR #56, opened 2015-02-23
Stop gregoriotex from overriding user defined \setfirstannotation

eschwab:
Bug fix for #55 : Setting the mode header in the gabc file would override the
user defined \setfirstannotation in the main tex file. Now \writemode
checks to see if \GreAboveinitialfirstbox has been set.


eroux:
Thanx!

PR #64, opened 2015-03-04
Fix 50 piece

rpspringuel:
Everything I've done so far is tested and working on my machine.  I'm pretty sure I've got just one thing left to do: convert the user settable distances to strings.

### DIFF
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1483,7 +1483,7 @@
 }
 
 \def\gredivisiomaiorsymbol{%
-  \ifnum\gre@stafflinefactor=10\relax %
+  \ifnum\gre@stafflinefactor=17\relax %
     %\gre@calculate@glyphraisevalue{g}{0}% bar glyphs are made to be at this height
     \raise\gre@glyphraisevalue\hbox{\gregoriofont \char \gre@char@divisiomaior}%
   \else %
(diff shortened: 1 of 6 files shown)


jperon:
Would it be possible, please, to have explanations on how to set spaces within a document ? With that commit, my old \grechangedim commands are deprecated. Is it necessary to use \makeatletter and \makeatother within the doc, or is there a cleaner way to define lengths ?
Thank you !

rpspringuel:
Sorry.  I haven't posted any explanation because this commit doesn't represent the final state of the code.  There's one more which implements a new, and hopefully cleaner, interface for adjusting differences.  I expected to be done with that by now, but have hit an unexpected snag in making the code work.  Several functions which work fine on their own in my testing, don't seem to want to play well together and I haven't been able to figure out why.  I'm in the process of preparing a MWE so that I can ask for some help on TeX Stackexchange.

For now, you'll have to make use of `\makeatletter` to change the dimensions.  All the dimension names are now `\gre@distance` where they used to be `\gredistance`.

jperon:
That's perfect ! I'll wait for other changes and docs. Meanwhile, I'll ask a beginner who asked me for help to wait a little…

rpspringuel:
I've got the fix.  Expect it to be in-bound latter today.  I just have to finish some updating the website downloads first.

Issue #65, opened 2015-03-08
Problem with gregorio and multiline scores


henryso:
When I try to compile this score using `lualatex --shell-escape --interaction=nonstopmode mwe.tex`, it fails.  See output below.

###### score.gabc
%%
(c4) Line(c) One(d) (z) Line(c) Two(d)

###### mwe.tex
\documentclass[11pt]{article}
\usepackage{luatextra}
\usepackage{graphicx}
\usepackage{gregoriotex}
\usepackage[utf8]{luainputenc}
\usepackage{times}
\begin{document}
\includescore{score}
\end{document}

###### output
This is LuaTeX, Version beta-0.70.1-2014113017 
 \write18 enabled.
(./mwe.tex
LaTeX2e <2011/06/27>
LuaTeX adaptation of babel <v3.8m-luatex-1.5> and hyphenation patterns for engl
ish, dumylang, nohyphenation, loaded.
(/usr/share/texmf-dist/tex/latex/base/article.cls
Document Class: article 2007/10/19 v1.4h Standard LaTeX document class
(/usr/share/texmf-dist/tex/latex/base/size11.clo))
(/usr/share/texmf-dist/tex/lualatex/luatextra/luatextra.sty
(/usr/share/texmf-dist/tex/generic/oberdiek/ifluatex.sty)
(/usr/share/texmf-dist/tex/latex/fontspec/fontspec.sty
(/usr/share/texmf-dist/tex/latex/l3kernel/expl3.sty
(/usr/share/texmf-dist/tex/latex/l3kernel/l3names.sty
(/usr/share/texmf-dist/tex/latex/l3kernel/l3bootstrap.sty
(/usr/share/texmf-dist/tex/generic/oberdiek/luatex.sty
(/usr/share/texmf-dist/tex/generic/oberdiek/infwarerr.sty)
(/usr/share/texmf-dist/tex/latex/etex-pkg/etex.sty)
(/usr/share/texmf-dist/tex/generic/oberdiek/luatex-loader.sty
(/usr/share/texmf-dist/scripts/oberdiek/oberdiek.luatex.lua)))
(/usr/share/texmf-dist/tex/generic/oberdiek/pdftexcmds.sty
(/usr/share/texmf-dist/tex/generic/oberdiek/ltxcmds.sty)
(/usr/share/texmf-dist/tex/generic/oberdiek/ifpdf.sty))))
(/usr/share/texmf-dist/tex/latex/l3kernel/l3basics.sty)
[...]

PR #67, opened 2015-03-08
Fix error with manual linebreaking in a score. #65

eschwab:
I think TeX could not do the `\ifdim` math with the rubber length of the `newskip`.


eroux:
Many thanks! @rpspringuel , as it concerns mainly your changes, I let you merge it.

rpspringuel:
Thanks for the catch.  Not sure why this didn't show up in my own testing.

eschwab:
Only happened when `(z) or (Z)` was used in the gabc. Perhaps your testing gabc does not have one?

rpspringuel:
Checked my test gabc an you're right, there wasn't one.  I've fixed that now so hopefully I won't see this sort of thing again.

PR #77, opened 2015-03-12
Simplify the fonts makefile.

eschwab:
From Henry So Jr.

This patch does two things:

Firstly, it consolidates tds, localinstall, and install.  I think this
will make it easier to keep things in sync since there's less
duplication of code.

Secondly, it makes texhash a variable so I can override it in a build.

### DIFF
--- a/contrib/gregorio-scribus.lua
+++ b/contrib/gregorio-scribus.lua
@@ -79,7 +79,7 @@ end
 
 local format = string.format
 
-texfile:write(format("\n\\includescore{%s-score.tex}\n\\end{document}\n", pathbase))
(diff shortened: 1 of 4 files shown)


eschwab:
I also added new macros to force gregoriotex to recompile scores. I'll update the docs for the new include_score behavior after the docs get merged in.

`\forcegabccompile` - By using this macro in the main tex file the user,
can force gregoriotex to recompile scores in the document. Useful for
when there are changes/bug fixes to gregoriotex and the API VERSION is
not changed.

`\normalgabccompile` - reverses `\forcegabccompile`. Gregoriotex will
return to its automatic score compiling behavior.

eschwab:
Added a fix for #83 to make scribus use `\includescore[f]{}`.

rpspringuel:
Please resubmit against develop.

PR #81, opened 2015-03-15
Bug fixing for 2.4.3

rpspringuel:
Before this is merged in, do we want to move the 2.4.3 tag to the point where all these bug fixes (plus whatever else the rest of you have) are in effect, or make a new release?  If a new release, then I'll need to update the CHANGELOG further.  If the tag gets moved, then these are part of stuff that's already listed there for 2.4.3


eschwab:
I vote for another release. That would help us in future bug testing if a user comes with `2.4.3` we don't know if that is pre-bug squashing or not.

rpspringuel:
We're going to yank this release and restructure the git workflow a bit.  These fixes will be incorporated, I'm just not sure where yet, so I'm going to close this request without merging for now.  Once I know where the fixes are going to go, I'll open the appropriate pull request (or push them directly).

PR #86, opened 2015-03-16
Backwards compatibility for include score

rpspringuel:
Picking up on something inspired by Henry and Br. Elijah: These proposed changes would make \includescore backwards compatible while still making the refinements available.  Behavior is controlled via the `\forcegabccompile`, `\normalgabccompile`, and `\forcedgtexinclude` macros, or via the `forceinclude`, `forcecompile`, and `autocompile` options for the LaTeX package.  Default behavior is set to mimic the old behavior of `\includescore`.

These changes have not been tested yet.
I'm presenting them here for discussion before doing that.  I'm especially interested to know whether this should be part of the 3.0 release (this pull request would merge into that release) in order to make upgrading easier on people.

#### Edit: Testing is complete.  See updates below.


eroux:
I would in favor of including it in 3.0 yes... Just a small question: wouldn't it be better to merge it into develop, and then into 3.0? Or will you merge it into develop afterwards?

rpspringuel:
Merging into develop and then into release-3.0 will require cherry picking the commit out of develop to make sure that no other feature enhancements get into 3.0.  On the other hand, any changes that are merged into release-3.0 can be merged into develop without worries, assuming no conflicts (since develop takes all changes).  As a result, I think it'll be easier if changes that are going into the 3.0 release are merged there first and then into develop.

tex/gregoriotex.tex
```diff
@@ -1287,45 +1287,90 @@
...
-	\gre@warning{\protect\greincludedgabcscore\space is deprecated.\MessageBreak Use \protect\includescore\space instead.}%
+	\gre@warning{\protect\greincludegabcscore\space is deprecated.\MessageBreak Use \protect\includescore\space instead.\MessageBreak Perhaps with the c option.}%
 	\gre@includescore{#1}%
 }%
 
 % A macro that passes the score directly to TeX without performing the API version check or if the gabc was modified since the creation of the gtex file.
 \def\gre@includescorewithoption[#1]#2{%
-  \input #2%
-  \relax%
+	\ifx #1f\relax%
```
eschwab: Very good. I wanted to add this test.

eschwab:
I have not tested yet, but looks good. It may be helpful to add `\normalgtexinclude` to give a full suite of macros to the user. Some may wish to use them all in one document. I could see when using `[autocompile]` one may want to `\forcegtexinclude` then revert back to `\normalgtexinclude`.

rpspringuel:
While I realize that two booleans give four possible states, `\normalgtexinclude` would be redundant in my opinion because there are really only three modes:

1. Forced include: accessible via option `forceinclude` or via macro `\forcegtexinclude`, this mode never compiles scores.  In terms of the boolean flags, this mode is used when `\if@gtexforceinclude` is true.  The value of `\if@gabcforcecompile` is ignored.
2. Forced compile: accessible via option `forcecompile` or via macro `\forcegabccompile`, this mode always compiles scores.  In terms of the boolean flags, this mode is used when `\if@gtexforceinclude` is false and `\if@gabcforcecompile` is true.
3. Auto compile: accessible via option `autocompile` or via macro `\normalgabccompile`, this mode compiles scores which fail one of three tests.  In terms of the boolean flags, this mode is used when both `\if@gtexforceinclude` and `\if@gabcforcecompile` are false.

Thus, if one wanted to force includes for part of a document and then switch to auto compiling, one would use `\forcegtexinclude` for the first part and `\normalgabccompile` for the later.

Perhaps renaming things to `nevercompile`, `autocompile`, and `forcecompile` would make the tripartite nature of these modes clearer?
3. Auto compile:

eschwab:
Looked closer at the code, I did not see the `@gtexforcecompilefalse` the first time. Yes, renaming to `{never,auto,force}compile` seems to be a clearer option.

rpspringuel:
In that regard, it occurred to me that it might also be clearer to use a single three state switch rather than two two state ones.  I.e. have `\gre@compileoption` which has values 0, 1, and 2.  I could then make case of `\ifcase` rather than have embedded if clauses.  I'll tweak the code and you'll see what I mean.

eschwab:
That would be a cleaner solution code wise.

rpspringuel:
Okay, I've modified the code and have had a chance to test now.  Everything should be ready to go.  I'll wait for others to take a look see before merging though.

eschwab:
DId not test, but looks good.

rpspringuel:
Something else that just occurred to me.  Perhaps we should have three optional arguments for `\includescore`, corresponding to the three possible behaviors: `n`, `a`, and `f`.  This would leave it so that the macros and options set the default behavior, but would now make it so that the User could override that behavior in any way they want for any single call of `\includescore`.

Thoughts?

(Sorry if this creates additional documentation work, Br. Elijah.)

henryso:
I like the idea of having these options, but I don't think it's worth it if it over-complicates the code or maintenance of such.

eschwab:
Have at it. Providing a full suite of macros is a good thing. I'll update the docs as needed.

rpspringuel:
@henryso the change is relatively trivial.  In fact I've done it already.:P

eschwab:
Tested, works well.

Issue #87, opened 2015-03-16
Unified Version Number


rpspringuel:
Given that we're releasing more regularly, @eroux and I have discussed the possibility of eliminating the GregorioTeX API version number and moving to a single version number, 3.0.0 in the case of this release.  This should make it easier on the user to keep track of what's going on.

This will require a couple of modifications to `VersionUpdater.py` and `VERSION` to implement.

VERSION should also probably include some guidelines for when and how incrementation is done.  The ones that @eroux and I discussed are:

> We adopt the Semantic Version numbering scheme of x.y.z and only have the one version number.  The alternate version numbers for other parts of the project would be officially dropped.
> 
> z - This number is incremented for bug fixes.  All the work I did recently correcting glues that were leaking onto the score would qualify.  This resets to 0 when x or y increments.
> 
> y - This number is incremented for enhancements.  The new EUOUAE tags would qualify.  This resets to 0 when x increments.
> 
> x - This number is incremented for interface changes.  The recent change to how distances are set and used would qualify.
> 
> Code refactoring (i.e. changes which do not effect the user's experience in any way) would not qualify for a change in any of the numbers.
> 
> Release Schedule - In any month where we make commits to the repository, on the last day of the month we assess those commits and decide whether they warrant a bump to the version number, and if so to which part of the number.  The goal here is not to say that we should release once each month, but rather that we shouldn't release more often than once each month.  If a month has been somewhat lacking in changes, we're still bug swatting those changes that we did make, or we're still in the process of documenting them, then we simply don't release that month and put things off until the following one.

eschwab:
Couple of things.

The API version checks will remain, just that instead of `yyyymmdd` it will have `3.0.0`? Fine with me.

If it is to be `3.0.0` in 'gregoriotex' can we either drop the .'s or replace them with a different sep char? Reason is that I am opposed to filenames with .'s that don't introduce suffixes. I really would not like seeing `antiphon-3.0.0.gtex` but rather `antiphon-3_0_0.gtex` or `antiphon_3-0-0.gtex`.

Either way, I can edit the versioning files when ready.

eroux:
@eschwab is right here I believe, dots should be avoided in filenames.

rpspringuel:
[...]

PR #92, opened 2015-03-17
Unify Version Numbers.

eschwab:
Addresses #87.

Removes the gregoriotex_api version and replaces it with one
gregorio_version.


rpspringuel:
This is my fault (since I wrote it wrong in the first place), but looking at http://semver.org/, it should be 3.0.0-beta, not 3.0.beta.

Also, please note the changes in CHANGELOG.

eschwab:
I'm done for the night, Will update the CHANGELOG tomorrow.

eschwab:
This should be good to go now.

eschwab:
No more false starts this time. If this looks good, merge away!

rpspringuel:
Hmm... I've been top posting my additions to CHANGELOG and I noticed that you bottom posted yours.  Does it matter which we do?  Should it?

The code changes look good on inspection, but I don't have time to run any tests tonight to double check.  I'll make some time for that tomorrow and merge if I don't find anything wrong.  (Unless someone beats me to it.)

eschwab:
I'm not sure it should matter where items are noted in the CHANGELOG. My thought is that as long as it is noted under the right release number its fine. But then it may be easier to dig through the commit history if the CHANGLOG is kept chronological.

eroux:
I'm not sure it's the best place to discuss this, but I was asking myself one question: if we now use SemVer, this means that users don't have to compile their score when changing from x.y.z to x.y+.z nor to x.y.z+1, but only when changing to x+1.y.z. Is it taken into consideration into version checking?

eschwab:
No, at this point any change in version number requires the score to be recompiled. I'm thinking that this PR only deals with unifying the version numbers and make a separate issue to change the rules for version checking. I can take care of that also, but these two tasks are different enough that it may be better to have separate commit chains for each.

eroux:
Sure, it was just a reminder, it's no emergency, it can be dealt with separately

rpspringuel:
Time got away from me today and my schedule tomorrow is too tight to work on this.  As a result it'll be Friday before I can double check this.  If someone else can get to this before then, feel free.

rpspringuel:
When I changed the version number and ran `VersionUpdate.py` to test, I found the following bugs:

1. the `-beta` didn't get removed from the version number in the iss file.
2. `gregoriotex.h` was not modified
3. `gregoriotex.lua` was not modified

rpspringuel:
Everything seems to be working great now.

Issue #89, opened 2015-03-16
UserManual updates.


rpspringuel:
If you've added (or changed) user functionality to the code recently, please make sure that it is listed in the UserManual (in the doc folder).  This is especially important for stuff which is new to 3.0.  I realize that there is a backlog of documentation updates that are needed, but for the moment I'm just concerned about the new stuff.

eschwab:
I'm working on enhancing `\includescore` documentation. @rpspringuel , I can deal with the things added in #86

eroux:
this could be closed right?

eschwab:
I was just looking through the commit log to check, and everything seems to be taken care of.

PR #93, opened 2015-03-17
Update Documentation for `\includescore`.

eschwab:
Addresses part of #89


rpspringuel:
I think it might be better to have a separate section for package options, rather than including the in the User commands section.  I realize that we don't have many options right now, but I could foresee turning all the flag controlled behaviors into options long term, in which case having the separate section would help make the document more clearly organized.

eschwab:
Good idea. Done.

rpspringuel:
Obviously you haven't done (or at least not pushed) the changes related to my most recent modifications, but it might be nice when you do so to include a warning that when using the never compile option (in any scope) the filename given has to be for an existing file and the only extension that can be omitted is `.tex` (since this filename gets passed straight to `\input` which is capable of adding that extension if needed).  This is slightly different than the other two options, where the algorithms can be given any of four different extensions (omitted, `.gtex`, `.gabc`, and `.tex`) and still find the right file.

Or maybe we should modify the code so that the never compile option can handle the same set of extensions?

eschwab:
I don't think that would be necessary. As it stands now the default is old gregoriotex behavior and should not suprise users. If someone uses the `nevercompile` feature it is their responsibility to provide a current `gtex/tex` with correct `gregorio_version` and they should be able to provide a good `\includescore` argument.

Definitly I can make a note as you suggest and also encourage omitting the file extension all together as the best practice.

eschwab:
This should be done now. Br. Samuel, could you proofread and make sure that I got all the arguments and their respective descriptions correct?

doc/Command_Index_User.tex
```diff
@@ -6,37 +6,147 @@ \section{User Controls}
...
+behavior of gregoriotex. It is the package default.
+
+\medskip The three previous macros can be combined in the same document to
+switch between different includescore behaviors: \par\medskip
+\begin{latexcode}
+  \usepackage{gregoriotex} % [nevercompile] is the default.
+  ----
+  \includescore{TecumPrincipium} % gabc never compiled.
+  \includescore[f]{TecumPrincipium} % gabc always compiled.
+  \includescore[a]{TecumPrincipium} % dabc auto compiled.
```
rpspringuel: typo: `dabc` should be `gabc` in comment

doc/Command_Index_User.tex
```diff
@@ -6,37 +6,147 @@ \section{User Controls}
...
+
+  \autocompilegabc
+  \includescore{TecumPrincipium} % gabc auto compiled.
+  \includescore[n]{TecumPrincipium} % gabc never compiled.
+  \includescore[f]{TecumPrincipium} % gabc always compiled.
+
+  \forcecompilegabc
+  \includescore{TecumPrincipium} % gabc auways compiled.
```
rpspringuel: typo: `auways` should be `always` in comment

rpspringuel:
Other than those two typos it looks good.  I'll merge once you fix them.

eschwab:
Thank you, and fixed.

Issue #95, opened 2015-03-18
Gregorio should abort if the output filename is the same as the input filename


henryso:
It's easy to accidentally run gregorio file.tex. Unfortunately, this attempts to compile file.tex (which will most likely fail) and then writes the result to file.tex, thus clobbering it.  I will prepare a pull request that aborts if the output filename is the same as the input filename.

eschwab:
Fix by #112

PR #96, opened 2015-03-18
Check to avoid clobbering the input file and GregorioTeX installation cleanup.

henryso:
Fix for https://github.com/gregorio-project/gregorio/issues/95

### DIFF
--- a/fonts/Makefile
+++ b/fonts/Makefile
@@ -39,4 +39,15 @@ greextra: greextra.ttf
 clean:
 	$(RM) $(TTFFILES)
 
+install:
+	@echo 'The Makefile in the fonts directory is no longer used for installing'
+	@echo 'GregorioTeX.  Please use install-gtex.sh from the parent directory'
+	@echo 'to install GregorioTeX.'
+	@echo
+	@echo 'If you were using "make install" before, now use'
+	@echo '    install-gtex.sh var:TEXMFLOCAL'
+	@echo
+	@echo 'If you were using "make localinstall" before, now use'
+	@echo '    install-gtex.sh var:TEXMFHOME'
+
 .DEFAULT_GOAL := all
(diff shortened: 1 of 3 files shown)


henryso:
Implemented https://github.com/gregorio-project/gregorio/issues/97

eschwab:
I'm getting a error:

    $ cd fonts
    # make install
    -------
    make: *** No rule to make target 'install'. Stop.

How should the install process be done with this patch?

henryso:
instead of `cd fonts; make install`, run `install-gtex.sh var:TEXMFLOCAL` from the root gregorio directory.

I am working on an install-everything script, so I haven't documented that any more than what you see in the install-gtex.sh file.

henryso:
Thinking about this, I should probably put in an install "rule" which tells you to run `install-gtex.sh` instead.

eschwab:
Ok, I got the `install-gtex.sh` working. I build a package for my distro (Arch) to keep things organized and easy to switch between builds, thanks for the `dir:` var.

    echo '%%\n(c3) r(f)` > test.tex
    gregorio -o test.tex test.tex

Fails, and wrecks test.tex.

henryso:
I wasn't sure I should guard an explicit file name in the case of doing some weird stuff with pipes.  It's probably unlikely anyone is doing that, but do you think I should extend this check to explicitly specified file names?

eroux:
@henryso : maybe it would be easier if there was a `--local` and `--global` option that would be shortcuts for `var:TEXMFHOME` and `var:TEMFLOCAL`? That would ease the invocation for most users...

@eschwab I don't believe "C" is a relevant label here, it has nothing to do with the C language... maybe "packaging" or "scripts" would be more relevant?

henryso:
I didn't realize that checking something in would add to my existing pull request.  I'll see if I can fix it, if only as an exercise for myself to get it right.  If I can figure it out, I will close this this request.

henryso:
Breaking into two individual pull requests.

Issue #99, opened 2015-03-18
Defining Changelog guidelines


eroux:
The changelog we have now in release-3.0 branch is extremely confusing for me, so I cannot imagine what users will understand when they'll look at it...

A few guidelines I propose (already discussed a bit [here](https://github.com/gregorio-project/gregorio/commit/29cc2e7b22801cc31405bc0c85a4c5289dc9e4c1#commitcomment-10255918). Please remember [these guidelines](http://keepachangelog.com/), and let's take an example on what I think can be considered as a good Changelog (though too big for us) [here](https://wiki.documentfoundation.org/ReleaseNotes/4.4).

A few thing I find extremely confusing in the Changelog:

 * if version 2.4.3 is yanked, why keep it in the Changelog at all? Ok, there's been confusion in the release cycle, but what does a yanked release mean? Are the changes listed in the changelog under 2.4.3 removed? Or are they in 2.4.2 or 3.0.0? If I ask myself these questions, users will... so I propose to yank it from the Changelog too, and to pass the relevant entries in 3.0.0. Also, remove "Fixed" entries if they relate to problems introduced in 2.4.3 (because it's another point that is ambiguous in the current state).
 * Changelog is for users, and something I'm 100% sure about, is that most users of gregorio will be confused about the following sentence: `Syllables were not being spaced correctly (see #79). This appears to result from the space calculations not terminating properly. There's a need for a \relax somewhere. Since the problem didn't show up in debug mode, I simply added a \else\relax to \gre@debug to eliminate the problem. We may need to go back and fix this better later.`. Apart from making people anxious, I think noone understands the technical description, and not many users (if any) care about this...
[...]

PR #100, opened 2015-03-18
Rewriting CHANGELOG entries for improved clarity

rpspringuel:
Based on discussions in #99 and on commit 29cc2e7.

Hopefully these are somewhat clearer.  Any advice on improving them further?


eschwab:
The section on 2.4.3:
- according to KeepAChangeLog the line should be: 2.4.3 - 2015-03-04 [YANKED]
- The `Deprecated` and `Changed` sections do not need to be there.

henryso:
I think the explanation for \includescore is too confusing.  This feature is probably too complex to fully document here, so maybe a pointer to the actual documentation is in order.

Some spelling errors:

- Solesme's should be Solesmes or Solesmes's
- bistroph should be bistropha
- tristroph should be tristropha

I believe Élie has said (and I agree) that line 48 (Fixed deprecated...) should be dropped.

This leaves one item ("This CHANGELOG") in 2.4.3.  I propose moving that up into 3.0.0, dropping the subsections below 2.4.3, and also removing bug fixes for bugs introduced in the development of 2.4.3 and 3.0.0 from the change log.

rpspringuel:
Okay, that cleans things up a bit more.  Any other suggestions?

eroux:
Sorry, maybe the yanked release should be kept... I also like the idea to make it one-line only... Also, a more complete documentation for includescore (as well as documentation for new features like <eu>, etc. but that would be my job) maybe should be in the migration rules, or a (precise) link to the documentation... What do you think?

We're really in the middle of the process of improving all that, so don't be affraid to make propositions and making mistakes, that's part of the process!

rpspringuel:
I think the migration rules (in whatever form they take) should focus on obsolete and deprecated functionality and how to take old projects which make use of such features and change them so that they'll work with the upgraded package.  New features might be mentioned (as a user might have been using a workaround for something that can now be done more directly) but a full description should probably be reserved for the UserManual, since it's not technically necessary to make any changes to an old document regarding these features.

On the issue of linking the documentation, when I ported the UserManual out of texrenaming to employ it to document my changes to the spacing code, I cut out some structural components that would have been empty because they weren't needed for the spacing documentation.  Perhaps we should restore that and settle on an outline for UserManual structure so that it's easier to point at a particular location within it in a stable fashion.

eroux:
(This time commenting @rpspringuel 's changes) this is much better thank you! The main improvement left (which is interdependant with documentation, migration guide, etc.) is to find the right balance between conciseness and preciseness. What I propose is to make descriptions a bit shorter in the changelog, and more "marketed" (I'm not sure it's the right term), so that users can see what use they can make, and put a link to a longer description. For instance,  I would replace:

`All distances now scale with staff size changes by default. Further, any distance set by the user can have this behavior changed by use of `\grenoscaledim`, and `\grescaledim`. (Previously some distances always scaled, others never did, and there was no systematic way of knowing which was which.)`

by 

`Possibility to switch on/off scaling of distances with staff size, as a consequence, it is now possible to define distances in em units (scaling with text font), see xxx for details.`

or something like that (if I understood the feature well), I think users can see more directly how they can relate to the feature with this phrasing... Also, as it is shorter, people will be less affraid to read the changelog. What do you think?

rpspringuel:
Your wording is definitely more concise and it manages to mention the two most significant consequences of the change, whereas I had only mentioned one.

It does make it so that after reading the CHANGELOG you don't yet know how to make use of the new functionality but with a good link, someone who wants to use it will know where to find the information they need.  And as I think on it, someone who is not interested in using the new functionality just needs to be made aware of its existence, not how to use it.

I probably won't be able to return to working on this until Friday, but I'll see if some of the other entries can't be modified in a similar fashion.

rpspringuel:
Other than completing the links (which is dependent on completing the documentation) I'm coming up blank on any further clarifications here.  Unless someone has a suggestion, I think this is ready to merge.  We can revisit the links once the documentation is updated.

henryso:
Very minor, but I think "setable" should be spelled "settable".

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
-- User settable distances now stored as strings (see [#50](https://github.com/gregorio-project/gregorio/issues/50))
-- `\gre@stafflinefactor` now uses same scale as `\grefactor`
-- Improved `\includescore` to remove outdated files ([#61](https://github.com/gregorio-project/gregorio/issues/61)).
-- TeXworks configuration script for Windows updated
+- Unify version numbers. The `gregoriotex_api_version` has been deprecated. The only version number is the gregorio release version. This simplifies the versioning process and should make bug reports easier to address.
+- Distances in default space configuration (`gsp-default.tex`) have been changed so that they represent the actual printed dimension at the default staff size (see [#50](https://github.com/gregorio-project/gregorio/issues/50)).
+- User settable distance names have been simplified by removing the `\gre` prefix.  See doc/UserManual.pdf section ?? for a full list of the distance names.
+- Possibility to switch on/off scaling of distances with staff size, as a consequence, it is now possible to define distances in em units (scaling with text font), see doc/UserManual.pdf section ?? for details.
```
eroux: It seems redundant with following one to me... maybe just incorporate the link to user manual in next one?

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
+- `\grechangedim` now takes three arguments: the name of the distance, value to change the distance to, and whether or not this value should be scaled with changes in the staff size.
+- `\setinitalspacing` now takes four arguments.  New argument specifies whether distances should scale when the staff size changes.
+- `\setspacebeforeinitial`, `\setspaceafterinitial`, and `\setaboveinitialseparation` now take two arguments.  The new argument specifies whether the distance should scale when the staff size changes.
```
eroux: This could be merged with the previous one, reformulating to "now take an additional argument: ...", wdyt?

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Improved `\includescore` capabilities.  See doc/UserManual.pdf section ?? for full details.
+- `\gre@stafflinefactor` (the internal control for the thickness of the stafflines) now uses same scale as `\grefactor`.
```
eroux: I don't believe this is necessary, I'm not sure I really understand myself, let alone users... if someone hacked the thickness of the rules while it's not documented, no doubt they'll find out how to update their code...

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
 - Windows installer documentation updated
-- Clivis stem length now follow Solesme's books conventions (see [#31](https://github.com/gregorio-project/gregorio/issues/31)).
-- Space Configuration Loading code refactored
-- Most distance calculations disentangled so that each calculate functions do not call each other (helps to avoid circularity)
-- `\grefactor` is now initialized at 17 (the default value).  Necessitated that most distances be rescaled (see [#50](https://github.com/gregorio-project/gregorio/issues/50)).
-- Distances renamed to `\gre@...`  Functions which calculate distances renamed to `\gre@calculate@...`
-- All code which defines or calculates a space (and nothing else) now reside in gregoriotex-spaces.tex
+- Clivis stem length now follow Solesmes' books conventions (see [#31](https://github.com/gregorio-project/gregorio/issues/31)).
+- TeXworks configuration script for Windows updated.  `greg-book` and `gregorio` engines will no longer be added.
```
eroux: I would replace "be added" by "be available". Also, maybe a little explanation on the reason why they won't be available?

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
 
 ### Fixed
-- Fixed deprecated code in autoconfig.ac (see [#57](https://github.com/gregorio-project/gregorio/issues/57)).
-- `\includescore` not finding files for autocompile under certain circumstances.
+- `\includescore` not finding files for autocompile under certain circumstances.  (see http://www.mail-archive.com/gregorio-users@gna.org/msg02346.html and following)
 
 ### Added
+- `\gresetdim` for setting distances in space configuration files.  The TeX primitive notation (`\somedistance = 3cm`) can no longer be used.  Takes three arguments: the name of the distance, the desired value, and whether the distance should scale with changes in the staff size or not.
```
eroux: From the user point of view, this is a change, not an addition, as this means that they'll have to change their `gsp-xxx.tex` files... I would turn the sentence more like "your curstom space files (`gsp-xxx.tex`) need to be updated, see `gsp-default.tex` for a model.

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
+- `\setstafflinethickness` controls the thickness of the staff lines.
```
eroux: A link to the user manual would be useful here I believe

CHANGELOG.md
```diff
@@ -5,59 +5,41 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
-- `\grefactor` is now initialized at 17 (the default value).  Necessitated that most distances be rescaled (see [#50](https://github.com/gregorio-project/gregorio/issues/50)).
-- Distances renamed to `\gre@...`  Functions which calculate distances renamed to `\gre@calculate@...`
-- All code which defines or calculates a space (and nothing else) now reside in gregoriotex-spaces.tex
+- Clivis stem length now follow Solesmes' books conventions (see [#31](https://github.com/gregorio-project/gregorio/issues/31)).
+- TeXworks configuration script for Windows updated.  `greg-book` and `gregorio` engines will no longer be added.
 
 ### Fixed
-- Fixed deprecated code in autoconfig.ac (see [#57](https://github.com/gregorio-project/gregorio/issues/57)).
-- `\includescore` not finding files for autocompile under certain circumstances.
+- `\includescore` not finding files for autocompile under certain circumstances.  (see http://www.mail-archive.com/gregorio-users@gna.org/msg02346.html and following)
```
eroux: I would just use the markdown notation to make the link under "this thread" instead of having the whole link visible

CHANGELOG.md
rpspringuel: I don't have a problem with adding a brief explanation on why, but I think "be added" should stand because the configuration script specifically adds them to TeXworks.  The bat scripts are still available in that they won't be removed from an existing TeXworks install and they still live in the contrib folder.

eroux:
@rpspringuel : do you want me to merge your request and make a few changes? Maybe this will be faster and less boring for you?

CHANGELOG.md
eroux: Why not. Also, why only under Windows? Are the engines still relevant for Linux and OSX? I'm asking that because I'm currently fixing `make dist`, and I'm blocked by this (the Makefiles are not up-to-date, and I'm not sure what to remove)

rpspringuel: There never was a TeXworks configuration script for Mac or Linux.  The engines had to be manually configured on those OSes.

My thoughts were to leave the engines in the contrib folder for the sake of backwards compatibility through the version 3 release cycle.  When we hit version 4, we update their status from Deprecated to Obsolete and drop them from the source.  This way the transition is eased for users who might not be willing to go all out changing their projects right away.

rpspringuel:
I don't mind making the changes myself as it helps me figure out what's proper to this file and what isn't.

I, too, was thinking of merging in the NEWS file, but was going to wait until after this was taken care of.  I can add it now, though, if you want.

eroux:
If you don't bother yes, that would be good to merge NEWS now.

CHANGELOG.md
eroux: Ok then, I'll update the makefiles accordingly

rpspringuel:
Okay, I've resolved conflicts and gotten this ready to merge again.  At this point it's probably a good idea to merge this and do any further modifications in a later pull request.

PR #104, opened 2015-03-20
Inconsistent baseline for alt text

rpspringuel:
Forgot a `\global` when initializing the temp dimension used to calculate the position of the baseline for the alt text. This resulted in it carrying in a value used elsewhere to start instead of resetting it to the user dimension `abovelinestextraise`. The `\global` is necessary because the initialization happens inside the `\greabovelinestextstyle` so that em and ex units are respected.


henryso:
Tested. Seems to work fine.

Issue #105, opened 2015-03-20
gregoriotex.lua fails to load


henryso:
! LuaTeX error .../share/texmf-dist/tex/luatex/luatexbase/modutils.lua:52: modu
le 'gregoriotex' not found:
	no field package.preload['gregoriotex']
	[luatexbase.loader] Loading error:
	/usr/share/texmf-site/tex/luatex/gregoriotex/gregoriotex.lua:192: ambiguous sy
ntax (function call x new statement) near '('
	[kpse C searcher] file not found: 'gregoriotex'
	[oberdiek.luatex.kpse_module_loader] Search failed
stack traceback:
	[C]: in function 'require'
	.../share/texmf-dist/tex/luatex/luatexbase/modutils.lua:52: in function 'requi
re_module'
	<\directlua >:1: in main chunk.
\lltxb@requirelua ...xluaescapestring {#2}" \fi )}
                                                  
l.73

PR #106, opened 2015-03-20
Fix for #105.

henryso:


eschwab:
Sorry, my text editor sometimes does silly things for line wrapping.

Issue #101, opened 2015-03-18
Version changes and score compiling.


eschwab:
Since adopting SemVer for gregorio, the user should only be required to recompile scores if the major version changes `x+1.y.z`. All minor and patch releases must be backwards compatible and thus all scores with version `x.y.z` must work when gregoriotex is in version `x.y+1.z`. The user can choose to recompile if there are new features (in `gregorio-write.c`) he wishes to use.

Current behavior is that a score must be recompiled for each version change (major, minor, or patch).

eschwab:
If a user is using the autocompile feature I am inclined to leave the lua script as it is, which is to compile a new gtex file if its version does not match the gregorio version. A new gtex file for any version change: `x+1.y.z` `x.y+1.z` and `x.y.z+1`. My thought is that a user running `autocompile` wants the best version possible. If there are bug fixes in `gregorio-write.c`. hence `x.y.z+1`, no manual intervention will be necessary to apply the fix to the gtex file.

What needs to change is `\gregoriotexapiversion` which is the final version test. This effects users running the default behavior of gregoriotex, `nevercompile`.

Thoughts?

eroux:
If I understand correctly, you would make `nevercompile` output an error only if major version is different, and let `autocompile` recompile at each version change? If so, why not

eschwab:
Yes, that is a simpler way of saying it.

eschwab:
Fixed by #107

PR #107, opened 2015-03-20
Changed version test to only fail on mismatched major versions.

eschwab:
Fix for #101 

If a score declares that is it for version `x.y.z` the score will only
need to be recompiled if the major version of gregorio changes,
i.e. `x+1.y.z`.

This does not effect the `[autocompile]` options. When `autocompile` is
`true` a new score will be compiled for any version change.

### DIFF
--- a/tex/gregoriotex.tex
+++ b/tex/gregoriotex.tex
@@ -94,8 +94,8 @@
 % #1 is the version of GregorioTeX the score is made for.
 
 \def\gregoriotexapiversion#1{%
-  \StrBefore{\gregoriotexversion}{.}[\gre@temp]%
-  \IfBeginWith{#1}{\gre@temp}{\relax}{%else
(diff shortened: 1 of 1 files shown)


rpspringuel:
Can you use something other than `\gre@temp` for storing the string slice?  I find that macro name to be so generic that its bound to create conflicts at some point if we allow it.

For instance, at one point during my rewrite of the spacing code I was using a skip register with that exact name.  I ended up renaming it, so there's no conflict for you to run into now, but down the road...

Maybe something like `\gre@majorversion`?  It's less generic and tells you a little about what's getting stored inside.

eschwab:
It is worth renaming `\gregoriotexapiversion` since we no longer have a gregoriotex api version?

rpspringuel:
While we don't have an API version number, per se, we do still have an API of sorts and that can only change with a change in the major version number, so I don't think the function name needs to change.

rpspringuel:
Checks out for me.

PR #108, opened 2015-03-21
Check to avoid accidentally clobbering the input file. Fixes #95

### LINKED ISSUE #95: Gregorio should abort if the output filename is the same as the input filename

henryso:

### DIFF
--- a/contrib/gregorio-scribus.lua
+++ b/contrib/gregorio-scribus.lua
@@ -79,7 +79,7 @@ end
 
 local format = string.format
 
-texfile:write(format("\n\\includescore{%s-score.tex}\n\\end{document}\n", pathbase))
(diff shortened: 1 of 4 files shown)


eschwab:
I reread #95 and this does address that. Additionally though it would make sense to me to add another check for an explicit `gregorio -o file.tex file.tex` since that will most likely fail everytime and ruin `file.tex`, possibly losing work.

henryso:
I somehow closed this and I can't reopen it, so I will create a new pull request once I've made the change suggested by @eschwab.

PR #109, opened 2015-03-21
Removed install rules from fonts Makefile and create separate install script. Fixes #97

### LINKED ISSUE #97: Should "install" from the fonts Makefile depend on the fonts?
Currently, the .ttf fonts are built from the .sfd files in the fonts directory and then are committed into the repository.  This is fine because only someone working on the fonts would be modifying them.  That person would then presumably commit the generated fonts.

However, if someone clones the repository from scratch, chances are the time stamps on the .sfd files will be the same or newer than the timestamps on the .ttf files.  Someone running make install just to install the files into their TeX directory will then be forced to rebuild the .ttf files.  It also means that the person installing Gregorio must have python, fontforge, and the python fontforge bindings.

The cleaner solution would be to have two Makefiles.  One in the fonts directory, whose only purpose is to make fonts, and one in some other directory, whose purpose is to install the GregorioTeX files.  This latter Makefile would fail if, for some reason, the .ttf files are missing, but it would not try to rebuild them at that point.

Alternately, we can rejig the current fonts Makefile to have two targets: one for creating fonts and one for installing GregorioTeX.  This gets a bit messy (and is why I say the previous is cleaner) because the "make the fonts" target needs the sfd-to-ttf dependency while the "install" target must not have it lest the fonts be rebuilt unnecessarily.

I suppose there is a third option: add `touch *.ttf` after `cd fonts` in the installation instructions to be sure they have a later timestamp than the .sfd files.

I can make the Makefile changes and submit a pull request once we decide if/how we want to handle this.

henryso:


henryso:
f8af32c somehow got into this pull request.  Closing to fix.  (I will re-add the build/install label once I've gotten it straightened out)

PR #110, opened 2015-03-21
Moved install logic from fonts/Makefile to install-gtex.sh. Fixes #97

### LINKED ISSUE #97: Should "install" from the fonts Makefile depend on the fonts?

henryso:


rpspringuel:
I like @eroux's comment from the previous thread:

> maybe it would be easier if there was a --local and --global option that would be shortcuts for var:TEXMFHOME and var:TEMFLOCAL? That would ease the invocation for most users...

I find using `--local` in a way that doesn't map onto `TEXMFLOCAL` confusing though, so I might use option names like `--home` and `--local` or `--user` and `--system` or even `--me` and `--everyone`.

henryso:
I was about to comment on having the same concerns about using the word local.  I was going to suggest system-wide and personal, but I like your suggestion of user and system much better.  I also see no need to have the dashes.  The syntax would be:

```
install-gtex.sh system
```

and 

```
install-gtex.sh user
```

I will push these changes shortly.

eroux:
Just a final request if I may, can you please add -j* (or even replace --jobs=* by -j*)? Also, I wanted to add some [hardening flags](http://security.stackexchange.com/questions/24444/what-is-the-most-hardened-set-of-options-for-gcc-compiling-c-c) in the build script, but that can be for later...

henryso:
I made it take `-j<N>` in addition to `--jobs=<N>`, but I can't make it accept `-j <N>` (with a space) without an overhaul to the existing option parsing logic.

eroux:
It's perfect, that's the way it's usually called with make

eroux:
I need to apply

```
diff --git a/install.sh b/install.sh
index d2fed79..d203378 100755
--- a/install.sh
+++ b/install.sh
@@ -18,7 +18,7 @@ make install || die "install Gregorio"
 echo
 
 echo "Installing GregorioTeX"
-install-gtex.sh "${1:-system}" || die "install GregorioTeX"
+./install-gtex.sh "${1:-system}" || die "install GregorioTeX"
 echo
 
 echo "Installation complete."
```

before being able to `./install.sh`, can you please apply it (if it works on your system)?

henryso:
Serves me right for having "." in my PATH.  Fixed.  Thanks for finding it.

Issue #69, opened 2015-03-08
Add Windows Metadata to gregorio.exe


eroux:
See [this post](https://www.mail-archive.com/gregorio-users@gna.org/msg02312.html) for the request, and [this](http://stackoverflow.com/questions/708238/how-do-i-add-an-icon-to-a-mingw-gcc-compiled-executable) and [that](http://gnu-automake.7480.n7.nabble.com/correct-windres-use-td4889.html) for hints.

Also, the icon could be added to the main `.exe` file, with the same process.

Low priority.

eschwab:
Addressed by #116

PR #111, opened 2015-03-21
updated Windows Gregorio icon to contain all the size formats.

BGMcoder:


eroux:
Thank you very much! Well, I realize I don't even know how to test this... I think I'll build a Windows installer with this icon, and send it to you for testing. @BGMcoder what do you think?

BGMcoder:
Elie,

The icon is really for the gregorio.exe, not for the installer.

I'd like to try my hand at compiling gregorio for windows, 
then I could, perhaps, help with the installer too.  I have 
and use INNO myself already.

But I have PellesC - a native C compiler, which compiles the 
.c files and the .h files.  I'm not terribly familiar with 
it, but have compiled one or two things in the past.  
Problem with that is that I'm not sure where to start, as 
I'd have to build a new compilation script.

eroux:
Ok, let's merge it then, it won't hurt. But I'm not sure gregorio.exe has an icon associated to it... does it? I think this one is used by the installer...

eroux:
About compiling, I believe trying to compile with something else than the build system it's usually build with will give you lots of trouble, require to learn many new things, and take you an incredible amount of time... If you really want to build gregorio under windows, the easiest (and cleanest) way to go is to use [mingw](http://www.mingw.org/), one you'll have it installed and running (I'm not saying it will be easy), you'll be able to compile gregorio in a very straightforward way. Otherwise you'll need to make a build script handling tons of options for linkage, etc. and if you don't know what linkage means, this will take quite a lot of efforts... Also, as it's not really necessary for Gregorio, I'm not willing to spend hours on this...

BGMcoder:
Elie,  no, at the moment, Gregorio.exe does not have an 
icon.  Only the installer has one.

eroux:
Adding the icon to gregorio.exe will be in the same process as https://github.com/gregorio-project/gregorio/issues/69

By the way, learning mingw, autotools, etc. (again, not saying it's easy) will possibly allow you to make the modifications in the build process to add the metadata. If you manage to add them with a new build script using you windows-only compiler, then it means that we depend on windows for the windows build, and it does'nt seem like a good direction to me...

BGMcoder:
Elie,

I"m installing mingw and will try it out.  Thanks!

eroux:
I think I've just saved you a lot of time with #113

BGMcoder:
Haha, Elie.  It would taken me a week to figure that out!  
Thanks!

eroux:
I think you're a bit optimistic :)

PR #112, opened 2015-03-21
Check to avoid accidentally clobbering the input file. Fixes #95

### LINKED ISSUE #95: Gregorio should abort if the output filename is the same as the input filename

henryso:
See older comments in #108 and #96.


eschwab:
Builds and runs fine.

PR #113, opened 2015-03-21
add Metadata (and icon) to windows executable

eroux:
This fixes #69 and also updates VERSION (more to come on this)

PR #116, opened 2015-03-21
fix #69

### LINKED ISSUE #69: Add Windows Metadata to gregorio.exe

eroux:
resources (including icon) are in windows/gregorio-resources.rc, compiled by
windres (if cross-compiling only) into gregorio-resources.o, linked with
the main executable (in src/).

### DIFF
--- a/VersionUpdate.py
+++ b/VersionUpdate.py
@@ -13,6 +13,7 @@ import re
 version_file = 'VERSION'
 
 gregorio_files = ["configure.ac",
+                  "windows/gregorio-resources.rc",
                   "windows/gregorio.iss"]
 
 gregoriotex_files = ["tex/gregoriotex.lua",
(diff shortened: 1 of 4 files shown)


eroux:
This replaces #113

eroux:
I think I'm the only one who can test this, hence the absence of comments... merging.

rpspringuel:
If I understand this correctly, I can test it, but I've been too busy to do so recently because it requires going through the whole build process for Windows and my personal scripts for that process need some attention to update them to the new branching model.  Sorry for not saying something.

PR #119, opened 2015-03-22
reorganizing documentation files

eroux:
* README -> README.md and updating
 * THANKS and AUTHORS -> CONTRIBUTORS.md and updating
 * adding CONTRIBUTING.md


eschwab:
My email should be `elijahschwab` rather than my old address `rcsniper`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+ * [git](http://git-scm.com/)
+ * the [GNU Build System](http://airs.com/ian/configure/) (often referred to as "Autotools")
+ * [gcc](https://gcc.gnu.org/) (and associated tools)
+ * [GNU indent](https://www.gnu.org/software/indent/)
+
+### Coding style
+
+The code uses two spaces for indentation. The C code follows the [GNU coding standards](https://www.gnu.org/prep/standards/html_node/Writing-C.html).
+
+Use indent on your code before commiting it, with the `.indent.pro` file in repository's root folder.
```
rpspringuel: Perhaps a little more detail here?  Or at least a link to a page explaining how to do this?  I have to be honest that I'm one of the ones guilty violating this (2 spaces just isn't enough for me to see the indention structure) and need to implement the conversion process to clean up my code.

eroux: just run `indent x/y/z.c` from the root directory and that's it. I'll document it

eroux:
New version ready for proofreading

henryso:
What does `Gregorio is the reference software for Gregorian chant scores engraving.` mean exactly?

CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+ * the [GNU Build System](http://airs.com/ian/configure/) (often referred to as "Autotools")
+ * [gcc](https://gcc.gnu.org/) (and associated tools)
+ * [GNU indent](https://www.gnu.org/software/indent/)
+
+### Coding style
+
+The code uses two spaces for indentation. The C code follows the [GNU coding standards](https://www.gnu.org/prep/standards/html_node/Writing-C.html).
+
+Use indent on your code before commiting it, with the `.indent.pro` file in repository's root folder: run `indent path/to/my/file.c` from the root directory.
+
```
rpspringuel: When I try this (on `lib/characters.c`) I get an error: `indent: .indent.pro: unknown parameter "//"`  I have no idea if this is related to me being on a Mac or not.

eroux:
I'm not sure what part you're asking for clarification on, but "reference software" means (if I translated "logiciel de référence" from French correctly) that it is the reference in terms of this kind of software, meaning that it seems to be higher than other softwares in the area (which is simple because there aren't many...). "score engraving" means the same as [here](http://www.lilypond.org/introduction.html). Or was it another part that wasn't clear?

CONTRIBUTING.md
eroux: Strange, can you please give the result of `indent --version` (I have `GNU indent: 2.2.11`)

rpspringuel: Hmm... indent isn't even recognizing the `--version` option, so I'm guessing that it is indeed related to the Mac implementation.  Let me see if I can't get the GNU version installed and I'll try again.

eroux: OSX indent is certainly the BSD indent... well, I'll document that the GNU indent is needed (it has more options)

henryso:
I think "reference software" doesn't translate properly from French.  [Wikipedia](https://en.wikipedia.org/wiki/Reference_software) says reference software is "software which emulates and expands upon print reference forms including the dictionary, translation dictionary, encyclopedia, thesaurus, and atlas."

I've heard the term ["reference implementation"](https://en.wikipedia.org/wiki/Reference_implementation) but I don't think that's exactly right, unless you mean the reference implementation of engraving score from gabc, or somesuch.

Two possibilities make sense to me here, depending on connotation: ["de facto standard"](https://en.wikipedia.org/wiki/De_facto_standard) or "leading".

"Gregorio is the de facto standard software for engraving Gregorian chant score on a computer."

or

"Gregorio is the leading software for engraving Gregorian chant score on a computer."

... though stating that in so many words feels a bit self-aggrandizing.

CONTRIBUTING.md
rpspringuel: Okay, I have GNU indent installed and it worked fine.

I might point out, that it did make several changes to lib/characters.c.  Perhaps we need to run it on all files to update things?

Another thing, `indent` is only for the C files.  Do we want to follow the same indentation rules for the TeX files?  If so it might be helpful to have a link to [this page](http://stackoverflow.com/questions/2316677/can-git-automatically-switch-between-spaces-and-tabs) which shows how to implement an automatic change-over between tabs and spaces in git.

rpspringuel:
I think @henryso is right about the wording, and also that it is a bit self-aggrandizing to say it about ourselves (even if it happens to be true).

rpspringuel:
Okay, I'm looking at indentation in the various files and noticing that we really should nail something down for all file types:

1. Python: the standard here from [PEP 8](https://www.python.org/dev/peps/pep-0008/#indentation) is 4 spaces.  I think all or our python scripts conform to this.
2. Lua: According to [the style guide](http://lua-users.org/wiki/LuaStyleGuide) the standard is 2 spaces, but there's flexibility.  The scripts of ours that I checked seem to use 4 spaces.
3. C: as @eroux established above, the standard (using `indent`) is 2 spaces.
4. TeX: As best I can tell, there are no standards here, de facto or otherwise.  I've always used a tab, but that's because I like whitespace (and thus clearly marked indentation levels).  Right now we're seem to have a mix of styles in these files (probably because I've been using tab but most everyone else is using 2 spaces).

Those are the ones I thought of real quick.  Are there others?  Once we nail something down, I'll go through and fix the existing files.

Note: I actually like the consistency between Lua and Python because I use the same editor for both and already have it setup so that tabs automatically expand to 4 spaces (the Python standard).

eroux:
Oh, ok, I have to admit I used this wording a few times (in other contexts), hopefully people in front of me were not native speakers... But that does'nt mean they understood... If it's too self-aggrandizing (another wording I didn't know, though it sounds a bit fancy, like "aggrandir" in French), then we can just skip this part and say something like "Gregorio is a software for engraving Gregorian Chant scores on a computer".

About indentation, I was about to put an [`.editorconfig` file](http://editorconfig.org/) that would set an indent of 2, but you're right, we can make an indent of 4 for python scripts. What do you think? I like this system more and more, especially since in my job I have to use 3 space indentation, which I never use in any other project, so have a per-project file seems useful... And setting it to two different things for Lua and Python would make your editor already configured (almost all editors have an editorconfig plugin). About TeX, I fear we'll have to do this by hand...

henryso:
Fun fact: it seems the English word "aggrandize" comes from French (and ultimately Latin), so I think it means exactly what you think it means.  I'm in favor of going with "Gregorio is a software for engraving Gregorian Chant scores on a computer".

eschwab:
It makes sense to me to keep python, lua, and TeX all at 4 spaces indent. I think I can change the TeX files automagically, at least for the most part.

rpspringuel:
I'm playing around with smudge and clean filters for git and seem to be getting somewhere.  Unfortunately my computer's Internet connection has decided to go wonky and it's already late for me so I won't be able to upload changes tonight.  Hopefully the problem will be gone tomorrow and I'll be able to post something.

eroux:
Who does also have a preference for indentation? My personal preference is 2, but if 4 wins, that's ok for me...

rpspringuel:
So, my internet connection is working this morning.  Check commit: 1e6d9ae

That commit changes all the tab indentations that have been introduced into the TeX files and switches them to 2 space indentation.  However, when the files open on my computer, I see tabs and can edit them as I want.  I did this through the use of smudge and clean filters, and thus it should be generalizable to anyone's set up, allowing you to have your personal preference on your computer but still submit code that conforms to the standards.  (No pull request because we haven't settled on an indentation standard for TeX files yet.)

Smudge and clean filters, however, are inherently a personal setting on your computer.  You'll notice that the commit doesn't add any control files to the repository.  Would it be a good idea, therefore, to mention this possibility in our contributing guidelines and provide some links for how to set it up?

eroux:
@rpspringuel thanks! I don't think it really belongs to this pull request, but if it's just indent, I think you can commit directly to `develop`... what do you think?

rpspringuel:
I can do that easily enough.  I just need to know I'm committing the right indentation standards.  Are we going to use 2 spaces for TeX?

eroux:
I'd be in favor of 2 yes... what do you personally think? @eschwab is in favor of 4... @henryso what do you think?

rpspringuel:
My personal preference is for tabs, but thanks to the filters I don't really care. It's just as easy to set the filters to convert from tabs to 2 spaces as 4.

eschwab:
For TeX 2 is fine with me. That is the default in my text editor anyways.

eroux:
ok, let's make that 2 for TeX, 4 for python... what about Lua?

henryso:
@rpspringuel You might want to contribute the smudge/clean control files (somehow) to the contrib directory.  It's for developers, though, so maybe that's not appropriate.  If that doesn't make sense, then I agree with mentioning it as an option for compliance within the guidelines.

@eroux As a *personal* preference, I generally like 4 spaces because I can see the indentation better, but for languages that tend to nest deeply, I switch to 2.  For C, I don't particularly like GNU-style indentation, but since we are using it (i.e., with indented curly braces), then 4 spaces would probably drive me crazy.  In the end, I try to follow the file I'm editing and will do whatever is decided here.

eroux:
To be honest, I don't like gnu style indentation anymore neither (especially indented braces). Can you suggest some changes in `.indent.pro` ?

rpspringuel:
@henryso I'll figure out some way to do that.  It may have to take the form of a set of instructions, but even that's better than figuring out the whole thing from scratch.

@eroux on Lua indentation I'd say we stick to the style guide.  While I haven't figured out the best way to convert from 2 space indentation to 4 space indentation and back again yet, I'm sure that I can figure out a filter which will automate the process for those who prefer the visuals of 4 space indentation.

eroux:
We'll leave the `.indent.pro` changes for another pull request, if it's ok for you, can it be merged soon? I need it for other coming pull requests

henryso:
My personal indentation style for C is "not-exactly-K&R".  But I've dealt with so many indentation styles that most of them don't bother me too much.  If we decide to move away from GNU style, why don't we choose one of the "common" styles (i.e. -kr) available in GNU indent as a starting point?

eroux:
yes, why not, I'll make a pull request later for this

rpspringuel:
Looks good to me.  Merging.

eroux:
Thanx!

Issue #121, opened 2015-03-22
remove compiled files from repository?


eroux:
Having compiled filles (doc/UserManual.pdf, fonts/*.ttf) in the repository seems like a anti-pattern to me, and make it a bit bloated (very long to clone).

The reason why I added fonts in the repository was because:
 * the build process used to be extremely convoluted (needed `perl`, some obscure tools for `ofm`, etc.)
 * due to the extremely slow release cycle, most people were using the svn version, and couldn't build the fonts

But the situation is different now:
 * the build process only requires `python-fontforge`
 * the new release cycle will drastically lessen the need for using the git version

So I propose to remove the compiled files from the repository, but to build and include them in the source code release (`make dist`).

What do you think?

henryso:
I agree this is an anti-pattern.  People using the current `cd fonts; make install` process were likely building fonts anyway.  UserManual.pdf seems easily buildable as well.  I'm fine with removing them both.  Of course, the various build scripts will need to be modified to make sure fonts are built.

eschwab:
Fine with me. The current state of the UserManual is not 100% user friendly, so no harm in dropping the pdf.

rpspringuel:
Are we making the assumption then that if you're installing from the repository then you're rebuilding everything?  I don't think that's necessarily a bad assumption, but we might want to make it clear in the installation guide (i.e. the website) that that's what you're doing.

eroux:
Yes, I think we can make that assumption and document it.

eroux:
what branch should I target for this? develop?

henryso:
If we do this in release-3.0, then the install instructions (on the website) should have separate `make` and `sudo make install` commands for the fonts directory.  If for develop, those instructions will have to change anyway (but to something else entirely).

eroux:
Not really, the distributed tarball will have the generated files (the one made with `make dist`), so make in the fonts directory won't be necessary

henryso:
Ok, so then the instructions will changed to be based on a tarball rather than git.  So no autoreconf.  Fine.

eschwab:
If it cleans up the install process, I'm in favor of doing it in release-3.0. Makes more sense to change the release/install process with a major version release. I know there really should not be new features introduced into release-3.0 but since we just adopted our new git workflow I'm fine with it.

rpspringuel:
[...]

PR #128, opened 2015-03-23
update UserManual to use LuaTeX and add build script and instructions

eroux:


henryso:
This will make building the (developer-focused) UserManual.pdf mandatory every time the application is built.  If you really think that's necessary, then fine, but I think that building the UserManual should be optional.  If there's a way to make building this a configure option, then that's probably the way to go.  If not, I think just giving the build instructions for the manual (since it's a one-liner) is good enough and will allow users to build it or not depending on their needs.

eroux:
I don't really understant what you mean... `make` won't build the `.pdf` file, only `make dist` will do it for the tarball. Does it sound better?

henryso:
My apologies. I misunderstood the Makefile.am.

eroux:
Also, this commit doesn't remove the UserManual.pdf yet... Please check that the new `.tex` and build process (latexmk with some complex options) works for, you, as there might be some escaping issues in the Makefile

rpspringuel:
See my most recent comment in #121 .

rpspringuel:
Opps...  I meant to make that a pull request to this branch (as @henryso had done), not to push directly to it.  Hopefully no harm done.

eroux:
Do you think it can be merged now?

rpspringuel:
Yes, but since @henryso made modifications to `.gitignore` on another branch to do the same thing as mine here, we'll need to watch for conflicts.

eroux:
Thanks!

PR #129, opened 2015-03-23
Fixes for automake / make dist

henryso:
I ran into a few issues while trying to test the build of UserManual.pdf.

1. contrib/Makefile.am was requiring files that haven't existed since November 2009.  They have been removed from the list of extra dist files.
2. The -lualatex option of latexmk is not in TeX Live 2012, which happens to be what I'm using, so I modified the script to hopefully work in more versions of TeX Live.  Please review!
3. The distribution tarball was not being ignored.  Added to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,9 @@ src/gregorio
 #emacs auto gen files
 auto/
 
+#tarball
+*.tar.gz
+
 !UserManual.pdf
 
 #gregorio specific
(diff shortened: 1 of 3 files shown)


eroux:
The first was addressed in https://github.com/gregorio-project/gregorio/commit/bfc3c9f619473d332d9a883e55190be673bf5552 but it'll merge easily I believe. Ok for the other, checking latexmk code, this is strictly equivalent.

henryso:
Just an idle thought: I wonder if cherry-picking 
https://github.com/gregorio-project/gregorio/commit/bfc3c9f619473d332d9a883e55190be673bf5552 and fixing the merge issue (if any) now would improve, have no effect on, or worsen the merge later.

eroux:
I think it's equal... we'll see!

PR #132, opened 2015-03-23
Move VERSION into VersionUpdate.py

eschwab:
VERSION --> VERSION.md


eroux:
To be honest, I tend to regret the good old days where VERSION contained only the version and was used by scripts directly... I don't really see the point of a VERSION.md file... what do you think its use would be?

eschwab:
Nicer formatting? There is no advantage to having VERSION in md. In fact when I first made VERSION I questioned the need for it. The version could just as easily be stored in VersionUpdate.py. I'm indifferent, if you would like an old school VERSION that works for me.

eroux:
Well, maybe having it in VersionUpdate directly would be the good solution... It should just be easy for a script (like https://github.com/gregorio-project/gregorio/commit/c34899c5bd21497a2b0e4d48c6a9b1376372d8ad) to fetch it...

eschwab:
Moved version into VersionUpdate. This should get the version number for the debian build sh:

    grep -oE '[0-9]+\.[0-9]+\.[0-9]+-?[[:alnum:]]*' VersionUpdate.py

rpspringuel:
The only reason for having VERSION as a separate file is to make it easier for scripts to retrieve the version number.  To that end, I would have left it as plain text, rather than converting to markdown, and may have even stripped everything out of the file except the version number (now that there is only one version number, there's less need for additional information).

While I don't object to putting the version number in `VersionUpdate.py` so long as a script which will reliably retrieve the version number is explicated in principle, my personal principle would be for a separate file simply because it'll be more obvious to a novice that `VERSION` contains the version number.  If I've never seen the code before, I wouldn't necessarily know to look in `VersionUpdate.py` to find the current version number, nor would it be obvious that simply running `VersionUpdate.py` won't change the version number automatically.

Which brings me to an idea: perhaps `VersionUpdate.py` could take arguments so that updating the version number is a one step process?  I would foresee the following arguments:
1. `-b` or `--bugfix`: increment the z version number (of x.y.z) and propagate that change.
2. `-e` or `--enhancement`: increment the y version number, resetting z to 0 in the process and propagate the change.
3. `-m` or `--major`: increment the x version number, resetting y and z to 0 in the process and propagate the change
4. `--manual=###`: change the version number to ### and propagate the change (especially useful for marking a beta release, a develop release, or other release with an unusual number)
5. `-h` or `--help`: print the list of options

A no argument call of the script would get the help options.

I don't think that a script can modify itself, so this would probably require a separate `VERSION` file and no mention of the actual version number in `VersionUpdate.py`  It might also be a good idea, in this case, if `VERSION` had a message which explicitly said not to modify the file directly but to use `VersionUpdate.py`

eroux:
@rpspringuel you're definitely right, having the script return the version is definitely the best option! Maybe an explicit `--get-current` would be better than no option, as it's a bit awkward that calling `VersionUpdate` returns the current version...

rpspringuel:
That wasn't quite what I was thinking, but I like that.  I've updated my previous post to list that and to clarify what I actually meant about a no argument call.

eroux:
Oh, ok, it's clearer now. Also, I didn't think about the script modifying itself... Globally, I think what you say would improve the process, but it could be kept for another pull request. Maybe now just:

 * renaming `VERSION` to the more discrete `.gregorio-version`
 * renaming `VersionUpdate` to `VersionManager.py` (as it's a more suitable name if it get all the functionalities you described)

would be enough... what do you think?

rpspringuel:
I wouldn't rename `VERSION` until after `VersionManager.py` (I do like your rename) is modified to output the current version number under some circumstances.  `.gregorio-version` is a hidden file on most systems and so there would be no obvious place to figure out what the version number is after the file is renamed.

eroux:
Ok, yes. @eschwab what do you think?

eschwab:
For this PR I'll add the `get-current` option to the script, rename `VERSION` to `.gregorio-version`, and rename the script. I'll open another issue to add the other options to the script. For now this will give us `VersionManager.py --get-current` as a reliable method to get the current version. That will not change in the future, so any other scripts can depend on it.

eschwab:
I think this is where I want to leave this PR at. `VersionManager.py --get-current` is now an easy way to get the VERSION. It will not change.

rpspringuel:
Looks good, but I can't test right now.  If this is still open I'll do something with it on Wednesday.

Issue #118, opened 2015-03-22
remove translation system?


eroux:
There is a translation system for the messages (errors, warning and help) output by gregorio, but it's never been used in 7 years... maybe we should just remove it? I don't think it's really helpful, and it would simplify the build process. What do you think?

eroux:
adressed by #120

henryso:
I assume that the translations are out of date (if there at all), but how difficult would it be to put the translation system back in when Gregorio rules the world and we have an army of translators?

eroux:
Gregorio uses the GNU gettext system in its very common form, it's very easy to set up... No translations were ever available, and I really don't think they will be necessary one day: with the new `\includescore`, people won't even know the `gregorio` command exists, and will thus never care about the very rare messages it issues sometimes...

henryso:
As long as it's not difficult to put back in, then I agree. Scrap it.

eroux:
Ok, let's wait a few hours for comments, and I'll merge #120

rpspringuel:
If it wasn't being used in the first place, then I'm all for removing it.

eroux:
It would be easier for me to fix the release system (debian, make distcheck, etc.) if I could merge it into release-3.0 too. What do you think?

henryso:
Because of where that was branched, it would pull in some of my changes in develop.  I don't think we should destabilize this release by doing that.  I will try cherry-picking your change against release-3.0 and see if it works against my complex project.

rpspringuel:
I agree with @henryso.  This is background enough that I think it could be included in the release (i.e. it doesn't really constitute a new feature), but it has to be cherry picked in.

eroux:
cherry picking works fine, I did it in my fixdebian branch, so we'll discuss that in the corresponding pull request (tonight I believe)

henryso:
I ran https://github.com/henryso/gregorio/tree/test-merge-remove-nls through my large project and the https://github.com/gregorio-project/gregorio-test test cases, and it seems so be fine.  I'm fine with cherry-picking this into release-3.0.

eroux:
Done in #133

eschwab:
NLS removed with #120

PR #133, opened 2015-03-23
Fixes source and debian release system

### LINKED ISSUE #90: 3.0 Upgrade Manual
An upgrade manual needs to be added to the doc folder which provides instructions for how to get old projects to work with version 3.0.  This is mostly for me, since it's the distance changes that require the biggest adjustments, but any non-backwards compatible changes should be mentioned here.

### LINKED ISSUE #122: remove OpusTeX and GregorioXML support?
I don't think these two are used by anyone, I think we should drop them in order to reduce the code base and make it less frightening... What do you think?

### LINKED ISSUE #127: Add build changes to CHANGELOG.md after release-3.0 is merged into develop
This is just a reminder for me so I don't forget to do it.

eroux:
This pull request includes:
 * the changes of #109 
 * remove the NLS system (see #118 )
 * integration of the `fonts/` directory in the main build system
 * various fixes for the `make dist` target (`make distcheck` now runs ok)
 * updates in the files for Debian
 * adding a small script to build Debian packages
 * remove outdated `cygwin/` and `Doxyfile`
 * fix #122 by deprecating OpusTeX and GregorioXML
 * fix #127
 * fix #90 by adding UPGRADE.md


.gitignore
```diff
@@ -81,21 +81,26 @@ src/gregorio
...
-!UserManual.pdf
+#Debian files
+*.changes
+*.xz
+*.dsc
+*.deb
 
 #gregorio specific
-
+doc/UserManual.pdf
```
rpspringuel: I don't think this line is needed.  Doesn't line 48 already exclude all `.pdf` files?

eroux: fixed

rpspringuel:
I don't spot any other problems, but since I don't use Debian, I'm probably not the one to evaluate all these changes properly.

olivierberten:
Please move the `debian` directory to `debian-src` or into `contrib` so that debian/ubuntu packagers could use the release package as is.
Right now I need to modify the original source package in order to distribute [my debian packages](https://launchpad.net/~olivier-berten/+archive/ubuntu/gregorio) since debian packagers need to sign their package, and therefore modify the debian/changelog.

eroux:
@olivierberten did you try the new `./build-deb.sh` script (new in this PR)? It does what you want I believe... Tell me if it doesn't! 

We're also discussing this on #134 if you want, and we'll probably move `debian/`, `macosx/` and  `windows/` to a `packaging/` directory (name may change). 

I didn't know you had a PPA, I just created the Gregorio team on Launchpad, maybe it would be easier to use it, so that it doesn't rely on you alone and others can help for the maintenance and updates? What do you think?

eroux:
About changing `debian/changelog`, currently there is a part you need to do by hand... but we could use `VersionManager.py` to do things such as  check the latest `debian/changelog` entry and see if it matches current version, if not, write an entry at the top. For a potential `gregorio-git` package, I don't really know how changelog should be handled... we could make `VersionManager.py` have a `--debian-version` that would return the commit hash or something like that... How could we ease the process for you?

eroux:
Taking a look at how [vlc](https://launchpad.net/~videolan/+archive/ubuntu/master-daily/+packages) does its daily builds, there is always an automatic entry at the top of the changelog reading "* Auto build.". I think it's the standard system... So I'll add the possibility in `./build-deb` to build nightly packages.

eroux:
(sorry for my last commit message, it should have been something like "improving git package build")

@olivierberten , the `build-deb.sh` script now takes a `--git` option to build a package based on the current git version. What I propose is this:

 * when a new version is out, update the changelog (should be done at realease time actually) and build regular packages
 * build regularly a `gregorio-git` package based on the `develop` branch

Both should be relatively straightforward with these changes, tell me what you think.

Also, this PR starts to be a bit fat, sorry...

eschwab:
Since the build changes were cherry picked into this, #127 should be addressed here.

eroux:
@eschwab : added in the commit mesage

rpspringuel:
This pull request seems to be pulling in more and more changes as time goes by. Any chance of merging it soon and dividing subsequent changes based on the issue they address?

henryso:
If we're committed to these changes for this release, I don't see why we can't merge this into release-3.0.

eschwab:
Everything looks to be in order and, yes this PR is beyond the scope of the original intent. I'll bite the bullet and merge so other work can get underway.

eroux:
Thanx for merging!

Issue #135, opened 2015-03-23
changing the formatting of first syllable after initial


eroux:
Currently Gregorio uppercases the first syllable after the initial, which make it impossible to "small capitalize". For example

    initial-style:1;
    %%
    (c4) P<sc>ri</sc>(f)<sc>mus</sc>(gh)

doesn't work. The solution, I believe, would be to have a `\firstsyllablestyle` that defaults to uppercase, but can be changed by the user. This requires some intervention of what is certainly the biggest spaghetti plate of gregorio, `lib/characters.c`, and requires bravery.

eroux:
If anyone is brave enough for this, #42 could be fixed at the same time (it is in the same function).

henryso:
I would like to try to fix this and #42, in order to get a better feel for that code, but I will most likely not have a satisfactory fix by release-3.0.  If we need it for release-3.0, I suggest someone with more experience take a look at it.

eroux:
Well, that's very brave of yours! This code is really terrible... I thought it would be ok for 3.1, but after thinking back about it, it might introduce a backward-incompatible change: the letters of the first syllable won't be uppercased in the `.gtex` files after this change.

Maybe the best solution would be to change the c code only to remove the current uppercasing, and do the uppercasing in TeX (it's easy to know if current syllable is the first one). It will solve the issue in a much simpler way. What do you think?

henryso:
Are you sure gregorio capitalizes the first letter?  It doesn't seem to do that for me.

###### foo.gabc
%%
(c3) test(d)

###### foo.tex
% File generated by gregorio 3.0.0-beta
\gregoriotexapiversion{3.0.0-beta}%
\begingregorioscore%
\greinitial{t}%
\grebeginnotes %
\gresetinitialclef{c}{3}{a}%
\gresyllable{}{e}{st}{1}{\gresetnextsyllable{}{}{}}{}{0}{%
\grelastofscore %
}{%
\greglyph{\char 178}{d}{g}{0}{}{}%
}%
%
\endgregorioscore %
\endinput %

henryso:
I think the bug is something else.  It's emitting something weird for \gresyllable.  Or at least it seems that way to me.

###### bad.gabc
%%
(c4) T<sc>est</sc>(e/ec)<sc>ing</sc>(d)

###### bad.tex
% File generated by gregorio 3.0.0-beta
\gregoriotexapiversion{3.0.0-beta}%
\begingregorioscore%
[...]

Issue #42, opened 2015-02-08
first syllable alignment when vowel after initial


eroux:
The alignment of first syllable text/neume should be improved when it contains a vowel

Currently, the alignment is made on the first letter after the the first one (which is the initial), in all cases (as was asked at the creation of Gregorio). The following behavior should be implemented:

 * when there is no vowel after the first letter (which is the initial), keep current behavior (alignment on first letter after the initial)
 * if there is a vowel in the letters after the first, align as if it was a regular syllable (ex: initial `C`, other letters `hri`, in this case the alignment is on the `i`)

henryso:
This should be fixed along with #135 in `develop`.

PR #142, opened 2015-03-24
Fix for #135 and #42

henryso:


lib/characters.c
```diff
@@ -1114,6 +1108,5 @@ gregorio_rebuild_first_syllable (gregorio_character ** param_character)
   current_character = *param_character;
   gregorio_go_to_first_character (&current_character);
   (*param_character) = current_character;
-  // this is not working, so people have to mind the way they put text styles on the initial.
-  //gregorio_rebuild_characters (&current_character, CENTER_FULLY_DETERMINED);
 }
+/* vim: set ts=8 sts=2 sw=2 noet: */
```
eroux: I admit I'm a bit against having editor-specific tags in files... I like the idea of editorconfig, and there's [a plugin for vim](https://github.com/editorconfig/editorconfig-vim#readme). This allows to put preferences once for the whole project, and change them easily... What do you think?

eroux:
Apart from my small comment (not really blocking), I think I'm ready to merge this PR (though I admit I don't understand each line of it, but I fully trust you). Maybe, before merging, there could be an associated test? This part of the code has proven to be a nest of bugs and things broke many times when there was a change in it, so a small non-regression test would be very useful I believe...

henryso:
Sorry, I didn't mean to leave that in.  I'll scrap that line.

henryso:
I will create tests for this in https://github.com/gregorio-project/gregorio-test

eroux:
Ok perfect!

eroux:
Does it address your last comment in #135 ?

henryso:
A qualified yes. As you say, this code is a nest of bugs, so I'm trying to run it through as many tests as I can before I say for sure.

eroux:
Just a reminder: can you please add this in the CHANGELOG?

Issue #144, opened 2015-03-25
examples/main-lualatex.tex should be updated for Gregorio 3


henryso:
We probably also want to remove examples/PopulusSion.tex from the repository.

PR #145, opened 2015-03-25
Fix for #144

henryso:
This fix for #144 does delete PopulusSion.tex.  If you disagree, I will re-do.


eschwab:
Dropping `PopulusSion.tex` is consistent with the decision to drop compiled files i.e. `.ttf`, `.pdf`.

rpspringuel:
Should there be a `.tex` file to go with `completetest.gabc` and `test_all_headers.gabc`?

eroux:
@rpspringuel I think these belong to the test/ directory or even better to https://github.com/gregorio-project/gregorio-test

rpspringuel:
Good point.

PR #148, opened 2015-03-25
Small fixes for Windows

eroux:


rpspringuel:
Since I think I'm the only one who can go through the whole Windows build cycle and test the results, I'll look at this as soon as I get my Internet back on my computer.

rpspringuel:
The icons and version information seem to be attaching correctly, but I'm experiencing a problem with the TeXworks configuration script.  It doesn't appear to be doing anything.  Allow me to play with it for a bit and I'll see what I can come up with.

eroux:
Ok!

rpspringuel:
I don't have a fix, but line 325 `current = current + 1` is where the script is throwing an error.

henryso:
Without fully understanding what's going on, I'd suggest moving `local current = 0` from line 307 to after line 295 `local lualatexfound = 0`.

rpspringuel:
Got it.  There was a scope problem.  I'm pushing the fix to this branch.

PR #151, opened 2015-03-25
Bug fix for #146

rpspringuel:
Fixes problem with completetest not compiling.  There was a typo in the distance names for `highchoralsignshift` and `lowchoralsignshift` in the algorithm that decided whether they were dimen or skip (resulting in them getting classified wrong).
Added in the checks against `\gre@scale@…` when changing `\grefactor`.  Not sure where this code went, because I know I've done it before.  Requests to not scale distances should now be respected.
Moved a couple of files from `examples` to `tests`.  They will eventually need to migrate to the gregorio-project/gregorio-test, but the state of my internet connection makes even getting this fix up difficult.


henryso:
Looks good to me.  Merging.

eroux:
This commit doesn't modify `examples/Makefile.am`, can you update it @rpspringuel ?

rpspringuel:
All I need to do is remove `completetest.gabc` and `test_all_headers.gabc` from the list of files, right?

If so I'll do that and open another pull request.

henryso:
That should be it.

eroux:
basically yes. It's a good habit to take when adding, removing or moving files... `make distcheck` is  a good way to test this kind of things

Issue #159, opened 2015-03-27
Using <c>color</c> tag introduces a spurious space


henryso:
###### mwe.tex
```
\documentclass[11pt]{article}
\usepackage{luatextra}
\usepackage{graphicx}
\usepackage{gregoriotex}
\usepackage[utf8]{luainputenc}
\autocompilegabc
\begin{document}
\includescore{mwe}
\end{document}
```

###### mwe.gabc
```
initial-style: 0;
%%
<sc>Small Caps</sc>(e) <c>colored</c>(e) x<c>colored</c>(eZ)
<i>Italic<b>Bold<sc>Small<c>Colored</c>Caps</sc>Face</b>Font</i>(iZ)
```

###### output
![mwe](https://cloud.githubusercontent.com/assets/421318/6862744/4ac51b26-d41c-11e4-99d2-b3e25127c9b6.png)

When typeset, the `<c>` is preceded by a space, which throws off the alignment of the note as well.   It's even more evident in `x<c>colored</c>` because gregorio uses the TeX `\grecolored` macro is for `c`, the first `o`, and `lored`, and there's a visible space before each.

PR #162, opened 2015-03-27
Fix for #159

henryso:
Fix for #159

### DIFF
--- a/tex/gregoriotex.sty
+++ b/tex/gregoriotex.sty
@@ -78,7 +78,7 @@
 
 \definecolor{gregoriocolor}{RGB}{229,53,44}
 
-\def\grecolored#1{
+\def\grecolored#1{%
   {%
     \color{gregoriocolor}#1%
   }%


eroux:
I don't know if some scripts can detect this, but it's a common error people do...

rpspringuel:
I can easily write a sed script that adds a `%` to every line that doesn't already end in a comment.  This could then be applied as a clean filter to people's setup, thus automatically adding them.  Would that be useful to people?

eroux:
Well, it's more subtle than this... The comments in TeX follow the following non-obvious rules:

 * a comment make TeX ignore
 what follows until the end of line and

henryso:
I think it's better if we write a lint-like tool to look for things like this.  Then let the developer fix as desired.

eroux:
Agreed, but this kind of tools doesn't exist and won't be very easy to write (TeX parsing is just madness)

PR #166, opened 2015-03-27
add hardening flags

eroux:
Add hardening flags (improves security), see [here](https://wiki.debian.org/HardeningWalkthrough) and [here](https://wiki.debian.org/Hardening) (not very fun reading). 

```bash
$hardening-check src/gregorio
  src/gregorio:
   Position Independent Executable: yes
   Stack protected: yes
   Fortify Source functions: yes (some protected functions found)
   Read-only relocations: yes
   Immediate binding: yes
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,7 +40,10 @@ AC_PROG_LIBTOOL
 AC_CHECK_TOOL([RC], [windres], [no])
 AM_CONDITIONAL([HAVE_RC], [test x$RC != xno])
 
-CFLAGS+=" -fstack-protector-strong -Wformat -Werror=format-security -fPIE "
+AX_CHECK_COMPILE_FLAG([-fstack-protector-strong], [CFLAGS+=" -fstack-protector-strong"])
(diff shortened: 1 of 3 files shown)


henryso:
My GCC doesn't have -fstack-protector-strong.  This is apparently a new feature for GCC 4.9.  Can this flag be detected?

eroux:
should be better now

henryso:
After `autoreconf -f -i`, `./configure` fails with

```
./configure: line 11995: syntax error near unexpected token `-fstack-protector-strong,'
./configure: line 11995: `AX_CHECK_COMPILE_FLAG(-fstack-protector-strong, CFLAGS+=" -fstack-protector-strong")'
```

What am I missing?

henryso:
I don't really know what I'm doing, but I was able to get it to work by adding `ax_cflags_sun_option.m4` and `ax_check_compile_flag.m4` from the Autoconf Archive into `m4/`.

eroux:
I think the second one is enough. I did the same, but I got confused when I removed the file from `m4` and it continued to work... I'm preparing a new commit

henryso:
Seems to work, though I can't check with stack-protector-strong.

PR #177, opened 2015-03-28
precise C standards

eroux:


eroux:
I'm wondering: as GCC5 and modern CLang use C11 by default, maybe we could target C11 too, this would allow to remove the hacks with Unicode and simplify this part of the code. What do you think?

henryso:
It depends on how long the compilers that substantially support C11 have been around.  If 3+ years, then I don't think it's unreasonable.  I'm afraid I don't know the answer to that.  C may have been my first language, but it's not what I use at work.

My second comment is that I think we should express our intent to allow code with GCC / Clang extensions to C99 / C11 (at least on a case-by-case basis), if I'm going to do any of the enum/inline stuff I mentioned on the sys-devel list.  The way the file is written it appears to limit us to C99 as implemented by GCC / Clang.

CONTRIBUTING.md
```diff
@@ -42,6 +42,8 @@ are the same under GNU/Linux and Mac OSX:
 
 ##### C files
 
+Gregorio uses [ISO C99 Standard](http://en.wikipedia.org/wiki/C99), and targets the [gcc](https://gcc.gnu.org/) and [clang](http://clang.llvm.org/) compilers only. Patches for other compilers can be accepted if they are clean, small and non-intrusive.
```
henryso: I think it should be "Gregorio uses the [ISO C99 Standard]..."  Also see my comment about GCC / Clang extensions in the pull request.

eroux:
Apparently, gcc supports C11 unicode since 4.7 (2012)... But complete C11 support is not as old, so I think it's best to stick with the current wording, but allow us to use `uchar.h` (not a priority though).

eroux:
@jakubjelinek don't hesitate to comment on this, you're the expert!

jakubjelinek:
I guess it depends on what exactly you mean by Unicode support.  Support for wchar_t and wchar.h/wctype.h is quite old (and is a library thing, so glibc or whatever other C library your OS uses).
uchar.h is a C library header, in glibc added 4 years ago, but I think many other OSes don't have that header at all, nor provide the library function.  So even if you require gcc or clang to compile gregorio, uses of recent C library headers could be a portability problem (dunno if gregorio is supported on AIX, Solaris, mingw, etc.).  If by unicode support you mean support for u"..." and U"..." string literals, then that is a compiler thing; without OS providing uchar.h, you'd need to use typedefs using __CHAR16_TYPE__ / __CHAR32_TYPE__ defines that gcc provides (dunno if clang does the same).

eroux:
Good point... maybe we should explicit that we target only `ARM`, `ARM64`, `x86` and `amd64`? would that be relevant? Or should it be more precise and also talk about `glibc`, `mingw32` and `mingw-w64`?

PR #179, opened 2015-03-29
Refactor - Phase I

henryso:
- Moved all C sources to src
- Removed dynamic plugin loading
- Removed OpusTeX and GregorioXML support
- Converted macros into inline functions
- Started converting sets of defines into enums


eroux:
Excellent ! this is a great step in lowering the barriers to Gregorio code! Do you think we should move the other folders at the same time, I was thinking about:

 * packaging
   * Windows
   * MacOSX
   * debian
 * GregorioTeX
    * fonts

What do you think?

henryso:
I have no strong opinion on packaging.  I don't feel it's too disorganized as it is right now.  That said, If we are going to do it, I would wait until all the remaining packaging issues are dealt with and merged, and release-3.0 is out the door.

Fonts, on the other hand, is actually a development model in and of itself.  I'd lean towards leaving it where it is, but that might just be my natural resistance to change, so if everyone wants to do it, I'm fine with that.

eroux:
Ok, I don't find it too disorganized neither, we can keep it that way.

Also, what do you think about stashing libtool completely and have only one Makefile.am that has everything (including files in `gabc/`, `dump/`, `gregoriotex/`)? That would make everything simpler... What do you think?

henryso:
I don't know autotools well enough to have a meaningful opinion on this.  I don't know the repercussions for current workflow and future development.

eroux:
I've pushed something in this direction, tell me what you think!

PR #180, opened 2015-03-29
Reorganize  and clean the C build process

eroux:
Please don't merge now, it's currently under discussion (the pull request is mainly to be able to discuss)


henryso:
Continuing the discussion from #179, I'm fine with how the build is set up after your changes.  Also, thanks for catching the `rindex` thing.  For some reason, I couldn't remember the proper name of the str* function to do what I wanted and a Google search found `rindex`.

eroux:
Well, I think this could be merged... @henryso what do you think?

BTW, just a thought: with the building process simplified, and the unique simple binary, a new possibility would be to include gregorio into TeXlive and GregorioTeX in CTAN, this would make the installation straightforward (with just one release every year, but for most people it should be good enough). What do you think? I'll discuss the possibility with the TeXLive people (for TeXLive 2016), without promising anything.

henryso:
Yes, let's merge this into develop.  In fact, I'll do it now.

As for TeXLive, it's definitely worth exploring.

PR #184, opened 2015-03-29
master is two commits ahead of release-3.0

henryso:
I kept the version from release-3.0.


henryso:
0 files changes, so it's just a version tree thing.

PR #189, opened 2015-03-30
Improve documentation

eroux:
* UserManual -> GregorioRef
 * append version to GregorioRef
 * various improvements to VersionManager


eroux:
@eschwab do you want to change the version in `doc/GregorioRef.tex` (you can push in this branch directly if you want). Also, I think the date should be set by `VersionManager.py` too, what do you think?

eschwab:
Yes, I can add the version to the titlepage. And the date too.

eroux:
sorry @eschwab I pulled the trigger... I there are no more comments, I think it can be merged

eschwab:
I just sat down to do it, so no problem. Looks good to me.

PR #200, opened 2015-03-31
using lzma instead of gzip for dist

eroux:
creating `gregorio-xxx.tar.xz` (extractable with `tar xJf` instead of `tar xzf`): this makes the tarball 645ko instead of 910ko (-30%). Seems like the most fashionable way to compress sources now (at least under Linux).

@rpspringuel @henryso @eschwab can you confirm it works for you?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@ AC_CONFIG_SRCDIR([src/gregorio-utils.c])
 AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_MACRO_DIR([m4])
 
-AM_INIT_AUTOMAKE([subdir-objects foreign])
+AM_INIT_AUTOMAKE([subdir-objects foreign dist-xz no-dist-gzip])
 
 AC_PROG_CC
 AC_PROG_CC_STDC
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@ AC_CONFIG_SRCDIR([src/gregorio-utils.c])
 AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_MACRO_DIR([m4])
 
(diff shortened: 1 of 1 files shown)


henryso:
Works fine for me.  I think you'll only run into problems with people who haven't upgraded their distribution in a while (like years) or who are using a server OS like CentOS.

eroux:
Well, their version of LuaTeX will simply be too old, or, if they have updated it, TeXLive uses xz too, so they should have it... Besides, I'm not so sure we have many such users, but who knows!

eschwab:
Fine for me, I live near the bleeding edge anyways.

rpspringuel:
I'm getting an error when running `./build.sh` here:

    config.status: error: cannot find input file: `src/config.h.in'
    Failed to configure Gregorio.

Should that be happening?

henryso:
`build.sh` doesn't do an autoreconf if the Makefile.in already exists (it assumes you have unpacked a tarball rather than pulled from git in this case).  Try deleting Makefile.in and run the script again.

rpspringuel:
Okay, that got things built.

Having made the dist, I can say that Mac OS 10.8 (Mountain Lion) doesn't recognize the format automatically and when I try to extract from the command line, I get an error that `J` is an unrecognized option.  Then again, the OS is nearly 3 years out of date.

I'm going to be upgrading to 10.10 (Yosemite) later this week (I finally broke down and bought the upgrade for the one program that was holding me to Mountain Lion).  I'll let you know how things work once that's done.

henryso:
We can compromise and go with bzip2.  Is `j` a recognized option on Mac OS ?

rpspringuel:
Yes, Mac can handle bzip2 files.  I've opened those before so I don't even have to check that one.

eroux:
should be ready to merge now

rpspringuel:
A bit late, but with the upgrade to Yosemite completed, I can say that it does recognize the format original proposed.

PR #207, opened 2015-04-01
Fix #205

### LINKED ISSUE #205: \setgrefactor -- Backwards incompatible
In `release-3.0` `\setgrefactor` must occur after `\begin{document}`. Previous versions did not have this requirement and `\setgrefactor` could appear in the preamble. Can everything be initialized earlier so `\setgrefactor` can be used in the preamble? If not this change should appear in `Upgrade.md` and possibly `Changelog.md` as well.

It would be great if the old behaviour could return. Then those who have their own config or documentclass can continue using them without needing to add `\setgrefactor` later in the document.

rpspringuel:
Changes when the default space configuration is loaded so that spaces and `\grefactor` can be changed in preamble.  This restores functionality that used to exist before I made all my changes to the way spaces are handled.

Addresses #205

### DIFF
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -727,8 +727,7 @@
 }%
 
 
-% We input the default config, for everything to work fine it has to be done after the gregoriotex package is completely loaded (so that all functions are well defined).  As a result we delay the loading to the beginning of the document.
-\AtBeginDocument{\GreLoadSpaceConf{default}}%
+% For everything to work fine the default configuration has to loaded after the gregoriotex package is completely loaded (so that all functions are well defined).  As a result you'll find that as the last line in gregoriotex.tex
 
 
 %%%%%%%%%%%%%%
--- a/tex/gregoriotex.tex
+++ b/tex/gregoriotex.tex
@@ -1441,3 +1441,7 @@
 \def\GreUseZeroHyphen{%
   \global\let\grehyph\grezerhyph %
 }%
+
+
+% We load the default space configuration.
+\GreLoadSpaceConf{default}%


eschwab:
Does the job. Thank you.

henryso:
Is there a place that tells me which things I can use before the `\begin{document}`?  After applying this commit (and maybe before -- I haven't checked), I get some really weird results when I move the following lines to before `\begin{document}`:

```
\setspaceafterinitial{2.2 mm}{1}
\setspacebeforeinitial{2.2 mm}{1}
\grechangedim{abovelinestextraise}{0 cm}{0}
\grechangedim{abovelinestextheight}{0.3 cm}{0}
\setgrefactor{15}
\renewcommand{\greinitialformat}[1]{{\fontsize{35}{35}\selectfont #1}}
```

I haven't figured out what causes to be weird, but I get something like this:
out-5

If I move these lines after `\begin{document}`, it works fine.

rpspringuel:
After this change all of those should be usable before `\begin{document}`.  Only the last (`\renewcommand...`) would have been usable before `\begin{document}` before this commit.

I think what's going on here is an interaction issue.  The width of the staff lines is automatically calculated by GregorioTeX because it depends on the width of the printable space in the document (`\hsize`).  Once the spacing configuration is loaded, or anytime you change a distance, GregorioTeX recomputes this length to make sure it hasn't changed as a result of the changes you made.  In what you posted, the staff lines are probably coming up short because after GregorioTeX does these calculations, something else changes `\hsize` to make it wider.  Since you get the problem when the commands are placed before `\begin{document}` but not after, my guess is that you're using a package or class that adjusts `\hsize` with an `\AtBeginDocument` hook.  To test, try putting the following after `\begin{document}`

    \makeatother
    \gre@calculate@linewidth
    \gre@calcualte@stafflinewidth
    \makeatletter

This will force GregorioTeX to recompute the line width numbers, and only those numbers.  Let me know if this works.  If it does, I might have an idea for how to patch things so you don't have to worry about it.

eroux:
IIRC, staff width was adapted to `\hsize` at each score (in `\begingregorioscore`), did it change recently or is my memory wrong? Whatever the answer, not doing this is a problem, each score should be typeset to `\hsize`. I know some people using different width for their scores in the same document...

eschwab:
I have similar commands as @henryso before `\begin{document}` without any issue.

rpspringuel:
@eroux It was.  That was one of the things I removed as part of my code clean up to prepare for my spacing work.  In my testing at least, it appeared to be redundant, as that calculation was being handled elsewhere already.  Clearly I missed this use case.

Working on a fix.

henryso:
Putting
```
\makeatletter
\gre@calculate@linewidth
\gre@calculate@stafflinewidth
\makeatother
```
after `\begin{document}` did fix the issue for me.

PR #211, opened 2015-04-01
Refactor - Phase III (?)

henryso:
Please review these changes.  I have converted the remaining enum-like defines to enums and put unions into the structures to reduce multi-purpose fields.  This was a fairly messy refactor, so I'm expecting bugs, but it does seem to work against my battery of tests.  gabc->gabc has not been tested extensively (do we still need that functionality?).

Please don't merge this until we've had time to discuss.

### DIFF
--- a/.indent.pro
+++ b/.indent.pro
@@ -32,3 +32,14 @@
 -T gregorio_word_position
 -T gregorio_lyric_centering
 -T gregorio_type
+-T gregorio_shape
+-T gregorio_bar
+-T gregorio_sign
+-T gregorio_space
+-T gregorio_liquescentia
+-T gregorio_h_episemus
+-T gregorio_glyph_type
+-T gregorio_word_position
+-T gregorio_lyric_centering
+-T gregorio_extra_info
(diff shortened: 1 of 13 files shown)


eroux:
Wow, that's a lot of great work! I was actually thinking about something more hackish but more simple, just using a basic union between "something not a note" and a note, just as something temporary for gabc to use before it can build the structure (that wouldn't have changed). But this is really a step further! I like it, I just need to review and test it more carefully, but I think it goes in a cleaner direction, which seems good! I'll let you know when I'll have more time to review (tonight Paris time certainly)

eroux:
I'm sorry, I haven't been able to take a real look at this... how stable/tested is this? I think what we could do is merge and test thing a bit further than usual?

henryso:
It runs through the tests in gregorio-test (though the dumps are different in an expected manner).  I have been running it against by current large project as well as scores in gregobase.  At this point, I think it's merge-able into develop but we should definitely do more testing there.

eroux:
Let's go then!

Issue #124, opened 2015-03-22
build system for OSX


eroux:
The current build system for OSX is quite old and might not be the best solution. It is (more or less) documented in [macosx/README.MacPackage](macosx/README.MacPackage) (which should probably be renamed as `macosx/README.md`), but according to a few people using MAC, it's outdated. I really have no idea what is the best solution to distribute Gregorio for OSX, but maybe it could take the form of a [brew](http://brew.sh/) or [ports](https://www.macports.org/) package? Is building a `.pkg` still relevant? What do OSX users think? What do you personally use, and what do you think would be best for the beginner user?

rpspringuel:
I've looked at creating a portsfile for Gregorio, and while I don't completely understand the process yet, I think that it should be fairly straight forward since we follow the standard configure/make/make install process and just need a post-flight script to install GregorioTeX.  Once the build and install scripts are finalized, I think it would be worth-while looking into this.

Of course, using MacPorts (or Homebrew, for that matter) still means that the user is building from source.  A `.pkg` file would be installing prebuilt software on to the user's system, and thus might be more user friendly to the less technically savy.  I've built a `.pkg` file before, but it's been a while and it wasn't as nearly complex as what needs to be done for Gregorio.  Give me some time and I'll look at what we've already got and give you my opinion on it.

eroux:
Ok good! I'm just making sure we'll be able to make a proper complete release for 3.0 (proper source tarball, installer for windows, package for OSX, maybe debian packages, etc.), but it's no emergency

rpspringuel:
Okay, after a gander, I'd say that what we've already got needs to be scrapped.  It's based on using PackageMaker which was deprecated in Xcode 4.6.  Of course, the last time I built a package, it was with PackageMaker (indeed, the program is still floating around on my system since I've always migrated everything when getting a new machine), so I don't know how to use the new system.

I'm willing to look into how to create a `.pkg` installer, but it's probably going to take me a while to figure it out.  For now I propose dropping what's there and replacing it with a simple README that points to the website instructions for building from source.

eroux:
Yes, if it's complicated (which it seems), we can skip it for 3.0

eroux:
[...]

PR #212, opened 2015-04-01
Macbuild

rpspringuel:
As requested in #124.

### Do not merge, WIP

Still need TODO:

- Make `VersionManager.py` aware of the new `.pkgproj` files which contain version numbers in them.
- Update `CHANGELOG.md`

Question: Right now there is one installer which installs everything.  This forces the contrib, examples, and doc components to be targeted to /Users/Shared/ even though they are relocatable on their own because the base files are not relocatable.  Should I create a separate installer which installs just these extras which would preserve their ability to be relocated?  If so, should the main installer still install these extras?


macosx/README.md
```diff
@@ -0,0 +1,23 @@
...
+This script contains all the necessary commands to create the distribution package.  It will first build and "install" the package into the `macosx` directory (this way we know where everything is).  It will then use Pandoc to convert the necessary documentation files from markdown to ??? (the format required by Packages).  Once that's done, it will use the command line interface for Packages to build the installer package.
+
+The distribution package is composed of several raw packages, each responsible for installing different things:
+
+1. gregorio-base (embedded in the main installer): This installs the command line tool and the GregorioTeX files.  These are the required components and must be installed.
+2. gregorio-examples: This installs the examples folder.  This is an optional install (installed by default).
+3. gregorio-contrib: This installs the contrib folder.  This is an optional install (installed by default).
+4. gregorio-doc: This installs the documentation (doc folder).  This is an optional install (installed by default).
+
+## Known Issues
```
eroux: I would replace "Issues" by "Limitations", I don't think these are really issues

macosx/README.md
```diff
@@ -0,0 +1,23 @@
+# Building a Mac OS X Package
+
+
+To build the distribution package for Gregorio for Mac, you first need to install [Packages](http://s.sudre.free.fr/Software/Packages/about.html) and [Pandoc](johnmacfarlane.net/pandoc/index.html).  Once those are installed, simply run:
+
+`create_pacakge.sh`
+
+This script contains all the necessary commands to create the distribution package.  It will first build and "install" the package into the `macosx` directory (this way we know where everything is).  It will then use Pandoc to convert the necessary documentation files from markdown to ??? (the format required by Packages).  Once that's done, it will use the command line interface for Packages to build the installer package.
```
eroux: I think `macosx` should be `build/` and `??? (the format required by Packages)` should be `html`

macosx/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The distribution package is composed of several raw packages, each responsible for installing different things:
+
+1. gregorio-base (embedded in the main installer): This installs the command line tool and the GregorioTeX files.  These are the required components and must be installed.
+2. gregorio-examples: This installs the examples folder.  This is an optional install (installed by default).
+3. gregorio-contrib: This installs the contrib folder.  This is an optional install (installed by default).
+4. gregorio-doc: This installs the documentation (doc folder).  This is an optional install (installed by default).
```
eroux: I don't know how things usually are under Mac, but aren't so many packages a bit overkill? I don't really see why it would be a problem for a user to have everthing installed? I'm quite sure the package metadata that will be installed will require more storage than the files it will install (130ko for the three last packages combined)... What do you think?

eroux:
Apart from a few comments, my main concern here is to be sure that GregorioTeX installation adapts to any TeX environment, through TeX variables TEXMFHOME, TEXMFLOCAL, etc. as it does under Windows and Linux? Or is TeX on OSX systems installed in the exact same dirs all the time? That's not the funniest part of an installer to make, but I think it would be a regression not to be able to adapt to any TeX system anymore... It used to be done with the same kind of lua installer as under Windows, this is maybe a bit wrong and could be converted to bash scripts using `kpsewhich --expand-var` like `install-gtex.sh` does, what do you think? `install-gtex.sh` could even be used as is...

eroux:
Sorry I didn't read the question in your pull request, but I think I kind of answered it in a comment: one package installing everything in `/Users/Shared/` is fine for me...

macosx/README.md
rpspringuel: I made them separate packages so that the user can decide which they want to install.  A standard install will install them all, but if the user selects a custom install, they can chose not to install any combination packages 2, 3, & 4.

eroux: I understood that, but I just don't really see the point... installing 130ko shouldn't be a problem for anyone... that's what is done by the Windows installer, the debian packages, etc. and I think it's quite ok...

rpspringuel:
As is, my installer targets the TeX files to `/usr/local/texlive` which is where TeXLive and MacTeX put the local texmf tree by default.  The installer also checks to make sure this directory exists before installing.  If it can't find it, it won't let the user install.

Mac installers require absolute paths for the files they are installing, so it's not a simple process of adapt to various TeX environments.  The solution would have to take the form of installing everything into a dummy directory and then running a post-flight script which would move things based on what it could actually discover about the user's system.  I would also need to create a check script which would look for the appropriate variables ahead of time to make sure there is a valid target destination.  Otherwise the user would get an installation failure after the installer tried to install, rather then the more useful message that tells them the installer can't install because it can't find a valid destination.

Both of these are possible, just a bit more complicated than what I've done so far.

eroux:
I think it would be good to go in this direction yes... A first check: does TeXLive put its binaries (mainly `kpsewhich` in the `$path`? If not, I'm not sure it's even doable to detect a custom install... Also, for the dummy directory, `/tmp` seems fine.

rpspringuel:
TeXLive and MacTeX both put their binaries in `$PATH` (or at least are capable of doing so).  I wouldn't know about other custom distributions.  On my machine `kpsewhich -var-value TEXMFLOCAL` pulls up the correct target directory.  I'll look at creating the scripts which will do what I described.

eroux:
Ok good, thank you very much!

eroux:
Also, this is less important: 90% of the _Packages_ files seem to actually be `.plist` (like [this one](https://github.com/atom/atom/blob/master/resources/mac/atom-Info.plist). I'm wondering if there is an option in Packages to import a `.plist` file? If so, maybe we could isolate it and just tell the package builder to import it into Packages? I admit I'm a bit affraid of depending on Packages, which seems developped a bit quickly and I would like to go in the direction of pure bash scripts using the XCode tools... but we can keep it for later.

rpspringuel:
The fact that so much of the file is a plist leads me to suspect that Packages is basically a GUI for the XCode tools.  Since it's open source, some time spent studying the source should answer that question.  When my OS upgrade completes, I'll download the source and take a look.

eroux:
That's what I suspect too and in fact I already took a look yesterday... and this look at the code made me want to swich to XCode directly to be honest... Also, another option: when you'll have a working .pkg (if possible a simple .pkg than a .mpkg), I can try to reverse engineer it: I already started with other pkg files (I can extract them, extract the bom, mount hfsplus, etc.) but I just need a working gregorio package to be able to build one entirely under Linux... if you have one now, can you send it to me by mail? Of course, building under Linux is almost useless, as I cannot compile the binary, but I could compile in macincloud...

eroux:
Thank you, this starts to be really good!

Only on remark left : if it's a single package, could it be just a `.pkg` instead of a `.mpkg`? (I'm not sure, but that what it seems when I google) Anyway, it's not blocking, and I believe this is the final step towards releasing 3.0.0-rc1, tell us when you think it's ready for rc1 (it could be now). Starting to release 3.0.0 for Easter is really good!

rpspringuel:
I'm having trouble with my internet connection again and haven't been able to update Xcode since my OS upgrade as a result.  This means I can't build Gregorio and run the final tests needed to make sure everything works.  I think I've got everything right at this point, but until I can build and test again I won't know for sure.

eroux:
Well, for me it seems ok... do you want to wait for your XCode to get upgraded before merging or should I merge now? Also, I'll just make a few more small tests, but otherwise I think we'll be ready to release after this merge

rpspringuel:
I'd prefer to test before merging if possible.  Today seems to be a good day for my internet connection so far.  The Xcode upgrade is now working on the command line tools.  Once that's complete the final step will be to update MacPorts.  Knock on wood that all goes well. 

If all doesn't go well, then I'm not opposed to merging and publishing the release candidate tomorrow without testing.  I can always make additional changes once something is reported as broken.

rpspringuel:
I'm up and running with my testing and so far the packages are a failure.  Files are installed correctly, but the post-flight scripts are failing (which means that GregorioTeX components are not being moved into place).  I'm trying to work out what is failing and why right now.

eroux:
No problem, take your time

rpspringuel:
I discovered the problem and found a work around, but I'm not sure how much better this solution is than my original installers: I've had to hard code the path to `kpsewhich`.  Is that really all that different from hard coding the path to `texmf-local`?

eroux:
Well, it's not much better but it's a bit better yes... To be sure: this means that PATH in the environment of the installer is different from the usual PATH? That's quite a questionable design choice... Anyway, the only little improvement that could be made is to first check if kpsewhich is in the path or not before trying `/usr/texbin`, you can do that with `which kpsewhich` (or by checking if `kpsewhich` returns an error), but it can be for a future version... Do you want to do it now or shall we release?

rpspringuel:
Let's go to the release candidate as it stands.  I'll try and come up with some other ways to look for TeX installation that are less path dependant.

eroux:
Ok, I'll tag the release and upload the tarball, can you upload the Mac Package and windows installer?

eroux:
@rpspringuel the release is now appearing on https://github.com/gregorio-project/gregorio/releases tell me when you'll have the files uploaded, I have a mail ready for the list.

rpspringuel:
Installers are uploaded.

eroux:
Sorry, I pulled the trigger too fast for my mail (I wasn't planning to send it before your go)

Thank you!

rpspringuel:
Not to worry.  I had the installers posted a good hour before I got around to posting here.

PR #213, opened 2015-04-01
Fix #207

rpspringuel:
Re: #207 

Changed timing of `\gre@computespaces`.  Now instead of calling it whenever we change a distance, we call it once at the beginning of each score.  GregorioTeX will now respect the document width on a score by score basis as a result.
Also removed some redundant code from `\setgrefactor` that was already being done as part of `\begingregorioscore`.

Please test.  I don't have time for anything more than the basics at the moment.


henryso:
This change looks good to me.

Issue #182, opened 2015-03-29
Version Number idea


rpspringuel:
So, this idea was sparked by the discussion going on in #181.  I thought I posted a remark on it there, but I don't see it so either it got lost or I'm mistaken.  In any event, I've had some further thoughts on the idea, so here it is fleshed out.

It would be nice if when you build on a branch which is not master, there was some indication of exactly where you were in the version number.  This way, it's easy to tell exactly when you most recently built and installed gregorio (by invoking the version number from the command line).

To that end the following would need to add the following code somewhere in the build process:

1. Before starting to build, inquire what branch we are on (`git rev-parse --abbrev-ref HEAD`).
2. If we are not on master, then get the description of where we are (`git describe`) and the current version number (`VersionManager.py --getcurrent`).  If we are on master, skip to step 4.
3. Change the version number to the description of where we are (note, the description takes the form `most_recent_tag-#_commits_since_tag-short_SHA_of_last_commit`, we might want to swap `#_commits_since_tag` for `branch_name` and possibly do some other jiggering to comply with semantic versioning.).
4.  Build gregorio.
5. If necessary (i.e. if we changed it), revert the version number to what it was before we built.

I'd personally prefer the code that does this in the Makefile, because that will make it harder to avoid.  If it's in build.sh, then some one who doesn't use that script might have these changes applied.

This would make the build process dependent on python, but only for those not building a stable release (i.e. something on master).  I don't think that's a huge issue, because these are our people on the cutting edge who are either developing or testing stuff for us.  I think having the extra information about where they are when they observe something outweighs the added dependency.

Thoughts?

henryso:
I'll have to admit that this would break my set-up, but that's because I'm using a non-standard git client to handle all my version control needs.  If we do this, I'll figure out a way to make it work for me.

eroux:
There is already a more or less similar version number for debian nighly builds, see `./VersionManager.py -dg`, there are a few differences though:

 * it adds the date, so that versions are incrementing (this is important for debian, it needs to detect that the new version is more recent than the old one
 * it doesn't print the branch, but this might be a good idea to add it... maybe instead of "git"?

[...]

PR #215, opened 2015-04-01
Supplied the branch version as VERSION for building the gregorio executable.

henryso:
This is an attempt to implement #182

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -8,15 +8,16 @@ gregorio_SOURCES = characters.c gregorio-utils.c messages.c struct.c \
 				   gregoriotex/gregoriotex-write.c gregoriotex/gregoriotex.h
 
 @MK@ifneq ($(wildcard ../.git),)
-@MK@  _branch_ = $(shell git rev-parse --abbrev-ref HEAD)
-@MK@  ifneq ($(_branch_),master)
-@MK@    DEFS += -DBRANCH_VERSION='"b$(_branch_)-$(shell date +%Y%m%d_%H%m)-$(shell git rev-parse --short HEAD)"'
+@MK@  _tag_ = $(shell git describe --exact-match HEAD 2>/dev/null)
+@MK@  ifeq ($(_tag_),)
+@MK@    DEFS += -DBRANCH_VERSION='"b$(_branch_)-$(shell date +%Y%m%d_%H%M)-$(shell git rev-parse --short HEAD)"'
 @MK@  endif
 @MK@else
 @MK@  ifneq ($(wildcard ../.hg/git),)
-@MK@    _branch_ = $(shell cat ../.hg/bookmarks.current)
-@MK@    ifneq ($(_branch_),master)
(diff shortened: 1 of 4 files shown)


src/Makefile.am
```diff
@@ -4,14 +4,28 @@ AM_CFLAGS =
...
 gregorio_SOURCES = characters.c gregorio-utils.c messages.c struct.c \
 				   unicode.c struct.h messages.h unicode.h characters.h \
-				   plugins.h dump/dump.c dump/dump.h gregoriotex/gregoriotex-write.c \
-				   gregoriotex/gregoriotex.h
+				   plugins.h config.h dump/dump.c dump/dump.h \
+				   gregoriotex/gregoriotex-write.c gregoriotex/gregoriotex.h
 
+@MK@ifneq ($(wildcard ../.git),)
+@MK@  _branch_ = $(shell git rev-parse --abbrev-ref HEAD)
+@MK@  ifneq ($(_branch_),master)
```
eroux: I think it would be good to detect tags too, for them we just want the normal version... You can check if `git describe --exact-match HEAD` fails or not for this

henryso: Maybe the tag check should replace the master check.  What do you think?

eroux: Right, master head is not supposed to be in a non-tagged state

henryso:
I've made the suggested change.

src/config.h
```diff
@@ -0,0 +1,30 @@
+/*
+ * Gregorio configuration headers. Copyright (C) 2015 Elie Roux
```
eroux: Just a detail: I tend to change the copyright to "The Gregorio project". Maybe we should discuss it with a broader audience, but I tend to prefer it: legally speaking it's less precise, but it's mostly for the symbol: I don't want Gregorio to be my personal software, with my name, etc. I'd prefer being just one of the authors, and to let other people participate without feeling underneath me...

henryso: `Gregorio project` or `The Gregorio project` or `Gregorio project authors`?  gregorio-utils.c has `Gregorio project` in the top-comment and `Gregorio project authors` in `print_licence()`.

If you care about my vote, I like `The Gregorio Project` (using title case), but I'm no legal expert or anything like that.

eroux: Legally speaking, The Gregorio Project doesn't mean anything, while authors are listed in CONTRIBUTORS.md... but I'm not sure it matters, nor that any lawyer could say that there is a strict difference...

henryso: How about `Élie Roux and the Gregorio Project authors`.  Then at least your name is on it to lend legal legitimacy?

eroux: Why not, but I'd prefer not to see my name at all... It would allow to have the same copyright notice everywhere (putting my name in files you create doesn't make much sense)

henryso: Fine, then which of the other options should I use?

eroux: What do you think of "The Gregorio Project (see CONTRIBUTORS.md)"?

henryso: Fine. I'll use that.

eroux: Ok good, I'll make a separate pull request to change in other places

henryso:
I updated the copyright comment as suggested.

PR #216, opened 2015-04-02
fix autoconf warnings

eroux:
* use `clean-local` instead of `clean`
 * use `install-data-local` instead of `install`
 * use `pdf-local` target, now `make pdf` acts as expected (in the root directory)

### DIFF
--- a/fonts/Makefile.am
+++ b/fonts/Makefile.am
@@ -39,10 +39,10 @@ gresym: gresym.ttf
 
 greextra: greextra.ttf
 
-clean:
-	@$(TOP_LEVEL_MAKE) $(RM) -f $(TTFFILES)
+clean-local:
+	$(RM) -f $(TTFFILES)
 
-install:
+install-data-local:
 	@$(TOP_LEVEL_MAKE) echo 'The Makefile in the fonts directory is no longer used for installing'
 	@$(TOP_LEVEL_MAKE) echo 'GregorioTeX.  Please use install-gtex.sh from the parent directory'
 	@$(TOP_LEVEL_MAKE) echo 'to install GregorioTeX.'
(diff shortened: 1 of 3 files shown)


fonts/Makefile.am
```diff
@@ -39,10 +39,10 @@ gresym: gresym.ttf
 
 greextra: greextra.ttf
 
-clean:
-	@$(TOP_LEVEL_MAKE) $(RM) -f $(TTFFILES)
+clean-local:
+	$(RM) -f $(TTFFILES)
 
-install:
+install-data-local:
```
henryso: The purpose for having this here was for people used to the old method of going into the fonts directory and typing `make install`.  The system would then tell them about the new method of installing.  By changing this target this way, it defeats the purpose of having an install target in the first place.

eroux: `make install` calls `install-data-local`, so no behaviour change here, it's just the clean way to do what we did before

henryso: You learn something new every day, especially if you didn't know much in the first place!

henryso:
I get this warning:
```
src/Makefile.am:5: warning: compiling 'dump/dump.c' in subdir requires 'AM_PROG_CC_C_O' in 'configure.ac'
```

eroux:
you get it when you run `autoreconf -fi`?

henryso:
Yes, when I run `autoreconf -f -i`.

eroux:
Apparently it's [deprecated](http://www.gnu.org/software/automake/manual/html_node/Public-Macros.html) since Automake 1.14 (end of 2013), but it doesn't issue a deprecation warning (yet?), so I've added it

henryso:
I guess it's "obsolescent" rather than deprecated (according to that link), so I guess it's fine to put it in for now.

eroux:
right

PR #218, opened 2015-04-02
Substituted "Gregorio" for "this program" in copyright messages.

henryso:
As commented in https://github.com/gregorio-project/gregorio/commit/ba4cc94a6a0442212d3a2b15977fe686c59a7273 for #217

Feel free to reject if you don't agree.


henryso:
In my excitement making this change, I forgot to separate out the addition of the `-Wno-portability` flag to `AC_INIT_AUTOMAKE`.  I needed to do this because the versioning changes (#215) need GNU make extensions.  If this change is rejected, I can apply this again later.

Issue #209, opened 2015-04-01
Code indentation redux.


henryso:
Something that is beginning to irk me with the indentation parameters we have set up for C is that ...

```
{
    if (there_is_a_long_function_name_with_arguments_or_deep_indentations(they,
                                                                          start,
                                                                          to,
                                                                          line,
                                                                          up,
                                                                          like,
                                                                          this)) {
        ...
    }
}
```

Maybe we can add something like -nlp -ci8 ?

eroux:
sure

henryso:
The auto-indentation have clobbered comment formatting over the years (they were already a bit messed up before we switched to K&R).  I'm not sure how to deal with that.  Should we consider disabling comment reformatting?

eroux:
ok

henryso:
Changes merged.

PR #219, opened 2015-04-02
Made indentation changes per #209.

henryso:


henryso:
Merging this as a few days have passed on #209 with no disagreement.

PR #223, opened 2015-04-06
PKGBUILDs for gregorio and gregorio-git for the AUR.

avbop:
These are working fine for me; how do they look?

I haven't heard back from the current package maintainer, so it might take a while to get the stable package on the AUR. I'll update the "Maintainer" fields once that's figured out.


avbop:
I've heard back from the old maintainer of the AUR packages, and he turned them over to me, so I'm ready to take over there for this and future releases. This would be my first package on the AUR, so I might need a little while to figure things out, but it looks pretty straightforward from what I've read thus far.

eschwab:
I've been thinking about the usefulness of having these `PKGBUILD`s in the repo. With the AUR it seems a bit unnecessary to include these here. The versions are auto updated based on the last commit and leads to a commit loop. Also for the stable version, the SHA of the tar needs to be updated which cannot be done until after a release is made, thus the commit which includes the proper SHA is after the tagged commit of the release. My thinking is that once a commit is tagged with a release version, there should not be any need to commit further things to finish a release.

I am not opposed to keeping these files here, but I think there are other ways of distributing them that need to be explored. After @avbop gets the AUR updated, we should add a note about it on our website. Perhaps make a subsection for different Linux distros (PPA, AUR, etc.). Thoughts?

henryso:
@avbop If you change the lines which have comments attached, GitHub will assume the comment is outdated and do exactly what you are seeing, so I think it's fine what you are doing.  You can always click on the commit hash above to see the comments in context.

@eschwab That smells like it needs a separate git repository.  Perhaps gregorio-arch or gregorio-aur?

avbop:
@eschwab I agree on the separate repo. I'd vote for gregorio-project/gregorio-aur, which I could fork and make a new pull request for. I don't have strong feelings so long as I can still get feedback from you all, especially for stable releases.

@henryso Thanks. I guess that makes sense. The last code review tool I used kept the comments, though, so I wanted to make sure.

eroux:
done: https://github.com/gregorio-project/gregorio-aur

avbop:
Could someone with write access please create a master branch so I can make a pull request?

eroux:
done

henryso:
These files now belong to their own repository.  Closing this pull request.

PR #228, opened 2015-04-06
Getting the new documentation name into Windows installer

rpspringuel:
Since the name of the documentation pdf changed, the ISS file had to be made aware of the new file name.
This also required a new rule set in `VersionManager.py` that marks the filename for updating (since it contains the version number).  Unfortunately, ISS doesn't allow inline comments, so this has to be handled by putting the comment on the line before and then modifying the next line.

### DIFF
--- a/VersionManager.py
+++ b/VersionManager.py
@@ -131,6 +131,7 @@ def replace_version(version_obj):
     print('Updating source files to version {0}\n'.format(newver))
     for myfile in GREGORIO_FILES:
         result = []
+        following_file = False
         with open(myfile, 'r') as infile:
             for line in infile:
                 if 'AC_INIT([' in line:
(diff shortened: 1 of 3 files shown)


VersionManager.py
```diff
@@ -131,6 +131,7 @@ def replace_version(version_obj):
     print('Updating source files to version {0}\n'.format(newver))
     for myfile in GREGORIO_FILES:
         result = []
+        following_file = False
```
eroux: shouldn't it be following_line instead of following_file?

windows/gregorio.iss
```diff
@@ -38,7 +38,8 @@ Source: "../CHANGELOG.md"; DestDir: "{app}";
 Source: "../README.md"; DestDir: "{app}";
 Source: "../CONTRIBUTORS.md"; DestDir: "{app}";
 Source: "../UPGRADE.md"; DestDir: "{app}";
-Source: "../doc/UserManual.pdf"; DestDir: "{app}";
```
eroux: to be more consistent with other comments, maybe `PARSE_VERSION_FILE_NEXTLINE` would be better?

VersionManager.py
rpspringuel: Originally I used just `following` but thought that I should indicate that we're expecting a file name format version substitution in the next line, hence the added `_file` (in case at some point a similar solution has to be implemented for a regular version number substitution).  Perhaps a variable name even more descriptive would be better, like: `following_line_filename`

windows/gregorio.iss
rpspringuel: Sure.  I can make that change.

VersionManager.py
eroux: ok

eschwab:
Looks fine.

PR #231, opened 2015-04-06
Expanding places searched for `kpsewhich`

rpspringuel:
The scripts now look in both the limited path available to the Installer and the default location for a TeXLive installation.
I also added some details to the error message so that the user who gets it has some more information to go on.

### DIFF
--- a/macosx/douninstall.sh
+++ b/macosx/douninstall.sh
@@ -6,7 +6,10 @@ PREFIX="/usr/local"
 BINDIR="$PREFIX/bin"
 PKGCONFIGDIR="$PREFIX/lib/pkgconfig"
 GREINCLUDEDIR="$PREFIX/include/gregorio"
-TEXMFLOCAL=`/usr/texbin/kpsewhich -var-value TEXMFLOCAL`
+TEXMFLOCAL=`kpsewhich -var-value TEXMFLOCAL`
+if [ -z "$TEXMFLOCAL" ]; then
+    TEXMFLOCAL=`/usr/texbin/kpsewhich -var-value TEXMFLOCAL`
+fi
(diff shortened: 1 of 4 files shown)


macosx/douninstall.sh
```diff
@@ -6,7 +6,10 @@ PREFIX="/usr/local"
 BINDIR="$PREFIX/bin"
 PKGCONFIGDIR="$PREFIX/lib/pkgconfig"
 GREINCLUDEDIR="$PREFIX/include/gregorio"
-TEXMFLOCAL=`/usr/texbin/kpsewhich -var-value TEXMFLOCAL`
+TEXMFLOCAL=`kpsewhich -var-value TEXMFLOCAL`
+if [ -z "$TEXMFLOCAL" ]; then
+    TEXMFLOCAL=`/usr/texbin/kpsewhich -var-value TEXMFLOCAL`
+fi
```
henryso: Is it possible TEXMFLOCAL will have a path in it (a la #235)? If so, it might be better to use `kpsewhich gregoriotex.tex` to find the location of the things to delete.

macosx/test.sh
```diff
@@ -2,12 +2,17 @@
 
 # Test to see if we have a valid TeX installation
 
-TEXMFLOCAL=`/usr/texbin/kpsewhich -var-value TEXMFLOCAL`
-
+TEXMFLOCAL=`kpsewhich -var-value TEXMFLOCAL`
```
henryso: I don't know if you do, but you might need to handle paths in TEXMFLOCAL here also.

rpspringuel: I've never seen it before, but I'll edit to handle it.

rpspringuel:
That should handle it.  Installers work on my machine after these changes.

henryso:
If you say this works, then let's merge it.

PR #236, opened 2015-04-07
Modified to handle paths in TEXMF variables.

### LINKED ISSUE #235: Update install.sh to handle multiple directory names in TEXMFLOCAL
In Arch Linux, `TEXMFLOCAL` is apparently set to `/usr/local/share/texmf:/usr/share/texmf`.  If this is a standard thing (the colon makes me suspicious...), then install.sh should be made to handle it.

henryso:
Fixes #235.

### DIFF
--- a/install-gtex.sh
+++ b/install-gtex.sh
@@ -56,65 +56,65 @@ FONTSRCFILES=("${FONTSRCFILES[@]/#/fonts/}")
 
 arg="$1"
 case "$arg" in
-	system)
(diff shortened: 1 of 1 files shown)


henryso:
I'm not entirely sure about using `-expand-path` here, but since people can put brace expansions in the TEXMF* variables, I figured this is the best  way to handle it.

henryso:
Any objections to merging this?

Issue #239, opened 2015-04-08
Move font glyphs to UCS Private Use Area


henryso:
See earlier part of the discussion in #232.

henryso:
According to [Wikipedia](https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Private_use), there are 137468 private use code points at U+E000 to U+F8FF, U+F0000 to U+FFFFD, and U+100000 to U+10FFFD.  How many glyphs do we currently use?

henryso:
If that's not enough, I *think* it's possible to reduce the number of code points needed by using multiple characters to select a glyph (similar to how ligatures like `ffi` work in Unicode fonts).

eroux:
The number of glyph is not really the most important, it's ridiculously low: 3137. The problem is how they are organized. Currently (look at lines 343 for `squarize.py` or in `gregoriotex-write.c`), the glyph is assigned to:

 * with ambitus (x) = ambitus between the xth note and the x+1th note
 * 512*shape + 256*liquescence + 25 * ambitus(3) + 5* ambitus(2) + ambitus(1)
 * or if it has only two notes: 512*shape + 64*liquescence + ambitus(1)

so it's a convention that allows gregoriotex to know the glyph number quite easily. The problem is that it takes quite some room, so currently the gregorio glyphs are between code points `u+00A2` (162) and `u+6FC3`, which is much bigger than the number of glyphs, but allows an easy computation. This makes 28611 code points used by Gregorio. Well, it largely fits into PUA-A, so let's do that!

eroux:
A question that needs a quick answer is: do we do that for 3.0.0? This would seem wise to be, but we need to make sure that scores computes with 3.0.0-rc1 will be recomputed.

henryso:
For me, it depends on the potential to break things that are working (other than the font issues from last time).  It doesn't sound like that's too much of a problem, and if that's the case, then I'm fine with 3.0.0.

eroux:
Well it might not be a problem from the user point of view... The only possible problems users might encounter are with other special characters in Unicode that are interpreted by other softwares (like small spaces, etc.). There might be Gregorio glyphs on these... And it would allow to finish things clean and be future-proof... cleaning things is in the air these days, I think it would be a good opportunity.

rpspringuel:
[...]

PR #240, opened 2015-04-09
First attempt at putting glyphs in the UCS PUA-A block.

henryso:
# ~~DO NOT MERGE - DOES NOT WORK~~

I tried to put the glyphs in the UCS PUA-A block to address #239 for release-3.0.

However, I ran into some problems rendering.  It seems like the wrong code points are being used by TeX.  I spot-checked values in the compiled gtex files as well as some of the values in `gregoriotex-chars.tex` and they seem to point to the right code points in the font.

If anyone has an idea what I might have missed, I would appreciate any input.


eroux:
Seems fine in my example, can you be more specific about the problem you encounter? Also, what is you LuaTeX version?

By the way, you can remove `gregoriotex-chars.tex.in` and `generate-gregorio-chars.lua` in this pull request.

eroux:
Also, I don't really understand why, but it slowed down the font generation process quite a lot... not a big deal, but I'm just surprised...

eroux:
Also, maybe you just need to clear LuaTeX cache?

henryso:
By way of example:

```
initial-style: 0;
%%
f.g!hwiHG'F(f.g!hwiHG'FZ)
```

produces

page

This is after clearing the luatex cache.  I'm using luatex 0.70.1, with some backported fixes, as packaged by Gentoo GNU/Linux as part of TeX Live 2012.  I'm beginning to suspect that the culprit is my relatively old version of TeX Live.  However, 2012 is the latest "stable" TeX Live in Gentoo.  This might mean that Gregorio needs a newer version of TeX Live if we put this change in.

It will take me some time to upgrade to the "unstable" TeX Live 2014 since Gentoo is a build-from-source distribution.

I too noticed the slow-down in font generation.  I don't know what's causing that either.

eroux:
It seems so, LuaTeX of TeXLive 2015 works fine, I'll try to investigate and find the minimal working version, then outputing an error if the version is inferior...

eroux:
It seems to be working with LuaTeX-0.76 (TeXLive 2013). I tried to test with TeXLive 2012 binary, but was unable due to other problems (in luaotfload for instance).

henryso:
The upgrade to TeX Live 2014 does indeed make it work for me as well. Do you think TeX Live 2013 is too new a requirement for Gregorio? If not, then I think this change is merge-able.

henryso:
Actually I still need to clean up the conversion files. I'll take care of that shortly.

eroux:
Ok, just two more things then, sorry to bore you with that: can you add a CHANGELOG entry saying that support for TeXLive < 2013 is dropped, and add an error (in gregoriotex.lua, it's the easiest) if compiled with LuaTeX < 0.76? If the error thing is not clear to you, I'll do it.

henryso:
I put the check in `gregoriotex.tex`.  Please review since my TeX is not really all that great.

eroux:
The TeX seems ok to me. About the CHANGELOG, I admit I'd prefer a CHANGELOG that would just read 3.0.0, without all the rc... otherwise all major changes will be in rc1, and final release will be something like "fix bug ...", which is not that significant I believe... What do you think?

henryso:
I don't disagree about the CHANGELOG.  However, I think it's valid to have multiple rc sections so that early adopters can keep track of what's going on and we merge them into 3.0.0 at GA.  Other developers may have other opinions, and I'd like to hear them, though.

eroux:
Sure, we can do this change it when we release final 3.0.0. (What is "GA"?)

henryso:
In this context, GA means General Availability (as opposed to RC = Release Candidate).

eroux:
ok thanx!

PR #242, opened 2015-04-09
Added a note about upgrading TeX Live to 2013 or newer.

henryso:


henryso:
Should I add this to `UPGRADE.md`?

Issue #76, opened 2015-03-12
Missing oriscus with left stem


jakubjelinek:
In Nocturnale Romanum, the standard oriscus like (fo) is often typeset with a left stem like e.g. (fV) has.  This is usually done to typeset virga strata or pressus or salicus.
E.g. in the first snippet I wanted to typeset what was originally virga strata followed by salicus, but
(dxef/e/eoc/co!dv) doesn't do the job (the first snippet - it automatically uses the reverse oriscus), I wonder if e.g. eoVc couldn't be typeset like the second gimp editted snippet.  Or in the third snippet, it is similar, but with just a single pitch difference instead of bigger one, there I've for now used
(f/fo!e) to avoid the reverse oriscus, but it is still missing the stem.
![oriscus](https://cloud.githubusercontent.com/assets/9370665/6625812/b4bc1de8-c8f2-11e4-832f-7ba22a82dd48.png)
Corresponding Nocturnale Romanum pictures:
![nocturnale1](https://cloud.githubusercontent.com/assets/9370665/6626407/b4c0f85a-c8f6-11e4-8d20-2c98b3c45b84.png)
and
![nocturnale2](https://cloud.githubusercontent.com/assets/9370665/6626409/bb6fea58-c8f6-11e4-9b9b-642d95362828.png)

eroux:
This should be relatively easy to add, can you just propose a name for this new glyph (oriscus with stem+punctum after)?

jakubjelinek:
I'm afraid I don't know how it is usually called, and am not good in naming such things.
Semantically it is perhaps pressus (minor, or second/third part of maior), but in the first example
the oriscus with the stem stands for virga strata and the punctum connected to it is already start of the next neume - salicus, with the next oriscus being the oriscus in the salicus and the last virga the upper part of it. Ideally there would be glyphs for different pitch differences of this, so one can use
(foVe) for the second example, (eoVc/codv) for the first one, but say also (hoVf), (hoVe) etc.

eroux:
Sure, that will be done with the same ambitus possibilities as other glyphs

eroux:
By the way, do you plan to reedit the nocturnale romanum or something like that? (that would be great)

jakubjelinek:
No, just typesetting scores for our choir for some feasts and want there complete officium + mass, and Nocturnale Romanum being a useful source for matins:
https://github.com/jakubjelinek/Editio-Sti-Wolfgangi
[...]

PR #245, opened 2015-04-10
Implement oriscus with stem

henryso:
This is my attempt at implementing the oriscus with stem. #76


henryso:
As a first step, I modified the fonts to add the stemmed oriscus "singleton" figures.  Did I do this correctly?

eroux:
That's perfect, thanx!

henryso:
This is ready for review.

eroux:
This is a go for me, I have not time to test but you can merge it if you tested it. The only doubt I have is about the gabc input: I see no problem in changing `O` meaning, but doing something with `v` or `V` might be more explicit? Like `ioVg` or `iovg`? Not sure about it, what do you think?

henryso:
If you feel strongly about it, then I'll do it, but I don't like the idea of having yet another modifier character.  I see capital-O as "attach this oriscus to the previous note," and if there's no previous note, it adds a stem to "attach it" as in the pressus with this figure in the middle (from #76).  I need to do some more testing.  If it looks good, I will merge it later on.

henryso:
Passes tests satisfactorily.  Merging.

eroux:
Sorry, just remembered something missing from this pull request: adding it to CHANGELOG!

PR #251, opened 2015-04-12
Converted defines in gregoriotex into enums.

henryso:


henryso:
This is more enum refactoring.  This code passes the same tests `develop` did before it.  Merging.

PR #254, opened 2015-04-12
Fix to Windows installer

rpspringuel:
Windows installer post installation script was failing.  The changes to remove the TeXworks configuration removed a few lines that were still needed and left a few that were no longer needed.  The result was the script attempting to use variables that weren't declared.
Also added some additional dialog to the post install script to make it easier to tell what is going on.
Clarified the post install option to indicate that it doesn't just install the fonts, it puts all the necessary files into the texmf tree.
Note: The entries I added to the change log are for the purposes of rc2 only.  Since they fix bugs introduced by rc1, they shouldn't appear in the change log for the final release.


henryso:
From an eyes-only review (since I can't test it), this looks good.  Merging.

eroux:
@rpspringuel thank you very much! Sorry for breaking it...

Issue #238, opened 2015-04-08
Spacing problem before stars under bars


jperon:
Perhaps it's related to #221 ? There's not enough space before stars that are under bars. See the example joined, especially in the verse.
![00-document](https://cloud.githubusercontent.com/assets/4775231/7042530/d64018f2-dde2-11e4-96a9-d4251e839a9a.png)

eroux:
What version/branch do you use?

jperon:
I use the default one when pulling the repo.

eroux:
do you specify the `develop` branch or is it the `master` branch that is checked out?

jperon:
It's the master one.

rpspringuel:
When did you last update?  The master branch was just updated this past weekend with the 3.0.0 release candidate.  Is that what you're using?

If it is and you're not a new user, did you see this behavior before updating?

Finally, please post a minimal example of the `gabc` that reproduces the problem.  The main `tex` file might also come in handy.

jperon:
Yes, I'm using a recently updated master branch. I didn't see this
behaviour before updating.
You'll find a mwe joined to this mail. I just let the first half of the
verse, that's significant enough.

Writing this mwe, I saw two other bugs : the default size of the initial
(at least with libertine) seems too small, and there should be an hyphen
within "Exal*-*tábo". Should I open other issues for those ?

2015-04-08 17:50 GMT+02:00 Br. Samuel Springuel <notifications@github.com>:

jperon:
The file…

2015-04-09 11:53 GMT+02:00 Jacques Peron <abbe.jacques.peron@gmail.com>:

eroux:
Sorry, I don't think you can attach any type of file in github messages (never tested before, sorry), can you send this to `gregorio-devel@gna.org`?

eroux:
Thanks, I believe it's also visible on the example in the repository... I just didn't manage to get the output of the example on an old repository... @rpspringuel as you did most of the code about spaces, would you have time to take a look?

rpspringuel:
I'll take a look later today or tomorrow.

rpspringuel:
@jperon, on the other two issues you noted:

[...]

PR #255, opened 2015-04-12
Bars with something under them were appearing too close to previous element

rpspringuel:
I found a commented out kern statement which when uncommented increases the spacing in the reported MWE.  However, checking the log, this line has been commented out since its introduction back in 7e2bfa4 (which was in 2008).  As a result, I'm not sure why this bug would not have appeared in earlier versions.  My only guess is that there were some missing `%` at the end of lines which resulted in the addition of uncontrolled whitespace and so while the line was added because theoretically it was needed, it was commented out because visually it resulted in too much whitespace.  As part of my spacing changes, I did a whole bunch of work to make sure that most, if not all, of the lines in the package terminated with a `%`, so that might have caused the problem to show up.  Unfortunately, my progressive testing seems to indicate that this bug shows up about the same time as the glue leaks, making it difficult to tell exactly what was going on at that time because I can't get a clean output.
For no discernable reason, `\GreSyllablenotes` has the wrong width for bar lines.  To get around this I take advantage of the fact that `\grewritebar` sets `lastglyphwidth` as the last step of its process and sets it correctly.  I thus make use of `lastglyphwidth` instead of `\wd\GreSyllablenotes` in `\grebarsyllable`.
Added the new english centering code in `\grebarsyllable`.

### Not ready to merge.

Addresses #238.  The solution for the bad width is somewhat hackish, so I'd like comments and suggestions for making it less so if possible before merging.

### DIFF
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -245,6 +245,7 @@
 \newbox\GreSyllablenotes%
 \def\gresyllablenotes#1{%
   \setbox\GreSyllablenotes=\hbox{#1}%
+  \gre@debug{Width of notes: \the\wd\GreSyllablenotes}%
   \relax%
 }%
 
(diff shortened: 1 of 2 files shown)


eroux:
I'll read that tomorrow in the morning (reading TeX is not something easy late in the evening), but what you're saying about a space in the code is very possible... So thank you very much for investigating it, I know how hard and unfun it can be to debug this kind of things, but fixing this kind of old bugs is really a good thing, and it will certainly improve many outputs (even if the difference will be subtle).

Thank you very much also for the redirects to the new website!

I just need to check if scores will be rebuilt between rc1 and rc2, but after that we're ready to release. I'll do it tomorrow in the morning, and after that you can upload the Mac and Windows installers when you have time.

Thanks again!

tex/gregoriotex-syllable.tex
```diff
@@ -648,42 +649,50 @@
...
   \divide\gre@dimen@notesaligncenter by 2\relax %
   \gre@dimen@begindifference=\gre@dimen@notesaligncenter %
   \advance\gre@dimen@begindifference by -\gre@dimen@textaligncenter %
-  \gre@calculate@enddifference{\wd\GreSyllablenotes}{\wd\GreSyllabletext}{\gre@dimen@textaligncenter}{\gre@dimen@notesaligncenter}{1}%
+  \gre@calculate@enddifference{\gre@dimen@lastglyphwidth}{\wd\GreSyllabletext}{\gre@dimen@textaligncenter}{\gre@dimen@notesaligncenter}{1}%
   #5%
   \gre@calculate@nextbegindifference{\gre@nextfirstsyllablepart}{\gre@nextmiddlesyllablepart}{\gre@nextendsyllablepart}{#7}%
   \greunsetfixednexttextformat %
-  % for LuaTeX syllable counting in optimize_gabc, we put an empty syllable box with gregorioattr set to 2. With this we can count every syllable.
-  \gregorioattr=2\relax %
```
eroux: Why did you remove that?

eroux:
It would be more practical for me to be able to add commits to your pull requests... What about always making pull requests from main repo? @henryso what do you think?

henryso:
My preference is to always start these things in a user's private repository.  This way, the workflow is more or less the same whether you are a collaborator or not, the only difference being the merge into the team repository.

For small changes, I'm in favor or creating a pull request to the other user's branch.  For example, let's say I have created henryso/gregorio:bigfix and a pull request from this branch to gregorio-project/develop.  You pull my branch into your repository eroux/gregorio, make changes to it, and the push them to your repository.  Then you create a pull request from eroux/gregorio:bigfix to henry/gregorio:bigfix.  Then I merge your changes into my branch, thus adding them to my pull request.

However, if bigfix turns out to be a collaborative effort, then I'd create a new branch in the main repository gregorio-project/gregorio:bigfix, create a pull request from henryso/gregorio:bigfix to gregorio-project/gregorio:bigfix, merge that, close the original pull request, and then create a new pull request from gregorio-project/gregorio:bigfix into gregorio-project/gregorio:develop.  I can then pull this branch back from gregorio-project/gregorio, work on it some more, and create/merge a pull request from henryso/gregorio:bigfix to gregorio-project/gregorio:bigfix to add to the big pull request.

That said, this is just my opinion.  I will work however you want me to work.

eroux:
Ok, let's do that. @rpspringuel just answer the comment and I'll merge #258 with this one

eroux:
By the way, I don't think bars are aligned like regular syllables, but always on the center of the global text, wherever the vowel is... Also, the algorithm should change a bit in next release (allowing text and bar not to be completely aligned, but with text centered between surrounding text, and same for bars, so I don't think using the same mechanism for centering as `\syllable` is really necessary, nor really good... Maybe #258 would be a better basis to continue debugging? The behaviour should be similar to the one you introduced here.

tex/gregoriotex-syllable.tex
rpspringuel: Overlapping thought processes.  You mentioned elsewhere that the `optimize_*` stuff was being removed.  Since these lines are related to that, I removed them.  That removal, though, doesn't strictly belong here, but with the removal of the `optimize_` stuff.

eroux: Oh sorry, I didn't even read the comment... right, this can disappear

rpspringuel:
It's easy enough to change the TeX so that the text and bar are always centered on each other.  Some change was necessary, though, as before my changes the behavior was always Latin alignment.

Might it be better if `\grebarsyllable` didn't take the text argument in parts?  It would require some rejiggering of the C, but would simplify things on the TeX side.

eroux:
Indeed... we must do it before 3.0.0 then, because that would make an incompatible change

PR #258, opened 2015-04-13
Fixtex

eroux:


eroux:
This fixes a bug in `\grelastispunctum` that introduced a bug in bars (not really big though), and factorizes some code. @rpspringuel can you merge you pull request with this one? Or integrate your changes here?

Also, I know why there is a difference between `\wd\GreSyllablenotes` and lastglyphwidth: when the last glyph before the bar contains a punctum mora, the bar is shifted left (see beginning of `\grewritebar`), so the width of the syllable notes becomes negative... I don't really know why it produces such a bad spacing (I don't believe it should), but it should give you a clue about what's going on... I don't have time to debug this more today, but tell me if it helps you, I can continue to debut it tomorrow.

eroux:
Also, `\debug` is really something dangerous, many other packages might provide it, so I changed it to `\gredebug`

rpspringuel:
No time to debug, but these changes cause the bar lines to disappear entirely for me.  Incorporating my changes doesn't fix that.

Including the leftwards movement in the width of the bar messes up the alignment because the calculations for `notesaligncenter` assumes that it's getting the actual width of the bar line.  The central alignment point for the notes is supposed to be half-way through the bar symbol.  If the calculations for `notesaligncenter` are fed something other than the actual width of the bar symbol, then point on which GregorioTeX attempts to align the bar symbol won't be the center of the symbol.

What we should probably do is move that kerning code out of `\grewritebar` and into `\grebarsyllable`.  That code is related to placement of the bar line, not the actual printing of the bar line.

eroux:
The fact that bars disappear for you is really weird, I don't have this when I test... Anyway, the code to shift left works for cases where the bar is inside a syllable with other elements (eg `ab(gh'hg)`), so it should stay there for these cases,but you're right, it should probably also be copied to `\grebarsyllable`. I think we should do that for 3.1, in the meantime, 3.0 won't have the shift left when there is a punctum mora, I don't think it's a release critical bug. What do you think?

I'll try to find time to debug this tomorrow... quite a nasty bug

rpspringuel:
I agree that the final clean-up of the spacing algorithms can wait for 3.1.  What we just need at this point is enough of a fix to get things working and looking approximately right.

rpspringuel:
Hmm... Just noticed another thing that disappeared in my tests: the punctum on the the preceding punctum mora.  That got me looking at the branching.  My fix branches directly from the release candidate, yours branches from the release-3.0 branch after @henryso's changes to the character numbers was effected.  I'm willing to bet that said changes have introduced a mapping problem between the fonts I have installed and the code points that exist on your branch.  I don't have time right this second, but I'll try reinstalling from your branch and then testing again.

eroux:
Seems I caught the bug... If someone can review it, I think we can release after this gets merged. @rpspringuel I really like the `\gre@debug` macro and now use it quite a lot, but I think the messages are not that helpful now, can you take a look at the messages I introduced in this pull request and tell me what you think? The idea is to print what Gregorio globally does (new syllable, add hskip, etc.) instead of the tests. Also, I introduced indentation, to get `New syllable` with no indentation, but everything happening inside syllable indented with two spaces, so that messages are more readable... It makes debugging faster, it took me very little time to find the bug thanks to that!

rpspringuel:
I haven't systematically added debug messages except in cases where I was tracking down a bug. Thus, since there was a bug in the `\ifdim` tests at one point, they all got marked.  Since I never remove the debug messages (that's part of the reason for having the function) those marks are still there, even if they aren't as useful for telling where you are as other possible messages.  At some point what I'd like to do is get every macro to declare itself and its arguments, but I haven't had the time to do that yet. 

As the messages proliferate some conventions on formatting will definitely make reading the easier.  I don't have time to test and integrate my changes (if they are still needed) right now, but I'll try to get to it later today.

eroux:
Maybe `\gre@debug` could take two additional elements:

* the verbosity level above which message should be displayed (with a `\gre@debug@setverbosity` to set the level)
 * the type of debugging, I don't know exactly what the different types could be yet, maybe `syllable`, `glyph`, `spaces`, `bars`, `initial`, etc.

What do you think? This won't be for 3.0 anyway, so let's release it first...

rpspringuel:
Okay, after testing, I find that the new code works and makes some of my changes unnecessary (in particular the use of `lastglyphwidth` instead of `\wd\GreSyllablenotes`).  I've cherry picked in the changes that do make a difference and have something that works well on my machine.

eroux:
As I told, I'm not really in favor of having the same centering mechanism as in `\syllable`, can you please revert it?

henryso:
I think this change makes all the pdf-based tests fail, though in an expected way.  I will have to visually confirm each pdf, so it might be a little while before the test project will pass things.

eroux:
Yes, spacing should be a bit different now, it should actually improve a little... But I don't think it's blocking, I think we can release RC2 when PUA-A changes are done if you agree...

eroux:
Sorry, there is indeed a bug somewhere, looks like a spurious space in the code...

henryso:
I am noticing that many hyphens between syllables appear to be missed.  Perhaps this is a symptom of the issue you mention above.

eroux:
I think so

tex/gregoriotex-syllable.tex
```diff
@@ -251,8 +251,21 @@
...
+% saves and restores a few values so that when really typesetting #1 in \gresyllable,
+% the result is identical.
 \def\gresyllablenotes#1{%
+  \global\grefirstglyph=1\relax %
+  \xdef\grelastispunctumsauv{\grelastispunctum}%
+  \xdef\greboxing{1}%
+  \xdef\gresavedlastoflinecount{\number\grelastoflinecount\relax }%
   \setbox\GreSyllablenotes=\hbox{#1}%
+  \gre@debug{Width of notes: \the\wd\GreSyllablenotes}%
+  \xdef\grelastispunctum{\grelastispunctumsauv}
```
rpspringuel: Need a `%` at end of this line

eroux: Thanx a lot! I'll add that in a pull request in a few minutes

PR #262, opened 2015-04-13
Fix lex

eroux:
@rpspringuel can you please test this when you have time (especially the removal of the linking with libfl)? I'm particularly interested to know if it still works under Mac and Windows... This would clean the binary even more than it is now, and would ease integration in TeXLive...

### DIFF
--- a/fonts/Makefile.am
+++ b/fonts/Makefile.am
@@ -56,8 +56,7 @@ gresym: gresym.ttf
 
 greextra: greextra.ttf
 
-clean-local:
-	@$(TOP_LEVEL_MAKE) $(RM) -f $(TTFFILES)
+DISTCLEANFILES = $(TTFFILES)
 
 install-data-local:
 	@$(TOP_LEVEL_MAKE) echo 'The Makefile in the fonts directory is no longer used for installing'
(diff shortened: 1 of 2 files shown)


rpspringuel:
When testing this is there some particular combination of options you want me to build with (besides the ones necessary for the platform)?  I ask because I normally build with ` --enable-all-static --disable-shared --enable-static-ltdl` when preparing the MacOSX installer.

eroux:
In the `develop` branch, these options do nothing, `libltdl` is not used and there is no non-static things anymore, so no particular option...

rpspringuel:
Works on Mac.  Will test on Windows tomorrow.

A request: can an option be added to `build.sh` that will force it to rerun `autoreconf`.  I had a hang-up in testing this that required rerunning `autoreconf` (some files that are on `.gitignore` on some branch but not another and which I deleted when switching branches had to be rebuilt).  It took me a few minutes to figure out that `build.sh` wasn't rerunning `autoreconf`, but once I knew that things worked.  Having an option that forced `build.sh` to go through all the steps, regardless of whether or not it thought them necessary, would have made things just that little bit easier.

henryso:
I'm working on the script now.  Expect something within the hour.

rpspringuel:
Works on Windows.  Merging.

PR #263, opened 2015-04-13
Pylinting - WIP

eschwab:
**Work in Progress - Do not merge**

Since we have established code quality guidelines, I am trying to make our code fit those guidelines.


fonts/squarize.py
```diff
@@ -210,144 +243,164 @@ def initialize_glyphs():
...
+    "Initialize the lenghts, depending on the font"
+    global BASE_HEIGHT, LINE_WIDTH, WIDTH_PUNCTUM, width1, width2
+    global WIDTH_DEBILIS, WIDTH_DEMINUTUS, WIDTH_INCLINATUM_DEMINUTUS
+    global WIDTH_FLEXUSDEMINUTUS, PORRECTUSFLEXUSWIDTHS, PORRECTUSWIDTHS
+    global WIDTH_INCLINATUM, WIDTH_STROPHA, HEPISEMUS_ADDITIONAL_WIDTH
+    global WIDTH_ORISCUS, WIDTH_QUILISMA, WIDTH_HIGH_PES, WIDTH_ORISCUS_REV
+    if (font_name == "gregorio"):
         # the base heigth is half the space between two lines plus half the heigth of a line
-        base_height=157.5
+        BASE_HEIGHT = 157.5
```
eroux: BASE_HEIGHT could be a global variable, Gregorio won't accept a different value anyway...

fonts/squarize.py
```diff
@@ -210,144 +243,164 @@ def initialize_glyphs():
...
-        width_deminutus=88
-        width_inclinatum=173
-        width_stropha=164
-        width_high_pes=154
-        width_inclinatum_deminutus=128
-        width_flexusdeminutus=186
-        porrectusflexuswidths=(340,428,586,670,931)
-        porrectuswidths=(490,575,650,740,931)
-        hepisemus_additional_width=5
+        LINE_WIDTH = 22
```
eroux: I'm affraid this (global variables) is quite anti-pattern... What should be done here is a `width` hash passed to the other functions. Quite intrusive but will be much better

eroux:
Thank you very much! Can you merge develop into your branch so that it can be merged automatically? (after fixing a few things)

fonts/squarize.py
```diff
@@ -379,913 +433,991 @@ def end_glyph(glyphnumber):
...
-    if (first_glyph=="qbase"):
-        width1=width_quilisma-line_width
+    if (first_glyph == "qbase"):
+        width1 = WIDTH_QUILISMA-LINE_WIDTH
     else:
-        width1=width_punctum-line_width
-    write_line(i, glyphnumber, width_punctum-line_width, base_height)
-    paste_and_move("rdeminutus", glyphnumber, width_punctum-line_width-width_deminutus, i*base_height)
-    set_width(glyphnumber, width_punctum)
+        width1 = WIDTH_PUNCTUM-LINE_WIDTH
```
eschwab: This `if` statement is currently useless. Should the next line (`590`) use `width1` rather than `WIDTH_PUNCTUM_LINE_WIDTH` ? This would be consistent with line `562`.

eroux: Sounds right

eschwab: Actually there is no need to keep `width1` as a global. It is only used twice and defined prior to each use.

eroux: Sure! The best would be to remove all global variables, except for things like MAX_INTERVAL, etc. that are really global

eschwab:
Before passing the `font_width` dictionary to all the functions that use a width, I just want to make sure that doing this will not adversely effect the efficiency of the script. I don't know how python handles function args, if it passes a copy or a pointer to the data.

eroux:
Don't worry, the pure python things are certainly around 5% of the cpu usage in the script, the rest is fontforge... and python won't copy the dictionary every time you pass it to a function, so no problem.

eroux:
This starts to look much better, thank you very much! Tell me when you're ready, I'll merge

eschwab:
I destroyed this PR by merging in `release-3.0` rather than `develop`.

Issue #266, opened 2015-04-14
Fonts under TeX Live 2013


henryso:
John Perry (from the mailing list) is having a problem with the new font encoding in PUA-A.  He is using LuaTeX 0.76.0 under TeX Live 2013.  I tried out his generated .tex file using LuaTeX 0.79.1 under Tex Live 2014, and everything renders correctly.  I have confirmed everything is installed properly and the fonts are encoded properly under PUA-A and that there is only one version on his system.

Can someone here confirm this as working or not under TeX Live 2013?  I can try to set up a VM or something, but that will take me some time.

If it doesn't work under 2013, should we consider rolling back this change until 2014 is in more common use?

henryso:
John Perry was able to get things working by upgrading to TeX Live 2014.  If we decide not to rollback the encoding, we should at the very least bump the required version to TeX Live 2014 and LuaTex 0.79.  But maybe 2014 is too recent to be a reasonable requirement?

henryso:
Does anyone know how I can get Tex Live 2013 ?  I'd like to try to verify that it really doesn't work under that version.

avbop:
It might be painful to install, but ftp://tug.org/historic/systems/texlive
seems to have the files.

>

rpspringuel:
I generally have the 3 most recent versions of TeXLive on my machine and can switch between them very easily.  Testing this, I can confirm that switching to TeXLive-2013 on my machine does cause things to go bonkers using the new font encoding (i.e. wrong glyphs show up in the score).  Switching back to TeXLive-2014 makes things work again.

On the issue of rolling things back, since TeXLive is free, I don't have a problem with requiring TeXLive-2014 or later.  It's been out since June of 2014 (a good 10 months) and TeXLive-2015 is in the process of undergoing tests for the public release in June (at which point 2014 will be unsupported as far as package updates go).  It might be slightly inconvenient, but shouldn't present a hardship to require 2014.

However, if we do roll things back for now, I would say that we should plan on re-implementing the font encoding change once TeXLive-2015 is out.

eroux:
Reverting the change to PUA for 3.0 is ok for me

PR #267, opened 2015-04-14
Upgraded requirements to TeX Live 2014 and LuaTex 0.79.1.

henryso:
If we don't roll back the font changes, it looks like we may need to at least require this version. See #266


henryso:
Rather that bumping the requirements, I will roll back the shift to PUA-A.

Issue #272, opened 2015-04-15
gregoriostylefont character references in gregoriotex-chars.tex are wrong


henryso:
While reverting the change to fix #266, I noticed that `gregoriotex-chars-translate.lua` indiscriminately updated the values for this font which is not offset into PUA-A.  However, undoing this to the version prior to that change does not appear to be selecting the right characters out of `greextra.ttf`.  If I subtract 161 (the old offset), the characters *still* seem to be off.  Does anyone know what the correct values are for the `\gregoriostylefont`?

eroux:
Good point... I admit it has been a few years since I haven't opened this font though, sorry...

henryso:
If I had to guess, the offset is 66.  I will use that for now unless someone knows better.

PR #276, opened 2015-04-15
Revert the change placing the glyphs into PUA-A

henryso:
In addition to reverting, I switched to using a LuaTeX macro to compute the offset so that making the switch later will be less painful.


henryso:
This also addresses #272.

henryso:
The switch to numexpr is done and it seems to work.

eroux:
Sorry, forgot to merge

PR #281, opened 2015-04-15
Hardcode the gregorio version in all the *.tex files.

### LINKED ISSUE #271: TeX file versions
I never realized that replacing the hardcoded versions in the different `gregoriotex-*.tex` by

    \gredeclarefileversion{gregoriotex-spaces.tex}%
     {\directlua{tex.write(gregoriotex.get_gregorioversion())}}%

changed the behaviour and made the process almost pointless: this means that an outdated `gregoriotex-*.tex` won't be detected anymore, as it will just use the version of the not-outdated `gregoriotex.lua`... So I think we should switch back to the old hardcoded versions and add the different files to `VersionManager.py`... What do you think?

This could be for 3.1 or 3.0, I don't know...

eschwab:
Fixes #271

### DIFF
--- a/VersionManager.py
+++ b/VersionManager.py
@@ -19,13 +19,19 @@ from distutils.util import strtobool
 
 VERSION_FILE = '.gregorio-version'
 GREGORIO_FILES = ["configure.ac",
+                  "plugins/gregoriotex/gregoriotex.h",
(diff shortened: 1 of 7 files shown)


henryso:
Looks fine to me.

PR #283, opened 2015-04-16
Fixpylint - replaces #263

eschwab:
I think this is ready. Not 100% passing for pylint, but much better than before.


eschwab:
@eroux could you double check that I got shortglyph flag right for all the shapes? Go ahead and merge if it looks good.

eroux:
Seems right, thank you very much for this!

Issue #279, opened 2015-04-15
Hard coded character code points in gtex files


rpspringuel:
Right now gregorio hard codes the character code point for most (but not all) glyphs into the gtex file when it creates it.  This means that there has to be coordination between the `.ttf` files, GregorioTeX, and the command line tool for those code points.

Would it be possible to reduce the amount of needed coordination by having the commandline tool use the character names defined in `gregroriotex-chars.tex` instead of hard coding the code points?

If not, can some sort of check be established that would verify that all three are working from the same character code points?

Heck, even if the coordination can be reduced, it might be nice if some sort of check can be built into `gregoriotex-chars.tex` that makes sure the code points defined there match the font.

henryso:
I have already modified the code to use \gregoriocharoffset to eliminate coordination of the offset with the tex files.  The next level is something Élie has mentioned to use macros and have the TeX code to do the computation of the code point based on ambitus.  I'm not sure this is necessary, but it is worth discussing.

eroux:
I've been wondering about this for a few years, the main reason why I didn't do this is because I have the weird belief that it will slow down computation a little bit... But this might not be justified at all, and there should be some tests about that... I think it's misplaced optimization, a few integer additions are certainly neglectible compared to the font handling... But I don't have a very long document to test compilation time, and I know some do books of around 1000 pages with Gregorio, I wouldn't want to make their computation ubearably slow... In order to bring some rationality, maybe there could be an extremely long document in gregorio-test, with a special test target about computation time? Of course it couldn't be automatic, but it would allow to test if some changes introduce slowness. What do you think?

henryso:
Do you know of a large non-copyright-protected document?  Should I just grab a bunch of scores from gregobase?

rpspringuel:
@henryso, the use of `\gregoriocharoffset` certainly makes changing the offset easier, but the code points are still coded into `gregoriotex-chars.tex`.  There just in two pieces: the offset and the relative position within the character block.  This then has to be coordinated with the `.ttf` files and with the numbers which are written to `.gtex` files (though I think that these latter numbers are also calculated from the offset and relative position, making coordination easier).  Further, the files don't currently talk to each other to check that they are coordinated.

[...]

Issue #270, opened 2015-04-15
Fonts metadata


eroux:
There seems to be some metadata missing from the generated fonts, here are a few ideas of improvements (not really urgent)

 * version in the `.sfd` files are old, they should be updated by `VersionManager.py`
 * no version is present in the TTF, `squaryze.py` should write it
 * there is no Unique ID nor XUID
 * I'm not sure about the OS/2 character set (nor what it does)

eroux:
When the version will be accessible in the font, GregorioTeX should check if the versions match between the TeX and font.

henryso:
What remains to be done here?

eroux:
I think it's ok now

PR #292, opened 2015-04-18
Remove hard-coding of code points

henryso:
This is a fix for #279.

Thus far, I've only done the main glyph fonts.  I still probably need to do at least the `greextra.ttf` font to eliminate the remaining code points in `gregoriotex-chars.tex`.

In addition to this refactor, I also took the opportunity to eliminate interposed blank glyphs since the glyph number formula would no longer be needed.  This allowed me to fit the glyphs in the BMP PUA using ISO10646-1 encoding, which then  did.

I have not done much performance testing, but it appears to be pretty similar in my document set.

All tests pass except the gabc-to-gtex tests, which I expect to fail due to this change itself.

Please review.  If possible, please try this in TeX Live 2013 as well and let me know the outcome.


eroux:
This seems really good! Maybe the lua font mapping could be cached? Not sure it would improve much the compilation speed though... I'd prefer to have a performance test before merging, but I don't believe it would discard this, so if there is no issue with TeXLive  (which I cannot test), I think it can be merged!

Off-topic: I'm currently playing with docker for my job, I just thought I could make some Dockerfile with different versions of TeXLive... I'll see if I have time for this soon, if so I'll share them.

henryso:
I assume that for caching, you mean to disk?  I believe it is already caching the mapping to a Lua table, so it shouldn't need to build the table more than once per run.

I am considering having `squarize.py` (or a new script, perhaps) produce a mapping file which would basically be a bunch of TeX macros with all available glyphs.  My problem is in detecting the "current font" from pure TeX.  I'm assuming it's possible, but I'm still limping around TeX, so I need more time to search for the solution.

eroux:
I meant to disk yes, but that's not priority (and that might cause more trouble in upgrading)... You can check `\gregoriofontname` yes, it shouldn't change in the middle of a score. `font.id(gregoriofontname)` should even give you the ID according to [LuaTeX manual](http://www.luatex.org/svn/trunk/manual/luatexref-t.pdf#53). But if you are in a context where you already called `\gregoriofontname`, and hence the current font is correctly set, `font.current()` should give you its id...

henryso:
Performance is definitely an issue.  The time to compile 1000 of PopulusSion went from 1:20 to 3:10.  I don't know how much of this has to do with the string handling (in TeX) versus Lua lookup.  I'm going to play with the pure-TeX route (and thus will use `\gregoriofontname` rather then `font.current()`) to see if I can make performance more reasonable.

henryso:
With the mapping pre-computed and loaded into TeX, it takes 2:35, to run 1000 of PopulusSion.  Better, but still about double the time compared to before.

Any suggestions on improving performance while still keeping the mapping in TeX ?

henryso:
Is it worth exploring the possibility of generating a header file and using that in C?  The code points would then be hardcoded again in the gtex output.

eroux:
Something that would marginally improve things: the difference between the three font mappings are only a few glyphs in the initial font, maybe starting to generate glyphs at, say, initial code point + 100 would allow the three mappings to be the same and allow removal of font check? Not really the core of the problem, but it might simplify things a little...

eroux:
What about something like writing `\torculus{1}{2}{0}` in the `.gtex` file, and have the `\torculus` macro do `\char xyz` where `xyz` is computed from the three arguments (first two are the two first ambitus, third is liquescentia)? That's what I was initially thinking about... It still requires sync between squaryze.py and the TeX macros, but at least C doesn't have to know it, and the `.gtex` file is independant of the initial codepoint... It is less generic than your solution, but maybe it would be faster? What do you think?

henryso:
I think if I factor out the font differences, evaluating a macro that evaluates to a character should be faster than evaluating a macro that does math.  I'm trying this now.

henryso:
I factored out the font differences and boiled it down so that the gtex looks like `\greglyph{\grecpPesThreeNothing}{e}{g}{0}{}{}` and `\grecpPesThreeNothing` is a `\chardef`.  It should be a direct substitution in the TeX engine.  However, all this only saves about 5 seconds, so clearly the bottleneck is elsewhere.  Note that this also cut the mapping macros to one third, will little effect to the running time.

That switching to macros which do a computation to find the code point might improve performance is a surprising hypothesis to me.  However, I'm no TeX-expert.  Switching is a fairly hefty change (not that what I did today wasn't) and involves more TeX, so I want to be sure there's a good possibility that this more complex solution will (against my intuition) perform better.

eroux:
It won't work better than this indeed. Can you try `\xdef` instead of `\chardef`? Just to be sure...

henryso:
I have tried `\def`, `\edef`, `\gdef`, `\xdef`, and `\chardef`.  The difference is negligible.  I just happened to have `\chardef` last and that's what got in.

eroux:
I'm affraid I don't see any possible speed improvement for now...

henryso:
So the question is, how do we proceed?  These are the options I see:

- Abandon this branch.  This is reasonable because doubling the run time is a significant performance degradation.
- Merge this branch as-is.  The advantage is that code points are hard-coded in TeX but not in C and that the glyphs fit in the BMP PUA, so we don't need to use the slow-to-generate PUA-A and maintain support for TeX Live 2013.  The disadvantage is the performance degradation.  There is also one additional point of failure if the font and the `gregoriotex-fontmapping.tex` don't match, but this isn't much worse than the font and `gregoriotex-chars.tex` not matching in the current version.
- Go back to the original version of the branch (revert https://github.com/henryso/gregorio/commit/8229f58d7a40f2622793b3a7f78ea6c4b113bd22, https://github.com/henryso/gregorio/commit/809a327ccf725c1d01f3919e442fbac53c2aefb6, https://github.com/henryso/gregorio/commit/c4e89590b0e65401d60348fd9c107749654f7ded).  The advantage is that the code points are not hard-coded and are will be discovered at runtime.  The disadvantage is even more of a performance degradation due to escaping into Lua to do the mapping.
- Generate a C file with the code points and use them in C.  This should return performance back to current levels (or identify what really caused the bottleneck).  The advantage is that we can still fit the glyphs in the BMP PUA, avoiding slow font generation and maintaining support for TeX Live 2013.  The disadvantage is the code points remain hard-coded both in TeX and in C, so if `gregorio`, GregorioTeX, or the fonts don't match, the system will fail as it does now.

Thoughts?

henryso:
I also forgot to mention this general advantage to any of the above except abandonment.  In addition to fitting in the BMP PUA, there is no longer a mathematical formula, so it is easier to add new shapes.

eroux:
I'd be in favor of 4th one: doubling compilation time seems really wrong, and abandoning would be a bit sad... Maybe, as you say, we'll discover the bottleneck is somewhere else!

rpspringuel:
I think the 4th option is at least worth a test, though I'm not the one doing the coding here.

Would it be possible for `squarize.py` to generate this C file?  That would require generating the fonts before compilation, but given that most users aren't generating the fonts on install (they're using a distribution in which the fonts have already been generated) that really only affects us.  The advantage to this would be that generating the fonts would then automatically update the mapping file.

As for the point of failure, is it possible to access a font's version information using the LuaTeX font loading?  If so, and if we get `squarize.py` to add the version information to the font, then we could get TeX to output a specific error message about the mismatch between the installed fonts and the score rather than simply having wrong or missing glyphs.  That would at least be more informative to the end user.

eroux:
Indeed, the generated C file should be included in the tarball. About font version, I added a comment on #270.

henryso:
I'll work on this today and see what I can come up with.

henryso:
Since `squarize.py` is called once per font, I think it might be more appropriate to generate the C file in `mapfonts.lua`, which is only run once, after the fonts are generated.

henryso:
After coding the changes to generate the code points for C, I found that the timing was still about the same.  This led me to the actual cause of the problem, which is my testing methodology.  Now that I have this sorted, I am going to try running the various versions of this change in an apples-to-apples comparison and will report back my results.

henryso:
This is not really a scientific test because the test is not in complete isolation and I didn't take all that many data points, but I first cleared the font cache (`luaoftload-tool --cache=erase`) and then ran each 5 times and took the average time, rounded to the nearest half-second.  The test consists of 1000 of PopulusSion in a single document.  Therefore, this is more of a TeX test than a C test.  The variance from these average times (per test) is about one second.

- base version: https://github.com/gregorio-project/gregorio/commit/aad1c46ee75d492bef4729c08cfcf057f3163c29 - 1:20.5
- Lua-based dynamic mapping: https://github.com/henryso/gregorio/commit/f1076b3819db3ff7994585e4ed31d9f606a121d1 - 1:35.0
- TeX-based pre-generated mapping with fonts having disparate code points: https://github.com/henryso/gregorio/commit/809a327ccf725c1d01f3919e442fbac53c2aefb6 - 1:18.0
- TeX-based pre-generated mapping with unified code points: https://github.com/henryso/gregorio/commit/8229f58d7a40f2622793b3a7f78ea6c4b113bd22 - 1:16.5
- C and TeX-based pre-generated mappings: https://github.com/henryso/gregorio/commit/868a587ece5a14b2d1bb256498782aa7940155fa - 1:16.5

The last two also benefit from increased scrutiny to performance.

So while not as bad as I originally thought, the Lua-based mapping is still 15 seconds (18%) slower than the base base version.  The TeX-based mapping with unified code points is certainly a reasonable option (5% faster).  The C-based mapping performs more or less the same, but the application is likely more efficient, doing away with string manipulation and using pointer arithmetic instead.  However, the C code hasn't really been a performance issue, so it's a question of our goals here.

Being so entrenched in the code, I think I've lost a bit of perspective, so any advice on direction would be appreciated.

eroux:
This is very good news! I like the TeX-based mapping with unified code points too, I'm just wondering how it can be faster than the original, did you optimize something in another part of the code?

henryso:
Yes, I stored the chars rather than the code points in the `\gre@char@` macros.

eroux:
Great! Do you have more work on this or can it be merged?

henryso:
I'd like to give it another cleanup pass and perhaps experiment with one more attempt at a Lua-powered dynamic mapping, where Lua generates the `xdef` for the characters at runtime.

rpspringuel:
@henryso, now that you've ironed out your testing procedure, do you think you could incorporate that test into `gregorio-test` some how?  A performance test probably shouldn't cause an outright test failure, but if it could compare itself to some sort of benchmark, it would allow one to see how changes are affecting performance.

Just to make sure I understand the various cases correctly:

1. base: our current situation with the font, C, and TeX having to be coordinated manually.  The C hard codes the code points into the gtex file and TeX has hard coded code points sprinkled through the various files (though most are in `gregoriotex-chars.tex`).
2. Lua-based dynamic mapping: TeX and C use glyph names and are thus independent of fonts.  Inside each font, the glyphs are named.  When typesetting, TeX calls a Lua script which retrieves the glyphs based on the glyph name.
3. TeX-based pre-generated mapping with fonts having disparate code points: C uses glyph names, TeX uses code points.  Map of names to code points is generated on build/install as the fonts are generated (thus ensuring that the two match) and loaded when typesetting.  Each font has it's own map.
4. TeX-based pre-generated mapping with unified code points: Same as 3, except all fonts use the same map.
5. C and TeX-based pre-generated mappings: C and TeX use code points. Map of glyphs to code points is generated on build/install as the fonts are generated (thus ensuring the two match).  C uses this map to put code points in gtex and TeX uses the map to retrieve the right character when typesetting.

Assuming the above is right, I agree that option 4 is best if you can't solve the performance issue with dynamic mapping.  The reason I like this option are as follows:

1. gtex files contain glyph names, not code points.  This makes them easier to read (and thus hack when you're trying to do something that the gabc doesn't allow).
2. Font changes won't necessarily require reinstalling the command line tool.  This should make testing easier for us.

If you can solve the performance issues in the dynamic mapping situation, then I would have to reconsider.

henryso:
Regarding the performance tests, I need to think about how this can be incorporated.  Part of the testing procedure problem was that if I use a not-installed GregorioTeX, the timings all go haywire.  I don't know why this is, so for now, performance tests can only run from an installed GregorioTeX.

Regarding the cases, you have summarized them perfectly, except I think all the hard-coded code points were, as of "base", in `gregoriotex-chars.tex`.

The "one more" experiment that I mentioned earlier is basically an attempt to generate a run-time version of `gregoriotex-mapping.tex`.  I don't know what the impact will be until I try it, but I will post my results here.  My biggest obstacle is figuring out where I can/should hook the call to Lua to do the mapping.

henryso:
If I break out and generate the `\chardef` font mapping dynamically from the font using Lua, I pretty much get the 1:16:5 performance.  I think this is the way to go.  The font mapping is not hard-coded anywhere, and just depends on a matching font.

Please review.  I have done my cleanup pass.

What remains is to remove the hard-coded code points for the greextra font, but I think I should do that on a separate pull request because this one has become pretty hefty.

eroux:
Well, apart from my small remark, for me it's ok, thank you very much for this excellent work! Greextra should be reworked a little, I'm opening an issue for this, it's quite separate.

henryso:
I originally hooked it there, but there were a couple of reasons I decided to change it:

1. If the selected font is not greciliae and I map it and the user decides to switch to greciliae later in the document, some mapping will be missing.  I would need more sophisticated logic to then remap previously unmapped glyphs in the new font, or at least to remap the entire font completely.
2. The character macros in `gregoriotex-chars.tex`, which are presently `\let` and were previously `\xdef` would have to become `\gdef`, which may incur some slight performance penalty.

Note: I seem to recall (but perhaps I was doing it wrong) that the lua font tables did not contain the glyph mapping because these are truetype fonts.  I think `fonts[n].characters[m].name` was `nil`.  Is there some other way to get the glyph names from the lua font tables?

I will try out these changes and see what I get.

eroux:
About the names in the lua structure, you can take a look a the `greciliae.lua` in the luaotfload cache to see, but I'm pretty sure `font.resources.unicodes` (or something like that) is what you're looking for. About the font not being greciliae, the good news is that gregorio always loads greciliae when loading `gregoriotex.tex`, so there should be no problem... This is a suboptimal behaviour, and it might have to be changed, but in the meantime, that can help! Also, due to this behaviour, you don't have to change `\let` for `\gdef` (if I understood the reason why you wanted to do this in the first place). The question is: do we want to keep this behaviour of loading greciliae by default, and then allow the user to change, or do we want to optimize this? I have no strong opinion on this....

henryso:
If I am reading it right, Gregorio does not load greciliae until `\begingregorioscore`, which is too late for `gregoriotex-chars.tex`, so I still think the switch to `\gdef` will be necessary.

I don't know what to look for to find `font.resources.unicodes`.  I looked through section 4.3 and chapter 7 of http://www.luatex.org/svn/trunk/manual/luatexref-t.pdf and can't find anything that resembles that.  Also, I have confirmed that `font.fonts[n].characters[m].name` is nil for all characters.

eroux:
Not everything is documented in the font structure, the final documentation will occur for LuaTeX 1.0 (in a few years), in the meantime, you can check the structure in the cache files of luaotfload, you'll see that there is a glyph name -> unicode character in `font.resources.unicode`.

Also, you're right, I was mistaken about the font loading... Maybe your way is ok for now, it doesn't seem to worsen the speed so we can keep it!

henryso:
I doubt it will change the speed of mapping much given the timings of using the pre-generated mapping.  I won't know the effect of switching to `\gdef` without trying it out.  It shouldn't take too much effort to try this out, so I'll see what I can do.

henryso:
Using the Lua font tables and switching to `\gdef` is slightly slower, 1:17.5.  The code is at https://github.com/henryso/gregorio/commit/49018a24b294314c71a7c058eabddd14a6f8f87a

I think I'll stick with what's here in this pull request.  It passes all salient tests (the gabc->gtex tests are expected to fail), and I'm happy with it in this state.

eroux:
Ok, let's do that, thanks!

tex/gregoriotex.lua
```diff
@@ -267,9 +267,23 @@ local function get_gregorioversion()
   return internalversion
 end
 
+local function map_fonts()
+  -- use greciliae because it is the most complete font
+  local font_info = fontloader.open(kpse.find_file('greciliae', 'truetype fonts'))
```
eroux: (sorry for thinking about that after merging) Here it might be useful to 

    log("Mapping font %s", kpse.find_file('greciliae', 'truetype fonts'))

because there might be edge cases where luaotfload doesn't use the font returned by kpse.find_file... luaotfload also looks in fontcache, etc. Might ease debugging...

eroux: In addition, it might be better to use

    fonts.names.font_file_lookup('greciliae.ttf')

(untested). This is defined [in luaotfload](https://github.com/lualatex/luaotfload/blob/master/src/luaotfload-database.lua).

eroux: Sorry for spamming, but it might be worth trying `fonts.names.resolve_cache()` instead of fontloader.open... just to see if performance improves a little... If we gain 1s per compilation (half a second per font), I think it's very helpful for small documents...

henryso: I have to understand the stuff in luaotfload better.  For instance:

`fonts.names.resolve_cache({name = 'greciliae'}) returns `'greciliae.ttf'`, `false` (from `luaotfload-lookup-cache.lua` in `TEXMFVAR`)

I'll see what I can figure out.

henryso: It would seem everything in the `fonts.names` namespace is really only to deal with resolving names of fonts to and from file names, rather than loading the cached font table.

henryso: Finding this discussion is difficult, so I'm going to continue this discussion in #279.

PR #293, opened 2015-04-19
Fix 270

### LINKED ISSUE #270: Fonts metadata

eroux:
* update font copyright
 * add version in font
 * check font version from TeX
 * small fixes and cleaning
 * fix #270 

OS2/unicoderanges can only be updated after #292, if necessary (not sure), I'll do a separate pull request for it.

### DIFF
--- a/VersionManager.py
+++ b/VersionManager.py
@@ -30,9 +30,12 @@ import re
 import argparse
 import subprocess
 import time
+import os
 from datetime import date
 
 from distutils.util import strtobool
+
+os.chdir(sys.path[0])
 
 VERSION_FILE = '.gregorio-version'
 GREGORIO_FILES = ["configure.ac",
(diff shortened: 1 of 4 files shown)


rpspringuel:
Why is it now obsolete to check that `gregoriotex.lua` and `gregoriotex.tex` are version coordinated?

eroux:
It's not. See #271 for the explanation of why the code is obsolete.

rpspringuel:
Ah!  So it's not so much that it's obsolete per se but rather that the code is redundant and thus obsolete.  Got it.

PR #294, opened 2015-04-20
More pylint cleanups.

eschwab:
More could be done to `gregoria2gregorio.py` but with the changes in #292 this script will be in need of updating.


henryso:
I think we should consider removing gregoria support. As it is, it won't work; there's stuff commented out in the C code.

eroux:
Indeed, it can be removed. I think I'll contact the gregoria creator one of these days, to see if a true integration is possible.

eroux:
@rpspringuel would that be ok to you if the copyright statement in `convertsfdtottf.py` is updated to have the same value as other files?

rpspringuel:
No problem on changing the copyright statement.

eschwab:
Should I remove `gregoria` here or in a separate PR?

eroux:
You can remove it here, I'll merge afterwards

eschwab:
I think I got all the references to `Gregoria`.

eroux:
Thanx a lot !

PR #297, opened 2015-04-21
Put the comment in the right place.

henryso:


henryso:
Just putting the comment in the right place.  Merging.

PR #300, opened 2015-04-22
Improving, cleaning and documenting Debian build system

eroux:
This is reasonable for 3.0, I think next step will be to have a separated `gregorio-debian` repository...

### DIFF
--- a/build-deb.sh
+++ b/build-deb.sh
@@ -12,9 +12,10 @@ until [ -z "$1" ]; do
     --clean      ) CLEAN=TRUE    ;;
     --lint       ) LINT=TRUE    ;;
     --git        ) GIT=TRUE    ;;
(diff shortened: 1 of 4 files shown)


henryso:
I am unable to test this, but eyeballing it, it looks fine.

eschwab:
Also unable to test, but looks fine. Perhaps it is possible to have one separate repository for all the build scripts (Debian, Arch, OSX, Windows...), i.e. `gregorio-build` or `gregorio-dist` or something like that.

eroux:
I think one repository by packaging is more usual... Something interesting would be to have something like Dockerfiles in the repositories, that would do all the job to compile arch packages, windows packages (can be entirely built from Linux), debian packages, etc. we can even imagine a build farm, jenkins, etc. but that's for the future...

PR #305, opened 2015-04-23
Added support for glyph variants

### LINKED ISSUE #301: Font variants
Gregorio currently supports in-font variants for punctum cavum and out-of-font variants for clefs and flats.  I (and some others, apparently) would like a different shape for the stropha in Greciliae.  We should probably implement some sort of variant system.

* This can be as simple as defining the additional glyphs and letting users redefine the ones they wish to override: `\let\grecpStropha=\grecpStrophaVariantB`
* Or perhaps the same with the assistance of a macro and some glyph naming system: `\greusealternative{Stropha}{VariantB}`
* Or something like what we have for `\greusehufnagelstyle`
* Or some other concoction of our imagination...

Thoughts?

henryso:
Fixes #301.

This is my first cut.

I also swapped in the strophicus forms sent by @eroux.

Still pending: documentation, change log, more testing.

I am also considering changing the _#### names in the base fonts into the actual glyph names since these numbers are basically meaningless at this point, but I think that's for a different pull request.

Whom should I credit for the new strophicus forms?


eroux:
You can credit *Abbazia Mater Ecclesiae (IT)* for the new strophicus.

eroux:
I didn't really have time to test, but the direction definitely seems the good one, perfect! About the `_####` names, you can do it in this one if you feel it, it already modifies many things in the `-base` font, let's consider it's also a cleaning PR... Anyway, it's very good that you do that now, because the new `-base.sfd` format will be a more robust base for gregoria4o!

henryso:
This one is ready for review.  It passes current output tests.  Performance testing to follow.

eroux:
Excellent! Waiting for your performance test, but ready to be merged for me

henryso:
I won't be able to get to the test until about 21:00 US/Eastern.

henryso:
I tested `develop` and `fix-301` and found that `fix-301` was slightly faster (about half a second) because it no longer maps the `greextra` font unless it's needed (which it's not for the 1000 of PopulusSion test), and since no variants are used, no variant fonts are loaded.

In inspecting the logs to confirm this, I found that the test was reloading the `gresym` font excessively.  I traced this to `\gre@reloadgresym`.  Apparently, on every *other* call to \gre@reloadgresym, `\current` was somehow set to 10 while `\f@size` was 10.95 even though a previous call had used `\let\current\f@size` when `\f@size` was 10.95.  I fixed this by using `xdef` instead of `let`.  This brought down the runtime by 5 seconds!

I liked what this code was doing (saving the font size to prevent unnecessary font loading), so I applied the same logic to the variant font loading.  It might be worth considering the same for `\gregoriofont`, but that is more complex since both the name and the size may change when `\setgregoriofont` is called when the score is included.  I have to think about that one, so I'll leave it for a different pull request.

Please review my changes.  I feel this is ready to be merged.

eroux:
Indeed, `\let` is just a pointer to another command, if the command changes, the pointer points to the changed value, while `\xdef` evaluates immediatly, so it cannot change... Anyway, that's very good news!

henryso:
For my own understanding: if `\let` strictly defines `\current` as a pointer to `\f@size` and `\f@size` changes, wouldn't `\ifx\current\f@size` always be true, and thus the font never be reloaded regardless of any change to `\f@size`?  What am I missing?

eroux:
Unfortunately I'm not expert enough in TeX to fully answer your question... From one point of view, yes, it should be always true, from another, it might also be always wrong, because `\ifx` might not be the right comparison tool... I'm not sure...

rpspringuel:
`\let` does not create a simple pointer.  The way functionallity is usually added to already defined commands in TeX demonstrates this:

    \let\oldsection\section
    \def\section{%
      \setcounter{\equation}{0}%
      \oldsection%
    }

The above code, for instance, would reset the `\equation` counter to 0 each time the section command was called.  Everything else that the section command does would still be done.  If `\let` created a simple pointer as you describe, this code would create an infinite loop.  Note, I don't have to specify arguments for section, because of the way TeX expands things and because I'm not using them in the additional code.  When `\section` gets replaced in the input stream by this code, `\oldsection` will be sitting there at the end, expecting the arguments and processing them as desired.

This is why I used `\let` over `\xdef` in the reloading code for gresym.  If that's not working, then `\f@size` must itself not simply store the font size, but has to contain code which evaluates to the font size.  What form this code takes, I'm not sure.  Clearly it must contain something that changes when the font size changes, otherwise the `\ifx\current\f@size` would never evaluate to false, and thus trigger the reloading of the font.  However, if `\ifx\current\f@size` is evaluating to false when it should be true, then the code behind `\f@size` must also change under some other condition.

PR #306, opened 2015-04-23
Add new bars, name gresym glyphs, remove precomposed A/, R/, V/

eroux:


eroux:
I think the best option is to merge this, finish the work on alternate glyphs, merge greextra and gresym and then provide macros for V/, R/ and A/? Or should I merge greextra and gresym in this PR? (I'm affraid it could interfere with @henryso 's work)

henryso:
If you are not working on the `*-base.sfd` fonts, it should not be difficult to merge our changes.  Alternately, merge my changes and continue your work.  On the surface, at least, my changes work, so I can do more testing and documentation afterwards.

eroux:
Ok, I'll do that tonight

eroux:
I did the merge, and also removed gregoriosyms.sty, which seems rather pointless... The "problem" I'm running through now is that it seems a bit stupid to provide default V/, R/, etc. for lmodern font (the default TeX font), so I'm thinking to add

    \usepakcage{libertine}

in the example... What do you think?

henryso:
Sounds like a good idea to me.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
 - Support for the font Gregoria.
 
 ## [Unreleased][ureleased]
+### Changed
+- `gregoriosyms.sty` is no more, its functionalities have been integrated in GregorioTeX.
+
```
henryso: This should be moved to the Unreleased section above.  This Unreleased section is for 3.0.0.

henryso:
Since the variant code from #305 scales the font according to the `\the\grefactor` multiplied by 100000 in `sp`, we probably need an alternate macro (or perhaps an argument) to pull them from the font since the glyphs imported from greesym should be scaled to `\f@size` in `pt`.

henryso:
I'll work on this and removing the hard-coded code points from in `gregoriotex-symbols.tex` and submit a pull request to `newbars`.

eroux:
Thanx, I'm still learning how to use your code properly, I'm glad you make a pull request, it will make it easier!

henryso:
I've removed the hard-coding but I didn't compose the barred letter glyphs.  I also don't know what to use for greletterbar and greletteraltbar, so I commented them out for now.

eroux:
Thanks a lot!

rpspringuel:
I don't think removing `gregoriotexsyms.sty` is a good idea.  It's functionality was always in GregorioTeX itself.  The purpose of the separate style file was simply to allow access to the special characters available in the `gresyms` font without having to load the whole GregorioTeX package.  It's use was for anciliary projects where you might not have a score, but wanted say the `\Rbar` or `\Vbar` so that you had a consistent look with a more elaborate project which did include scores.

eroux:
That makes sense...

henryso:
Because the code points here are no longer hard-coded, `gregoriosyms.sty` will need to load `gregoriotex.lua` (or some portion thereof).  If you think that is still a valid case to keep it, then let's resurrect `gregoriosyms.sty` and get it to work with the new system.

henryso:
These compositions look good to me for `\usepackage{libertine}`, but I don't really know what I'm doing with boxes, so I'm probably doing something bad with them:

###### gregoriotex-symbols.tex
```
\gredefsymbol{greAbar}{greextra}{ABar}%
\def\Abar{%
  A\rlap{\vbox{\moveleft.375em\hbox{\lower.1em\hbox{\greAbar}}}}%
}%

\gredefsymbol{greRbar}{greextra}{RBar}%
\def\Rbar{%
  R\rlap{\vbox{\moveleft.275em\hbox{\lower.15em\hbox{\greRbar}}}}\hbox to .1em{\hfill}%
}%

\gredefsymbol{greVbar}{greextra}{VBar}%
\def\Vbar{%
  V\rlap{\vbox{\moveleft.575em\hbox{\lower.1em\hbox{\greVbar}}}}%
}%

% a V with bar more adapted to small text
\gredefsymbol{greVbarsmall}{greextra}{VBarSmall}%
\def\Vbarsmall{%
  V\rlap{\vbox{\moveleft.575em\hbox{\lower.1em\hbox{\greVbarsmall}}}}%
}%
```

henryso:
(I edited the previous comment to delete the \hbox at the end of the `\Rbar` composition.)

eroux:
IIRC, you can call `\RequireLuaModule` several times, it should load the lua code only once, so you can call it in gregoriotex-symbols.tex. Also, you can define `\gregoriotex@symbols@loaded` at the end of `gregoriotex-symbols.tex`, and do something like

    \ifcsname\gregoriotex-symbols@loaded
      \endinput
    \fi

(the syntax of ifcsname might be totally wrong, this is just an indication) at the beginning of `gregoriotex-symbols.tex` to load it only if it has not already been loaded, this way including `gregoriotex.sty` and `gregoriosyms.sty` will be harmless.

I'm working on the bar macros, should be finished soon...

henryso:
I created #317 to restore `gregoriosyms.sty` and get it working again.

henryso:
The babel fixes are in #318.

eroux:
What do you think of the current state?

eroux:
Do not hesitate to criticize the wording in the doc, I really have doubts about many sentences...

henryso:
I think a PlainTeX test is a good idea, but I've also never attempted such a thing.  It'll take me a little time to figure it out, but if you have something that can kick-start me, I'd be happy to use it!

Regarding these changes, I'm having trouble using them.  I need some time to figure out what's gone wrong.

eroux:
A small file like

    \input luaotfload.sty
    \input gregoriotex.tex
    \input gregoriotex-symbols.tex
    
    \includescore[a]{PopulusSion}
    
    \bye

should be enough... There seems to be a package loading `graphicx.sty` which is not compatible with Plain TeX... if it's luaotfload, I'll report a bug.

eroux:
(updated previous comment)

henryso:
The old macros have a lower case `b` (i.e., `\Abar`) whereas the ones defined here have a capital `B` (i.e., `\ABar`).  Also there seems to be a problem with `\VBar` (capital `B`) not being defined.

eroux:
One thing I forgot when testing which made me have problems like yours was that I forgot to rebuild `greextra.ttf` from `greextra.sfd`... About the case, you're right, do you think

    \def\Abar\ABar \def\Vbar\VBar \def\Rbar\RBar

would be ok?

eroux:
PlainTeX compatibility problem reported in #319

henryso:
I have to use `\def\Abar{\ABar}\def\Rbar{\RBar}\def\Vbar{\VBar}` in my test to work around this.Not sure why we need to do this though, as `\gresimpledefbarglyph` can just pass `#1bar` instead of `#1Bar`.

That said, I still get `\VBar` as an undefined sequence.  I don't know why.  

If I actually put `\gresimpledefbarglyph{V}{.1em}` in my preamble, it no longer gives me the undefined sequence error, but the figure is not right.  It looks like V and then the bar to the right of V.

henryso:
Ok, I figured out the `VBar` problem.  It was a problem with my test.  So that said, besides the lower case `b` and upper case `B` thing, I think this looks fine.

henryso:
I think the simplest solution to the `B` thing would be
```
\def\gresimpledefbarglyph#1#2{%
  \greexpldefbarglyph{#1bar}{#1}{gre#1Bar}{\f@size}{#2}{0pt}%
}%
```

though the `\def\Abar{\ABar}\def\Rbar{\RBar}\def\Vbar{\VBar}` works as well, but leaves three additional macros.

eroux:
@henryso I've put the *B* in the macro names back in lower-case

henryso:
This pull request is getting pretty heavy, and although there are still some issues with the documentation, (which I will handle with a different pull request), all tests pass (given the changes), so I think this is ready to be merged.

PR #309, opened 2015-04-25
Upgrading functionality of debug

rpspringuel:
`\gre@debug` becomes `\gre@debugmsg` in code.
The message function can now take two arguments.  The first, optional, is a keyword, the second, required, is the message.  If the keyword appears in the list of keywords stored in `\gre@debug` then the message is printed to the log.  If no keyword is given (or keyword is `all`) then debug message will always be printed to the log when debug is active.
`\gre@debug` is created via the debug option by the `kvoptions` package.  Syntax is `debug` (print all debug messages) `debug=keyword` (print only those messages associated with keyword, or `debug={list,of,keywords}` (print those messages associated with any of the keywords in the list).
As demonstration, all `ifdim` debug messages are now associated with that keyword.

### DIFF
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1480,7 +1480,7 @@
       \grehskip\gre@skip@temp@four %
     \fi %
   \fi %
-  \gre@debug{Width of bar just printed: \the\wd\GreTempwidth}%
+  \gre@debugmsg{Width of bar just printed: \the\wd\GreTempwidth}%
   \global\gre@dimen@lastglyphwidth=\wd\GreTempwidth %
   \relax%
 }%
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1480,7 +1480,7 @@
       \grehskip\gre@skip@temp@four %
     \fi %
   \fi %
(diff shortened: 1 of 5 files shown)


henryso:
I actually think there should always be a keyword (it should not be optional) and that `all` should not be allowed as a keyword, but usable as a shortcut to specify `debug=all`.  General debug messages can have a keyword like `general` and you can decide if you want then by specifying `debug={general,whatever)`.  Is this possible?  Is this a bad idea?

eroux:
Thanks a lot, this is going to be really helpful! I agree with Henry, all messages should have a keyword

rpspringuel:
Actually, forcing all messages to have a keyword is simpler than what I've implemented.  If that's what people want, I can make that happen.

rpspringuel:
Oh, and it should be noted that care should be taking when choosing keywords.  If `testing` and `test` are both keywords, then `testing` messages will appear when the keyword `test` appears in the list.  Personally, I consider this a feature, rather than a bug, because it allows the creation of a "class" of keywords which all contain some other keyword.  However, it does mean that short, common letter combinations should be avoided when picking keywords.

eroux:
Seems mergeable to me, do you have more work planned for this PR?

rpspringuel:
Not unless someone doesn't like what I've done.

PR #320, opened 2015-04-26
Missing argument in shebang for fontforge

rpspringuel:
At least on a Mac following the website instructions, fontforge needs to be told that the script is written in python before it will process it.  Oddly, this only appears to apply to `squarize.py`, not `convertsfdtottf.py`.

### DIFF
--- a/fonts/squarize.py
+++ b/fonts/squarize.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env fontforge -script
+#!/usr/bin/env fontforge -script -lang=py
 # coding=utf-8
 # pylint: disable=too-many-branches, too-many-arguments, too-many-lines
 # pylint: disable=import-error, no-member
--- a/fonts/squarize.py
+++ b/fonts/squarize.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env fontforge -script -lang=py
+#!/usr/bin/env fontforge -lang=py -script
 # coding=utf-8
 # pylint: disable=too-many-branches, too-many-arguments, too-many-lines
 # pylint: disable=import-error, no-member


eschwab:
According to the fontforge manpage for linux:

    -script
        This must be the first argument passed to fontforge. Any other arguments are handled by
        the scriptfile itself. Any other command line arguments will be passed to the script.

I did try `fontforge -lang=py -script` and that worked fine. Fontforge does not seem to be too strict on `-script` being the sole argument.

rpspringuel:
I tried it the other way on my system and it worked fine too.  Already commited the swap.

henryso:
As an affirmation,  `fontforge --help` shows the following:
```
	-script scriptfile	 (executes scriptfile)
		must be the first option (or follow -lang).
		All others passed to scriptfile.
```
... so I think this is fine.  However, I think we should probably also do it to `convertsfdtottf.py` and the Makefile (even though they are currently working) as whatever algorithm fontforge is using to detect the script language may fail on us in the future.  I'll merge and make the other changes.

henryso:
Spoke too soon.  Shebang with /usr/bin/env has some issues with passing arguments on some systems.  I will see if I can come up with a workaround.

eschwab:
After looking at #321 I'm wondering if this would work:

    #!/usr/bin/fontforge -lang=py

Reason being that the shebang should only be read when the file is executed directly i.e. `./squarize.py gregorio` This would make the `-script` flag useless since that is used when running `fontforge` and giving it an external file `fontforge -script squarize.py gregorio` (as done by `fonts/Makefile.am`).

Would eliminate the need for a wrapper script.

henryso:
That would work, as long as fontforge is in `/usr/bin`.  If it's not, then you have a problem.  The problem is, how do you search on the PATH while passing the proper arguments?  If there were no arguments, then `/usr/bin/env` would work, but if there are arguments, then you run into the problem described on the Wikipedia page linked above.  The only workaround I could come up with was a secondary script, which is what I implemented in #321.

PR #321, opened 2015-04-26
Explictly specified python as the language of choice for fontforge scripts.

henryso:
#320 has a portability issue described on (Wikipedia)[https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability].  This works around the issue by introducing a different intermediate script which should pass arguments properly to fontforge.


henryso:
@rpspringuel Please try this out on MacOS X and let me know if there are problems.  This should work fine on pretty much all Unixes and Cygwin.

eroux:
Maybe it would be simpler to drop the shebang and remove the executable bit of `squarize.py`? I cannot see a use case where calling

    ./squarize.py greciliae

is more practical than

    make greciliae

? Also, it makes an additional file in the repository for a strange avoidable use case, it seems typically the kind of things anyone looking at the repository with fresh eyes in a few years would remove... Unless I'm missing something?

henryso:
I concur.  I always use `make fonts`.  I will close this pull request and open a new one with the suggested fix.

henryso:
For some reason, Makefile.am was also executable, so I will fix that in the other pull request.

PR #323, opened 2015-04-26
Changed GregorioRef date to %d-%b-%Y so date components are obvious.

henryso:
Using %d/%m/%y can leave the date components ambiguous depending on where you live, so I propose changing it to %d-%b-%Y (with `LC_TIME` set to `'C'`).  Feel free to close this request without merging if you disagree.

### DIFF
--- a/VersionManager.py
+++ b/VersionManager.py
@@ -31,9 +31,12 @@ import argparse
 import subprocess
 import time
 import os
+import locale
 from datetime import date
 
 from distutils.util import strtobool
+
+locale.setlocale(locale.LC_TIME, 'C')
 
 os.chdir(sys.path[0])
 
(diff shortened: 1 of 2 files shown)


eroux:
Why not something more human friendly like *April 26th, 2013* ?

eroux:
(sorry, it seems I merged the wrong PR...)

henryso:
It can do something like *April 26, 2013* pretty easily, but to put the proper ordinal *th* would require some extra work.  I don't see the logic in American date ordering, so might I suggest "26 April 2013" instead?

eroux:
I don't see the logic in American date ordering neither... Maybe it would be worth inquiring date formatting in LaTeX?

    \usepackage{datetime}
    \newdate{date}{26}{04}{2015}

then

    \displaydate{date}

(untested). But if it complicates things, I'm fine with *26 April 2015*.

henryso:
New pull request is #330.

PR #327, opened 2015-04-27
document removal of Vbarsmall, greletterbar and greletteraltbar

eroux:


henryso:
I just noticed that *4.0* is bigger than *3.0*.  Is that the style we are adopting, the current release larger than the prior releases?

eroux:
I don't really understand your question...?

henryso:
If you look at `UPGRADE.md`, 4.0 has `#` and 3.0 has `##`, so the output makes 4.0 look bigger than 3.0.

eroux:
Oh, sorry, my mistake, fixing this

henryso:
Looks good to me.

PR #333, opened 2015-04-27
Modified to properly record read/write operations on gabc/gtex files.

henryso:
This allows `latexmk -recorder` to detect that a change in an auto-compiled (or force-compiled) gabc file requires the document to be regenerated.


eroux:
Can you please give a more detailed explanation? If I understand correctly, this updates access time of gabc and gtex file... is `latexmk -recorder` based on file atime?

eroux:
Hmm, sorry I think I understand now: it opens the files, so that the recorded list of files contain them... did I get it right?

henryso:
No, the act of opening a file for read causes LuaTeX to record input and the act of opening a file for write causes LuaTeX to record output.

See http://tug.org/pipermail/luatex/2010-August/001942.html

It seems that this functionality was lost and restored in 0.78.2: https://www.tug.org/pipermail/luatex/2014-January/004726.html

henryso:
Yes, you have that right. (Sorry our messages crossed in transit)

eroux:
Thanks, I had never used the `-recorder` option. Maybe you could comment the code a bit more? This is particularly non-obvious...

henryso:
Done.  Is that good enough?

eroux:
Perfect, thanks!

eroux:
BTW, maybe it would be worth an entry in CHANGELOG?

Issue #265, opened 2015-04-13
GREGORIO_VERSION in gregoriotex.h


henryso:
We might want to move this to a separate header file (included from gregoriotex.h) to ease merge issues when the version number is bumped by VersionManager.py.

Thoughts?

eroux:
Well, now that the gabc/gregoriotex/gregorio versions are the same, `VERSION` from `config.h` should fit?

henryso:
Great.  I'll make the change in `develop` once the dust from rc2 settles a bit more.

henryso:
If I change it to use `VERSION`, then it won't work properly for non-tagged (off-master) builds.

PR #335, opened 2015-04-27
Switched non-tag version format to {version}-{branch}-{commit}-{count}.

### LINKED ISSUE #182: Version Number idea

henryso:
Fixes #182 and #265.


eroux:
That's ok for me

PR #336, opened 2015-04-27
Added a chant to the title page.

henryso:
On a somewhat whimsical note (puns not really intended), and based on something @eroux mentioned in passing, since we are already loading gregoriotex, I figure we might as well show off a little of what Gregorio can do on the title page.  I picked a chant I like which has some sort of musical connotation, and put it on the title page.  Let me know what you think and feel free to just close this request if you think it is a bad idea.

### DIFF
--- a/doc/GregorioRef.tex
+++ b/doc/GregorioRef.tex
@@ -37,6 +37,8 @@
 
 \usepackage{gregoriotex}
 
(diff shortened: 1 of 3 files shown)


eroux:
Very nice, thanx!

PR #337, opened 2015-04-27
Fixed faulty cases of determination as scandicus.

### LINKED ISSUE #20: enhanced parsing of liquescences
gabc `gwhi>` should be the same as `gwh!i>`

henryso:
Fixes #20.


henryso:
The scandicus is still determined for auctus ascendens and descendens figures.  However, these glyphs do not exist, so gregorio emits them as punctum-punctum-punctum auctus.  Currently, we only generate figures for scandicus and scandicus deminutus.  Initio debilis and auctus ascendens/descendens are not currently supported.

In its current state, this pull request causes `(-ghi)` to be equivalent to `(-gh/i)` in addition to preventing figures that don't start with a punctum from being interpreted as scandicus.  I'm not sure it makes sense to do the same for auctus figures, so I'm wondering if it's worth drawing the full set of liquescentia for scandicus.  I have no idea if those figures are valid chant shapes or how common/uncommon they are.

Does anyone have an opinion on this?

henryso:
Should I leave it at this or draw auctus figures for the scandicus in `squarize.py` ?

eroux:
Good question... these figures seem quite unlikely to appear... so I'd be in favor of not drawing them, as the private use area we're using is not endless

eroux:
(so you can merge, I just don't do it because it can't be done automatically)

henryso:
I'm currently in transit. I will get to it as soon as I can.

henryso:
Resolved conflicts.

PR #338, opened 2015-04-28
Modified to use the gtex extension by default for writing gtex files.

henryso:


eschwab:
Is fine for me. Letting the dust settle a bit on gregorio-devel before merging.

Issue #311, opened 2015-04-25
Glyph documentation


henryso:
As mentioned by @rpspringuel in #301 and seconded by @eroux, we need an index of available glyphs with variants on display.

eroux:
Also, when doing this list (in an annex of the doc file), it would be good to make another annex with all the glyphs in greextra.

henryso:
Please review #359 and let me know if I should change or add something.

PR #354, opened 2015-05-02
Renamed VirgaReversaDescendensLongqueue to VirgaReversaLongqueueDescendens.

henryso:
I'm working on glyph tables (#311), and renaming this will make sorting make more sense.


henryso:
Tests pass with this change. Merging.

PR #360, opened 2015-05-03
Remove unused glyphs for #357

### LINKED ISSUE #357: Remove unused glyphs?
In creating the font tables, it seems to me that `AuctumDuplex`, `Circumflexus`, and the `FlexusNobar` glyphs are unused.  Should they be removed?

There's also a special punctum used for measurement.  Why don't we just use a normal `Punctum` for measurement?

henryso:
Fixes #357

Additionally, I found an issue using the `Circumflexus` (now `VEpisemus.circumflexus`) because of the need to redefine macros created by the `\gredefvariant` macro in addition to using `\grechangeglyph`.  I eliminated `\gredefvariant` and now everything should be done strictly through `\grechangeglyph`.


henryso:
Tests pass after ab98646a6574e6c74b9f0c49a3607abec8f88a01.  Please review.

PR #361, opened 2015-05-04
Fixed problem rendering manual custos at the end of a score.

henryso:
This change also makes the distance between a bar and a manual custos in the middle of the line the same as at the end of a line.

Note: I also took the opportunity to change some of the sequences of if statements into an equivalent switch statement since I spent the time to understand the logic in the first place.  It might be a good idea to do this more generally as well, as it should theoretically allow the compiler to optimize the code better.


henryso:
Tests pass acceptably, given the changes.  I think this is ready for review.

PR #363, opened 2015-05-04
Converted if statement sequences to equivalent switch statements.

henryso:
I found a few more if statement sequences to convert into switch statements.


henryso:
Tests pass, so this should be good to go.  Please review.

PR #364, opened 2015-05-04
Revert 4.0.0-rc1 to 3.0.0-rc2

eschwab:


eroux:
If we really want that, we should document it in the wiki (before it's merged). Also, if there's a beta, maybe there should be an alpha? alpha could be when the features that will be present in next version are not decided (current state), beta is when a new branch is created, release-4.0 in our case, and rc1 is when there is a rc1 released...

What do you think?

rpspringuel:
I would say that the "alpha" version is whatever is currently on develop (and other feature branches).  I would not give it a version number like `4.0.0-alpha`, but would stick with the system we just put in place where branch, revision, and commit are identified.  "Beta" would be the first version number after the creation of a release branch and said release branch would then progress through rc1, etc. until the release is made official and merged into master.

rpspringuel:
Oh, and that means that develop's version number should be reverted to 3.0.0-rc2 (to be changed to 3.0.0 when that is official) for now.

eroux:
I tend to disagree on your last comment, though it would make it easier for merging

rpspringuel:
The idea, for me at least, is that the version number on develop (and all feature branches) should be the version number of the current stable release.  In this manner #335 results in the following:

{most recent stable version on which a build is based}-{branch on which a build is based}-{commit on which a build is based}-{number of revisions between this commit and the stable version along this branch}

Once we get 3.0.0 out the door, the version number on develop should never again have beta, rc1 etc.  The only reason it does right now, is because of how screwed up the yanked 2.4.3 made our master branch.  In the future, then, all changing of the version number would happen on hotfix and release branches and would come into develop only via a merge.

eroux:
Why not... if you all agree on that, I let someone document it on the wiki page.

eschwab:
I tend to agree with @rpspringuel on this. When 3.0.0 is released, `release-3.0` gets changed to `3.0.0`, that is merged into `master` and `develop` leaving `develop` at `3.0.0`. Changing the version in `develop` makes conflicts for merging.

`-alpha` `-beta` `-RC?` tags should only be set on hotfix and `release-?.?` branches. I can edit the wiki.

henryso:
In that case, this should be (as @rpspringuel pointed out) 3.0.0-rc2 rather than 4.0.0-beta.

eschwab:
Version changed here and wiki updated.

eroux:
ok

PR #365, opened 2015-05-04
Clarifying Post Install Check Box for Windows Installer

rpspringuel:
Made explicit in the option the fact that the installer can add files only to the TeXLive texmf tree.


henryso:
Seems fine.

PR #366, opened 2015-05-04
Many improvements and fixes in fonts

### LINKED ISSUE #328: make squarize.py more generic (handling possibly more fonts)
This request is with gregoria4o in mind. squarize.py should be able to handle more fonts (with the same format) than the three in the Gregorio repository. To do so, an important step is to get rid of the hardcoded widths in `get_lengths`. I think they come from the old convoluted system where python was generating a fontforge script, but it would be fairly easy to get rid of them now, by having get_lengths read the lengths in the base font directly.

Also, it would be best if the table returned by `get_lengths` was having the width of all the glyphs in the base font.

Finally, Gregoria has different widths for base4, base2, base5, base3, etc. so squarize.py shouldn't consider them all to have the same width.

eroux:
- fix #322 partially (drawings to be done later)
- fix #328
- fix `-ghi~`
- fix horizontal episemus length
- add more glyphs in the final font, for measurements (not used on TeX side yet), with unperfect names
- add `-o` argument to squarize
- fix horizontal episemus edges
- fix gregoriotex-main missing from VersionManager


henryso:
Spacing is really messed up.  It "smells" like a missing % somewhere.

henryso:
By way of example, page one of https://github.com/gregorio-project/gregorio-test/blob/master/tests/gabc-output/glyphs/pes.gabc :
page-0

henryso:
It would appear that the width of PesOneNothing in the generated `.ttf` file is wrong.

eroux:
Oh, ok, I'll fix that tomorrow (I wasn't available tonight)

eroux:
fixed

eroux:
BTW, it would be good to have tests for `-gig~`, as the glyphs were not drawn correctly before (I didn't recheck, but I think it was broken before), as well as horizontal episemus coming on the entire porrectus, I think it didn't have the correct width before...

henryso:
We might have tests for these things, only I didn't know what is correct and what is not -- the test for bars with all the vertical episema (from the original all-glyph test), for example -- I have no idea what is correct or not and take it all as good until someone tells me otherwise.  With this latest change, there are still some differences, spot checked as mostly minor tweaks in spacing.  I will review these in detail tonight.

henryso:
Some differences I've noticed.

- In the pes test, the pes initio debilis figures are furtther apart and the ictuses on the pes initio debilis figures are offset slightly farther off center to the right.

before:
image
after:
image

before:
image
after:
image

- One-ambitus clivis figures have their notes slightly tighter together.  I suppose this is an aesthetic choice, but the `gof~` figure has the deminutus too far to the left and the connection is very abrupt (looks like a stair-step).

image

I will add a tests for `-gig~` and `hgh___` through `hch___` (some of these already exist in some form)..

eroux:
Thanx! it should be fixed now

henryso:
Changes look good.  Merging.

henryso:
I'm not sure if you want to keep the branch for more development, so I will leave the branch open.  Please delete it if you don't need it.

PR #367, opened 2015-05-05
add quickcompile macro (by @jperon, from jperon/greluatex)

eroux:
This is a very small macro, but it could be useful... What do you think? I'll document it once I have everyone's opinion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Added
 - The ability to substitute an arbitrary glyph for one used by GregorioTeX.  This adds four macros: `\grechangeglyph` to make a score glyph substitution, `\greresetglyph` to remove a score glyph substitution, `\gredefsymbol` for (re-)defining an arbitrary non-score glyph that scales with the text, and `\gredefsizedsymbol` for (re-)defining an arbitary non-score glyph that requires a point-size to be specified.  See GregorioRef.pdf for full details.
 - Support for `lualatex -recorder`.  Autocompiled gabc and gtex files will now be properly recorded so that programs like `latexmk -recorder` can detect the need to rebuild the PDF when a gabc file changes.
+- With thanks to Fr. Jacques Peron, it is now possible to embed short gabc snippets directly into a TeX document.  The command is `\gabcsnippet`.  See GregorioRef.pdf for full details.
 
 ### Removed
 - GregorioXML and OpusTeX output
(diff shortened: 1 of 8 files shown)


henryso:
I suppose this is for inline gabc?  If so, I like it. I've been thinking it would be a good feature, but I couldn't think of how to do it.

I have two comments.  First, I don't think `quickcompile` is a good name.  Perhaps `inlinegabc` or `directgabc`?

Second, there needs to be a way to specify `initial-style` because the logic changes depending on that setting.  Other stuff that affects the score, such as `mode` or (in the future?) `annotation` might be nice also, but these can be handled in TeX itself so it's not as necessary.  Maybe it's possible to have an optional argument for `initial-style` or headers in general?  If done in general (as opposed to just for `initial-style`), catcodes would need to be controlled to handle the newlines.

I can attempt these changes if you don't have time.

eroux:
Indeed, the current code is quite imperfect... I agree that initial-style should be settable simply, but this code is more intended for small gabc snippets than for real scores, for which the separate file option should still be preferred... I admit I didn't test newlines (only thinking about small snippets), but there shouldn't be an error if there is a newline, you're right...

I was also thinking about two other features:

- possibility to include complete gabc scores (with headers), with a LaTeX environment (could be inspired from [luacode](https://github.com/lualatex/luacode))
- possibility to translate one gabc glyph into its TeX equivalent. That would be practical for tex/neumes mixing, but that would require some work on the gregorio side...

eroux:
Also, I don't really plan further developments on this, so fell free to take the lead on this PR

henryso:
Newlines should not be a problem in the gabc part.  I was just saying that if the optional argument took in all headers (as opposed to just `initial-style`), newlines there are significant, so catcodes would need to be handled.

So the big decision here, which affects future flexibility/backwards compatibility is:
- have no optional argument: this is for snippets only, with a leading initial
- have an optional argument with the `initial-style` value: this presumes that any other customization will be done in TeX and that there will probably not be other header-controlled features in gregorio that cannot be handled in TeX
- have an optional argument which takes all headers: This is the most flexible, but requires users to type out the full headers in the optional argument.

Which of these resonate with you the most?

eroux:
I hesitate between second and third solution... but maybe for small snippets the second would be enough, people with higher standards will have to use the complete environment or a separate file... What do you think?

henryso:
I am undecided between the two, so I'll go with your choice.  About naming, how does `\gabcsnippet` sound?  It sort-of expresses that this is not for "full" gabc compilation.

eroux:
Yes, I think it's an appropriate name

jperon:
I'd personally prefer the second solution : though inline gabc is a good
thing for small parts, I think it wouldn't be a good practice to do that
for big songs. The third solution would encourage such a practice, and
would be overkill for small extracts.

2015-05-05 13:06 GMT+02:00 Elie Roux <notifications@github.com>:

henryso:
I made the changes as discussed here.  Please review.

eroux:
Well, for me it's ok... I think we can merge?

henryso:
Ok, let's merge!

PR #370, opened 2015-05-05
Fix wrong stream type for Popen.

eschwab:


eroux:
Can you be a bit more specific about the bug that it fixes?

eschwab:
Running `fontforge -script squarize.py gregorio` results in:

    Traceback (most recent call last):
      File "squarize.py", line 1622, in <module>
        main()
      File "squarize.py", line 108, in main
        version = proc.stdout.read().strip('\n')
    TypeError: 'str' does not support the buffer interface

This commit fixes it. Setting `universal_newlines=True` makes the stream a text stream rather than a binary stream.

eroux:
ok thanx!

PR #371, opened 2015-05-05
Implements <eu> block

### LINKED ISSUE #9: Euouae block features
Euouae blocks have several features that can be different from other parts of the score. Among which:

 * no line breaking
 * different spacing

There should be a `<euouae>` (or `<eu>`?) markup in gabc so that these can be automated without ugly TeX inside gabc. Or maybe gregorio should detect it automatically? Would be feasible.

Gregorio should also not break line before the last syllable before the Euouae, and keep the main score and the euouae block on the same page.

eroux:
Fix #9 (that was about time!)

The only problem is that it only recognizes

    <eu>E(i) u(i) o(i) u(h) a(h) e</eu>(fe..) (::)

and not

    <eu>E(i) u(i) o(i) u(h) a(h) e(fe..)</eu> (::)

but that should be fixable... If someone has time, please fell free to do it, otherwise I'll finish tomorrow.


henryso:
I started to gave it a try, but I think it'll require a rewrite of major portions of the parser grammar since the lexer can no longer be depended upon to find `CLOSING_BRACKET_WITH_SPACE`.  And then, the post-bracket content (`</eu>` and `</nlba>` in any combination, or not there at all) will conflict with the regular text rules.  Not that it can't be done, but I would need a solid block of time devoted to this change to work it all out -- more than I typically have on work days.  If it's still pending by the weekend, then I'll give it another try.

eroux:
Well, let's take the current syntax as the expected syntax then...

PR #373, opened 2015-05-06
Have \gabcsnippet work in restricted \write18 mode

jperon:
In shell-restricted mode, even when `echo` and `gregorio` are allowed, direct use of
gregorio with `io.popen` doesn't work ; this change makes use of a temporary file in that case.

There is a problem though : I don't know how to remove the temporary file after passing `'\\input{|"gregorio -S '..tmpname..'"}'` to the tex document ; if I put `os.remove(tmpname)` just after `tex.print` (line 293), it seems it's removed "too quick".

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -281,24 +281,18 @@ local function include_score(input_file, force_gabccompile)
 end
 
 local function direct_gabc(gabc, header)
-  local gabccode = 'name:direct-gabc;\n'..(header or '')..'\n%%%%\n'..gabc
-  local f = io.popen('echo "'..gabccode..'" | gregorio -sS', 'r')
-  -- In shell-restricted mode, with gregorio (and even echo) allowed,
-  -- the previous command doesn't work ; in that case, use a temporary file.
-  if f == nil then
-    tmpname = os.tmpname()
-    local p = io.popen('gregorio -s -o '..tmpname, 'w')
-    if p == nil then
-      err("\nSomething went wrong when executing\n    gregorio -s -o "..tmpname..".\n"
-      .."shell-escape mode may not be activated. Try\n\n%s --shell-escape %s.tex\n\nSee the documentation of gregorio or your TeX\ndistribution to automatize it.", tex.formatname, tex.jobname)
-    end
-    p:write(gabccode)
-    p:close()
-    f = io.open(tmpname)
+  tmpname = os.tmpname()
+  local p = io.popen('gregorio -s -o '..tmpname, 'w')
+  if p == nil then
+    err("\nSomething went wrong when executing\n    gregorio -s -o "..tmpname..".\n"
+    .."shell-escape mode may not be activated. Try\n\n%s --shell-escape %s.tex\n\nSee the documentation of gregorio or your TeX\ndistribution to automatize it.", tex.formatname, tex.jobname)
   end
+  p:write('name:direct-gabc;\n'..(header or '')..'\n%%%%\n'..gabc)
+  p:close()
+  f = io.open(tmpname)
   tex.print(f:read("*a"):explode('\n'))
   f:close()
-  if tmpname then os.remove(tmpname) end
+  os.remove(tmpname)
 end
 
 local function check_font_version()
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -288,6 +288,10 @@ local function direct_gabc(gabc, header)
   if f == nil then
     tmpname = os.tmpname()
     local p = io.popen('gregorio -s -o '..tmpname, 'w')
+    if p == nil then
+      err("\nSomething went wrong when executing\n    'gregorio -sS'.\n"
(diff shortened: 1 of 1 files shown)


eroux:
Here what you should do is to make gregorio output the tex on stdout and read it, like in the first solution, I don't think your current solution will work well... Do you see what I mean?

jperon:
The problem is that, to read gregorio output in lua, I need to use `io.popen`, that doesn't work at all in restricted mode. On tests I've made, that solution works.

henryso:
I think you're going to have to remove the -S option, have gregorio write a file (maybe using -o so the filename is controlled), and then read that file, writing it into TeX.  After that, it should be safe to delete both files.

jperon:
I've thought about this solution, but : 

- what about performance (writing two files instead of one) ? For a small doc, this wouldn't matter, but for a bigger one, this would ;
- I don't understand the problem with `\input{|"gregorio..."}` : is this a compatibility issue, or something like that ?

eroux:
I don't think you'll have a performance difference between small and big files (unless you have scores of several MB, which would be odd), but yes, using the filesystem slows things down... Also, I admit I don't really understand what `\input{|"gregorio..."}` does... is this a special input syntax? Anyway, I think the solution Henry proposes fixes the problem of the file that was difficult to remove. But there could also be a list of files in lua, and at the end of the document, lua could be called to remove the files, this would make less access to the filesystem...

henryso:
What the script does when the popen fails is:

1. Write the contents passed to a temporary file
2. write \input{"|gregorio -S {temporary file}"} into TeX

When this returns to TeX, the \input command instructs TeX to run the command and place the output into the document.  Therefore, until TeX reads the temporary file, you cannot delete it.

One thing I would try is using io.popen("gregorio -s -o "..tmpname, 'w') and writing the gabc to the pipe.  Then, you can open that file and read it, then delete it.  This would only create one temporary file.

According to the documentation, restricted write18 mode *should* allow gregorio to run, so my hypothesis is that echo (which may be a shell builtin) was preventing it from working.

eroux:
indeed, I think that would be the best solution... after the io.popen, do something like (untested)

    tex.sprint('\includescore[n]{'..tmpfile..'}\greremovetmpfile{'..tmpfile..'}')

and define \removetmpfile to a command removing the temporary file (in lua).

jperon:
I've just updated the code thanks to @henryso's advice.

eroux:
Ok, thanx, now there should also be an error in case gregorio is not in the authorized commands (the same error as before)

jperon:
Oops… Excuse me, as it wasn't necessary with `\input…`, I forgot to put it back. It's now done !

jperon:
Sorry… Just corrected.

tex/gregoriotex.lua
```diff
@@ -281,13 +281,24 @@ local function include_score(input_file, force_gabccompile)
...
   if f == nil then
-    err("\nSomething went wrong when executing\n    'gregorio -sS'.\n"
-    .."shell-escape mode may not be activated. Try\n\n%s --shell-escape %s.tex\n\nSee the documentation of gregorio or your TeX\ndistribution to automatize it.", tex.formatname, tex.jobname)
+    tmpname = os.tmpname()
+    local p = io.popen('gregorio -s -o '..tmpname, 'w')
+    if p == nil then
+      err("\nSomething went wrong when executing\n    gregorio -s -o "..tmpname".\n"
+      .."shell-escape mode may not be activated. Try\n\n%s --shell-escape %s.tex\n\nSee the documentation of gregorio or your TeX\ndistribution to automatize it.", tex.formatname, tex.jobname)
+    end
+    p:write('name:tmp;\n%%%%\n'..entree)
```
henryso: this should be `p:write(gabccode)`

eroux:
Out of curiosity: does the direct version (`gregorio -sS`) work under Windows?

jperon:
Sorry again : how distracted I am !

henryso:
@eroux I tend to doubt it will.  If that command uses `cmd`'s echo then it won't handle a multiline snippet very well.  I think it's probably best to just use the newer method (write to the pipe and read it, followed by deleting the temp file).  That should work in all environments and there's no need to detect restricted write18 mode by trying one.

jperon:
@henryso Yes, I corrected that and another typo (lack of two dots…).

eroux:
@henryso As this method might be also used by a potentiel environment where performance might matter, do you think it's worth detecting windows (there is an os.name in LuaTeX) and doing the second method?

henryso:
If performance is absolutely importance, then you wouldn't want to use the snippet feature, which compiles every snippet every time, no?

eroux:
Right. @jperon could you use only the second method (the one with the temporary file) so that it works under Windows please?

jperon:
Here you are. This time, I really tested it before committing…

eroux:
Thanx! That's a good habit to take :)

@henryso if it's ok for you, you can merge

henryso:
Looks ok to me.  Merging.

PR #379, opened 2015-05-07
Implementation of fused torculus resupinus flexus.

### LINKED ISSUE #18: Missing torculus resupinus flexus
Here are some samples from the dominican Graduale: in these, the first interval is 1 or 2. In the last one, the first interval is 1 but the last note is a liquescence.
dominican neumes 2

dominican neumes 3

dominican neumes 4

henryso:
Fixes #18.

This is ready for review.  It passes all tests except for the ones expected to fail because of the new shape and some very minor changes at the join points of episema.


eroux:
Thank you very much, that seems perfect!

PR #381, opened 2015-05-07
Unrevert "Implementation of fused torculus resupinus flexus."

henryso:


henryso:
I committed to the branch without review.  We can work out the kinks in the branch.

Issue #70, opened 2015-03-09
New horizontal episemus on porrectus except last note


eroux:
Page 125 of the Graduale Romanum (1961) lies a strange horizontal episemus currently impossible to reproduce with Gregorio, not even to represent in gabc:

![horizontal_episemus_porrectus](https://cloud.githubusercontent.com/assets/60868/6555219/2bee363a-c664-11e4-83ff-27d9ee2c6b5d.png)

I don't even know how to represent this oddity... Any suggestion?

henryso:
What does that even mean? Are there episema on the first two notes or the first three?  Assuming this means the first three, I would have two suggestions.

My first suggestion is, since the _'s count from the the current note backwards, maybe have some other character as a "skip" character (I use = here for demonstration):  fhfg=___

My second suggestion is to not have the _ break up the neume: fhf___g or f_h_f_g

Is this chant in the Triplex?  Maybe we can look at the nondiastematic neumes and see if maybe this figure is just a typo.

eroux:
This comment is not meant to reopen the issue or anything, it's just for another reference: page 98 of Graduale Triplex lies something similar, but the lines goes up to the middle of the final note, which makes it look like a printing error... maybe the original was too...

henryso:
Well, the Laon is medium uncinus, large uncinus (with "a"), small uncinus, virga, so I'm inclined to believe this really is an episema on the first three notes but not the fourth.

henryso:
Of course the St. Gall just shows a regular torculus here,  but that could also be a rhythmic variance between monasteries.

Issue #40, opened 2015-02-08
short horizontal episemus


eroux:
In order for many horizontal episemus to render correctly, it would be useful to introduce a new feature (that will actually be used in some new books): horizontal episemus with 2/3 width. For instance, `gd_` would use this (not automatically).

These horizontal episemus would be left-aligned, right-aligned or centered, according to the wish of the user. They could be noted by 

 * `_1` for left-aligned
 * `_2` for right-aligned
 * `_3` for centered

and they would have a width corresponding to 2/3 of the width they have now, which is proportional to the note they are above (or under).

These short episemus can also be under the notes, in which case they are noted `_01`, `_02` and `_03`.

The episemus are combined like other episemus in certain cases: left-aligned short episemus, it combines with the episemus on the left, and right-aligned short episemus combine with episemus on the right.

olivierberten:
In some exotic cases, some fine-tuning would be useful:
* episemae on repeated notes: http://gregobase.selapa.net/chant.php?id=2282
* episemae not extending to the last note of a porrectus: http://gregobase.selapa.net/chant.php?id=1329

This proposal would fix the first issue but I don't know how to fix the latter one...

eroux:
@olivierberten I admit I don't understand at all :) Can you give the line number (in the printed score on gregobase) and the gabc code of the problems?

olivierberten:
For the first example, it is on the 3rd line of LU and 4th of gregobase, on the word "calceamenta". The original scores has 2 distinct episemae, while gregorio puts only a big one.

The second one is on the 6th line of the 2nd page of GR, and 8th line of gregobase, on the word "Ita". The episemus stops before the last note of the porrectus, which is afaik impossible to achieve with gabc...

eroux:
Oh, thanx! Did you update the scores since I reduced horizontal episemus size (a few weeks ago)? It might have dealt with the first prolem... About the second, it's interesting, but quite unrelated to this issue (though it might be solved at the same time. I've opened a [separate issue](https://github.com/gregorio-project/gregorio/issues/70) for it.

henryso:
How would this work for larger neumes like porrectus or torculus resupius flexus?

eroux:
I guess so... the feature is really not about them, but I guess just mimicking the others will be ok

eroux:
[...]

PR #384, opened 2015-05-08
Fix for #70 and some porrectus episemus bugfixes

henryso:
These are two commits primarily to address #70.

While testing, I noticed some weirdness with the heights of horizontal episemus decorating porrectus and torculus resupinus figures.  See the following tests in gregorio-tests to see the difference:

- gabc-output/glyphs/torculus_resupinus_quilisma.gabc
- gabc-output/glyphs/porrectus.gabc
- gabc-output/glyphs/torculus_resupinus.gabc
- gabc-output/glyphs/torculus_resupinus_flexus.gabc

I made some semi-educated decisions around positioning the episemus on the final two notes of a porrectus or torculus resupinus, which may be wrong, so please review these changes carefully before accepting them, and please let me know if I should change anything.

A few notes:
- The widths of the porrectus and ambitus-one porrectus episemus glyphs are the same as they were, but the figure within the glyph is reduced horizontally by the size of `PunctumSmall` and left-aligned.  This is the essential fix for #70.  I did this so that thee calculations for what modern notation calls ledger lines will be correct.
- In the same way that torculus resupinus flexus falls through into porrectus flexus (in the C switch statement), torculus resupinus now falls through into porrectus.  Thus the tail handling (of the "porrectus part") should be identical.
- In the "porrectus part" of high-low-high, if the episemus is on the low note, it will always be below and right-aligned with the vertical stem unless there is also an episemus on the first high note.  Auctus figures (with the head to the right of the stem) should now be taken into account correctly.
- If the all three notes of a "porrectus part" have episema and the final high note is a deminutus, it will draw a punctum-sized episemus to ensure that the episema melt together properly.
- There are still some anomlies like `(d_h_g_j)`: the height of the episemus above `hg` is in a different place than the episemus above `d`.  In the older code, the episema would be at same height, but actually end up below the `i` note.  I will look into this on another day unless someone beats me to it.

### DIFF
--- a/fonts/squarize.py
+++ b/fonts/squarize.py
@@ -503,26 +503,34 @@ def hepisemus(widths):
     write_hepisemus(widths, get_width(widths, 'PunctumLineTR'), 'HEpisemusPunctumLineTR')
     write_hepisemus(widths, get_width(widths, 'PunctumLineBLBR'), 'HEpisemusPunctumLineBLBR')
     write_hepisemus(widths, get_width(widths, 'PunctumAuctusLineBL'), 'HEpisemusPunctumAuctusLineBL')
+    reduction = 0.75 * get_width(widths, 'Punctum')
     for i in range(1, MAX_INTERVAL+1):
-        write_hepisemus(widths, get_width(widths, "porrectus%d"%i), 'HEpisemusPorrectus%s' % AMBITUS[i])
+        write_hepisemus(widths, get_width(widths, "porrectus%d"%i),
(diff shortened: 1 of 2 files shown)


henryso:
Might as well get the ASCII chart in as well.

henryso:
If you are satisfied with the changes thus far, then please merge this, I will tackle the remaining anomalies as part of #40 / #385 / #386 since the episemus code needs to be rewritten to handle those issues anyway.

PR #394, opened 2015-05-09
Initial St. Gall ancient neume support.

jakubjelinek:
For more details see https://github.com/jakubjelinek/nabctrial and https://github.com/jakubjelinek/Editio-Sti-Wolfgangi , will try to cook up documentation in *.tex form including the nabctrial neumetable table soon.
Autoloading is not working, thus one has to use
\input gregall.tex%
explicitly.


henryso:
This doesn't appear to break anything, so I'm going to merge it.

eroux:
@jakubjelinek Thank you very much for your work, this is really incredible! Just a question: you didn't include the images from e-codices in the font, so I guess you don't want to consider them as quotes falling in the quote exception to copyright? Where can we find the font with the images?

jakubjelinek:
Yeah, as I haven't asked Stiftsbibliothek for permissions, I think it is better to try to play safe.
gregall.sfd with the background images is in my https://github.com/jakubjelinek/nabctrial repository.

As for automatic loading of gregall.tex, can a TeX macro \input an *.tex file so that it is available outside of the scope of that macro (and \input it conditionally, only if it hasn't been input yet)?

eroux:
Maybe we could ask the permission? I can handle it if you want...

About the macro, it should be possible indeed... Loading it conditionally is straightforward, I admit I've never thought about scope problems, but it's a good question... An easy way to fix it is certainly to declare all the macros global in your TeX file... Or a cleaner approach: load all your macros anyway (always input `gregall.tex`) but add a `grenabcinit` (or whatever) macro that loads the lua file conditionally, that's probably the best solution. But I can also take care of it if you want (after I finish Gregoria integration).

jakubjelinek:
Yeah, help with that would be appreciated, will try to concentrate on documentation and eventually on the gresgmodern font.  If you could ask them for permissions, that would be appreciated too (though, supposedly that would need to be at least two permissions, one from St. Gallen Stiftsbibliothek and one from Einsiedeln (there is one image from Bamberg codex, but that is probably not enough to care, 51 from Einsiedeln, 347 from St. Gallen, I can supply detailed list on what came from where).

eroux:
Ok, I'll do both tomorrow

jakubjelinek:
FYI:
https://www.dropbox.com/s/206bzhtvukqkqsn/GregorioNabcRef.tex
https://www.dropbox.com/s/qcxz7vb9ek9btwf/GregorioNabcRef.pdf
is my current version of the documentation, spend most of time reproducing Cardine's table with nabc.
Have also updated http://gregoriochant.org/dokuwiki/doku.php/language and plan to duplicate it in the .tex file, perhaps start with some sample score with nabc and then describe the syntax.
Haven't been successful in typesetting of the included GregorioRef.tex BTW, the error I get is
! Unable to parse FlexusAmOneLineBL.
l.46 ...e_glyphs('greciliae','gregorio','parmesan')}
But presumably this document should be included as a separate document.

eroux:
This is really impressive, thanks a lot!

I've got the error too... @henryso it's in the glyph table appendix, could you take a look?

eroux:
Edit: well, the error is quite obvious, I just never looked at the code... @jakubjelinek, you can apply

    diff --git a/doc/GregorioRef.lua b/doc/GregorioRef.lua
    index 80b2949..760eb90 100644
    --- a/doc/GregorioRef.lua
    +++ b/doc/GregorioRef.lua
    @@ -41,6 +41,8 @@ local EXCLUDE = {
       HEpisemusPunctumLineBR = true,
       HEpisemusPunctumLineTL = true,
       HEpisemusPunctumLineTR = true,
    +  FlexusLineBL = true,
    +  FlexusAmOneLineBL = true,
     }
         
     local GABC = {

and it will work, I'll fix that tomorrow (if it's not done before)

eroux:
@jakubjelinek , I'm working on the nabc branch of the main repo, you can do pull requests against it

eroux:
@jakubjelinek I've sent the mails, I'll tell you when I'll get answers (if I get any). Also, what about harmonizing the size of the two fonts? From what I understand of the TeX part, gregall is twice the size of gresgmodern. What about scaling all glyphs of sgmodern by 2? I'll do it if it's ok for you.

jakubjelinek:
Thanks.  Scaling gresgmodern by 2 is ok with me too.

eroux:
done in the nabc branch

PR #397, opened 2015-05-10
Increase default intersyllablespacenotes

jperon:
With 0.17 cm, spaces inter syllables are sometimes too small. Here, 0.24 cm is suggested.


eroux:
I don't have any opinion on this, of noone disagrees, I'll merge (tomorrow)

henryso:
Is the number 0.17 arbitary? Is 0.24 arbitrary?

rpspringuel:
The distance is fairly arbitrary, but I think the idea is for the default to mimic the style found in a standard liturgical text.  I'm not sure if there is a specific text we're trying to match or just an overall look.

eroux:
It's just some tries so that it looks nice, nothing very studied...

henryso:
Since both are arbitrary, and you can adjust the spacing as you like anyway, I have no objection.

PR #398, opened 2015-05-10
Diagnostictools

rpspringuel:
Note: I do not use the `where` function in order to maintain compatibility with XP (which doesn't have it).
I've tried to make the output somewhat pretty and easy to read since I expect us to be reading these ourselves, rather than processing them with code.

Tested on XP and 7.  Windows 7 requires that this be run as an Administrator, but does otherwise work.  @BGMcoder, could you test on Windows 8 (I still don't have that testing environment set up?

Is there anything else that I might have forgotten?


rpspringuel:
Oh, for those interested, and example output (in this case for Windows 7 with MiKTeX installed, but not gregorio):

    	Gregorio Windows Setup Results 
    	Created: Sun 05/10/2015 
    ----------------------------------------------------------------------------- 
     
     
    	Windows Version 
    
    Microsoft Windows [Version 6.1.7601]
    ----------------------------------------------------------------------------- 
     
     
    	LuaTeX Version 
     
    This is LuaTeX, Version beta-0.76.0-2013062821  (MiKTeX 2.9 64-bit) (rev 4627)
    
    Execute  'luatex --credits'  for credits and version details.
    
    There is NO warranty. Redistribution of this software is covered by
    the terms of the GNU General Public License, version 2 or (at your option)
    any later version. For more information about these matters, see the file
    named COPYING and the LuaTeX source.
    
    Copyright 2013 Taco Hoekwater, the LuaTeX Team.
    
    	LuaTeX Location 
     
    C:\Program Files\MiKTeX 2.9\miktex\bin\x64\luatex.exe 
    ----------------------------------------------------------------------------- 
     
     
    	TEXMFLOCAL 
     
    C:/Users/RPS/AppData/Roaming/MiKTeX/2.9
    ----------------------------------------------------------------------------- 
     
     
    	Gregorio Version 
     
    'gregorio' is not recognized as an internal or external command,
    operable program or batch file.
     
    	Gregorio Location 
     
    ----------------------------------------------------------------------------- 
 
 
	GregorioTeX Location 
     
    -----------------------------------------------------------------------------

rpspringuel:
Also, what sort of documentation do you think should be added for this?

I also just realized that I need to add this tool to the installer manifest for Windows (but don't have the time to actually do that until tomorrow), so don't merge yet.

eroux:
About the documentation, maybe some kind of trouble shooting guide on the website ?

henryso:
Not a criticism, but I wonder how long we need to keep supporting Windows XP when Microsoft has already called an end-of-life on it: http://windows.microsoft.com/en-US/windows/end-support-help

eroux:
As long as people are widely using it I suppose... It's still the second most used OS apparently.

rpspringuel:
I chose to support XP because I still have and use it myself.

Also, I'll point out that in this case, supporting XP isn't all that hard.  Lines 41 and 59 are a bit more complicated than `where luatex` or `where gregorio` (respectively), but they are still single lines of code which continues to work on more current versions of Windows.  If I had to do `if..then..else` statements based on the OS version, then I'd probably be looking to drop XP support sooner rather than later, but as is, I see no pressing need to get rid of it.

contrib/system-setup.bat
```diff
@@ -0,0 +1,75 @@
+@echo off
+SETLOCAL ENABLEEXTENSIONS
+
+set output="c:\system-setup.log"
```
eroux: I suppose this is why it has to be run as Administrator on "modern" windows? Maybe the file could be written in user directory or user temporary directory (`%TEMP%` ?) and then automatically opened with notepad?

contrib/system-setup.bat
```diff
@@ -0,0 +1,75 @@
...
+echo 	GregorioTeX Location >> %output%
+echo. >> %output%
+kpsewhich gregoriotex.sty >> %output% 2>&1
+echo ----------------------------------------------------------------------------- >> %output%
+echo. >> %output%
+echo. >> %output%
+
+echo %output% created.  Please attach the file to an email to 
+echo gregorio-users@gna.org or create an issue at 
+echo http://github.org/gregorio-project/gregorio/issues
```
eroux: maybe adding something like "and copy-paste the content of this file in the description"?

contrib/system-setup.bat
rpspringuel: Actually, I originally had it writing to the same directory as the batch file, but on "modern" Windows, the file wasn't showing up.  It was only by specifying the root that I was able to find the file.

The idea of writing to temp and then opening automatically is worth trying though.

contrib/system-setup.bat
rpspringuel: That's probably a good point thing to add.  It might also be useful to make the contents more markdown friendly.

eroux: Indeed, making it markdown compliant is a really good idea!

rpspringuel:
A more markdown friendly sample, this time from XP with everything installed and working.

###	Gregorio Windows Setup Results 
####	Created: 2015-05-12 
----------------------------------------------------------------------------- 
 
 
###	Windows Version 

Microsoft Windows XP [Version 5.1.2600]
 
 
----------------------------------------------------------------------------- 
 
 
###	LuaTeX Setup 
####	Version 
 
This is LuaTeX, Version beta-0.79.1 (TeX Live 2014/W32TeX) (rev 4971)

Execute  'luatex --credits'  for credits and version details.

There is NO warranty. Redistribution of this software is covered by
the terms of the GNU General Public License, version 2 or (at your option)
any later version. For more information about these matters, see the file
named COPYING and the LuaTeX source.

Copyright 2014 Taco Hoekwater, the LuaTeX Team.

 
####	Location 
 
C:\texlive\2014\bin\win32\luatex.exe 
 
 
#### 	TEXMFLOCAL 
 
C:/texlive/texmf-local
 
 
----------------------------------------------------------------------------- 
 
 
###	Gregorio Setup 
####	Version 
 
Gregorio version 3.0.0-rc1.
Copyright (C) 2006-2015 Gregorio project authors (see CONTRIBUTORS.md)
 
#### 	Location 
 
C:\texlive\2014\bin\win32\gregorio.exe 
 
####	GregorioTeX Location 
 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.sty
 
 
-----------------------------------------------------------------------------

eroux:
Thanks, that looks good! I think it would be useful to search all the tex, lua and font files though, with `kpsewhich -all` so that duplicates, possibly outdated or user-modified files appear... Also, sometimes users have some files in the directory in which they compile, it would be useful to detect them, but I suppose it would require the user to copy the script and run it from where he/she compiles (not a bit problem). Also, if you need to put an absolute path for the output path, [apparently](https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/ntcmds.mspx?mfr=true) `chdir` returns current directory.

rpspringuel:
I've swapped over to writing to `%TEMP%/system-setup.log` and Windows 7 no longer requires the script to be run as administrator.

I'll work on adding the full inventory to the GregorioTeX Location section.

eroux:
An idea crossed my mind... Maybe it would be enough to create a `\gredebugreport` that writes all necessary output to the log file? There are interfaces for kpse in lua, which is the main things we need, and we can print os.name... The advantage is that it will have all the context, and spot possibly wrong files in the compilation directory... What do you think? Sorry to think about that only now...

eroux:
Other advantages:

- works under all OSs
- can tell if lualatex is run in shell-escape mode

rpspringuel:
That won't work because it would necessitate GregorioTeX having been successfully installed.  One of the big reasons for needing this tool (especially on Windows) is because of installation problems causing GregorioTeX to not work.

rpspringuel:
Revised output, am I missing anything?

###	Gregorio Windows Setup Results 
####	Created: 2015-05-12 
----------------------------------------------------------------------------- 
 
 
###	Windows Version 

Microsoft Windows XP [Version 5.1.2600]
 
 
----------------------------------------------------------------------------- 
 
 
###	LuaTeX Setup 
####	Version 
 
This is LuaTeX, Version beta-0.79.1 (TeX Live 2014/W32TeX) (rev 4971)

Execute  'luatex --credits'  for credits and version details.

There is NO warranty. Redistribution of this software is covered by
the terms of the GNU General Public License, version 2 or (at your option)
any later version. For more information about these matters, see the file
named COPYING and the LuaTeX source.

Copyright 2014 Taco Hoekwater, the LuaTeX Team.

 
####	Location 
 
C:\texlive\2014\bin\win32\luatex.exe 
 
 
#### 	TEXMFLOCAL 
 
C:/texlive/texmf-local
 
 
----------------------------------------------------------------------------- 
 
 
###	Gregorio Setup 
####	Version 
 
Gregorio version 3.0.0-rc1.
Copyright (C) 2006-2015 Gregorio project authors (see CONTRIBUTORS.md)
 
#### 	Location 
 
C:\texlive\2014\bin\win32\gregorio.exe 
 
####	GregorioTeX Locations 
 
##### gregoriotex.sty 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.sty
##### gregoriosyms.sty 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriosyms.sty
##### gregoriotex.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.tex
##### gregoriotex-main.tex 
##### gregoriotex-chars.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex-chars.tex
##### gregoriotex-ictus.tex 
##### gregoriotex-signs.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex-signs.tex
##### gregoriotex-symbols.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex-symbols.tex
##### gregoriotex-spaces.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex-spaces.tex
##### gregoriotex-syllable.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex-syllable.tex
##### gregoriotex.lua 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.lua
##### gsp-default.tex 
c:/texlive/texmf-local/tex/luatex/gregoriotex/gsp-default.tex
 
 
-----------------------------------------------------------------------------

eroux:
Right, let's continue that way, I forgot this use case... If it's relevant, it might be useful to add the macro I described, but it can be later

rpspringuel:
Unless I'm missing something, I think I'm ready to merge.

henryso:
Is there some indication of whether the TEXMFLOCAL directory exists?

rpspringuel:
There is not, but I could add that.  I'll work on it tomorrow.

henryso:
Only do that if you think it's important.  I seem to recall an issue (with MikTeX?) where the TEXMFLOCAL directory didn't exist.

rpspringuel:
Would it be better to have the script output a non-existent directory for TEXMFLOCAL and denote that, or to not output a value for TEXMFLOCAL if the directory doesn't exist?

henryso:
I think it would be better to output the value of TEXMFLOCAL as well as indicate if this directory exists.

rpspringuel:
Done and done.

eroux:
Ok to be merged for me

PR #399, opened 2015-05-11
Nabc improvements

eroux:
* make code more generic
* fix bugs
* add example


eroux:
@jakubjelinek can you test a little and tell me if it's ok for you?

jakubjelinek:
I will test it this evening.  Thanks for your help on it.  I'd probably have chosen a different example which would show more significative letters, for ~ 2 line score e.g. Factus est repente again, or AL Veni Sancte Spiritus, or for shorter ones e.g. Loquebantur variis, all from Conventus/Pentecostes/

eroux:
Ok, I changed it

jakubjelinek:
From quick testing, cd doc; make doc doesn't build anymore, it can't find gregoriotex-nabc.lua , while according to strace gregoriotex.lua is loaded according to LUAINPUTS, which in my case is .:../../tex/ .  For gregoriotex-nabc.lua it only tries to open it as relative pathname.  cp -a ../tex/gregoriotex-nabc.lua fixes the build of the documentation.

Trying to build my larger scores, this time with the nabc gregorio installed already, has the same problem, gregoriotex-nabc.lua only searched in the cwd.  If I copy it there, there is another problem:
! Illegal unit of measure (pt inserted).
<to be read again> 
e
l.160 }
     %
? 

This can be reproduced e.g. in checkout of:
git clone https://github.com/jakubjelinek/nabctrial
cd nabctrial
cp -a /usr/share/texlive/texmf-dist/tex/luatex/gregoriotex/gregoriotex-nabc.lua .
rm -f gregall.lua SGModern.ttf gregall.tex gregall.ttf

and applying:

diff --git a/Makefile b/Makefile
index 5d41325..4281cd1 100644
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,9 @@
 		   offertorium-RegesTharsis
 SCORES = $(addsuffix .gtex, $(ANTIPHONS) $(RESPONSORIES) $(PROPERSOFTHEMASS))
 all: test.pdf neumetable.pdf
-test.pdf: test.tex $(SCORES) gregall.ttf SGModern.ttf gregall.tex
+test.pdf: test.tex $(SCORES)
 	lualatex $< && lualatex $<
-neumetable.pdf: neumetable.tex gregall.ttf SGModern.ttf gregall.tex
+neumetable.pdf: neumetable.tex
 	lualatex $< && lualatex $<
 $(SCORES): %.gtex: %.gabc
 	sed 's/+Z)/Z)/' $< | gregorio -s -o $@
diff --git a/test.tex b/test.tex
index f85896c..50b514e 100644
--- a/test.tex
+++ b/test.tex
@@ -18,7 +18,7 @@
 
 \textwidth 5in
 
-\input{gregall.tex}
+%\input{gregall.tex}
 
 % to change the font to something better, you can install the kpfonts package (if not already installed). To do so
 % go open the "TeX Live Manager" in the Menu Start->All Programs->TeX Live 2010

Then make test.pdf
gregoriotex-nabc.lua still has sp after all uses of \\kern, so it is unclear to me what it is complaining about.

jakubjelinek:
On test.pdf, tracked that to ant-magn-vesp2.gabc and the problematic syllables in that file which if commented out fix this are:
Ma(hg|cllss2lsc2)
le(fe|cllss2lsc2)
Changing them to e.g. cllss2lsc3 fixes it.

jakubjelinek:
BTW, the formatting in gregoriotex-nabc.lua is in many places weird, doesn't reflect the nesting level (as if the tab that used to stand for 8 spaces in gregall.lua is now replaced with 2 spaces instead of 8).

Adding debug dump to the end of the parsing routine shows:
cllss2lsc2
\kern 93847.552sp\char58405\kern -412614.656sp\raise 300941.312sp\hbox{\char58504}\kern 232783.872sp\kern -232783.872sp\raise 300941.312sp\hbox{\char58476}\kern -2.9103830456734e-11sp

With the 20150509 gregorio + nabc patch and similarly adjusted gregall.lua, I got for the same nabc
\kern 93847.552sp{\GreGall\char58405}\kern -412614.656sp\raise 300941.312sp\hbox{{\GreGall\char58504}}\kern 232783.872sp\kern -232783.872sp\raise 300941.312sp\hbox{{\GreGall\char58476}}\kern 0sp

Perhaps we need to explicitly flush to zero all values that are smaller (in absolute value) than some epsilon, to avoid lua printing those using scientific format.
Wonder what has changed; the

jakubjelinek:
https://github.com/gregorio-project/gregorio/pull/400 fixes the parsing error, no idea what to do for the loading though.

eroux:
@jakubjelinek can you try with my latest commit (it should fix loading)? Also, What you describe looks like an old bug of LuaTeX, which version do you have? But yes, values that are less than an epsilon should be zeroed. I won't have time to work more on this tonight, I'll try tomorrow.

jakubjelinek:
The zeroing of small values has been solved by the patch you've merged in - the %.3f formats in there.
With the latest nabc branch installed I can build my pentecostes scores (90 pages) just fine and haven't found any issues during quick skimming.
So I think this pull request can be merged.

For GregorioNabcRef.tex, there will be a problem because gregalltab is now local.  Do you think you could add say
    gregoriotex.internal_nabc_table[fontname] = gregalltab[fontname]
line immediately after:
    gregalltab[fontname], gregallmetrics[fontname] = gregallreadfont(font.current())
in gregoriotex-nabc.lua (at least assuming that doesn't copy the whole table, but just adds another reference to it).  That way it would be clear that it is gregoriotex internal, yet GregorioNabcRef.tex could still iterate over it to find out the names of the characters and their corresponding \char values.

eroux:
Can you just try if adding

    gregoriotex.nabc_font_tables = gregalltab

at the end of the file works? If it does, I'll push that tomorrow.

jakubjelinek:
Yes, that works just fine.

eroux:
Ok, pushed in nabc

PR #402, opened 2015-05-12
Nabc

jakubjelinek:
Will still have to describe the nabc grammar, but at least the generated part and tables are in there.


eroux:
This is quite impressive, thanx a lot! I think it could be merged with main documentation though, what do you think? @henryso @rpspringuel any opinion on that?

henryso:
I could go either way on this.  I see merit to combined as well as separate documents.

PR #403, opened 2015-05-12
Reduce *minus* tolerance around bars within syllables

jperon:
With 0.05469 cm, there could be clashes with notes around bars. As for #397, 0.03469 cm is quite arbitrary and empirical.


eroux:
Thanks!

PR #407, opened 2015-05-14
Fixed compiler warnings.

henryso:


jakubjelinek:
Just small comment, in post-K&R C, you should use (void) instead of () for functions taking no arguments, both in prototypes and actual function definitions.

eroux:
I didn't know that, thanx! @henryso are you planning to add the `void` now or should I merge?

henryso:
@eroux I will fix the void problem, so please wait a bit.

jakubjelinek:
In C++ () means the same as (void) in C, but in C e.g.
int foo ();
void bar ()
{
  foo (1, 2, 3);
}
is valid, you get no argument type checking in that case.  In K&R the arguments were declared in the definition like:
int foo (x, y, z)
  int x, y, z;
{
  return x + y + z;
}
but the function definition doesn't serve as a prototype.

eroux:
Very interesting, I didn't know the K&R style... apparently it's been deprecated in C11. I've added the `-Wstrict-prototypes` warning.

henryso:
Ok. Ready to merge unless there's some other warning we should enable.

eroux:
Many thanks for this!

PR #408, opened 2015-05-14
Nabc

jakubjelinek:
Can the nabc branch be merged into develop again?  I'll get at further improving the documentation probably only after Pentecost.


eroux:
Ok let's do that

eroux:
also, I got no answer from Eisiedeln, but I can take the responsibility in case of a problem, so do you think you could add the images of the sfd (and make sure they're not in the final ttf), and make the proper citations in [fonts/README.md](fonts/README.md)? That can be done after Pentecost of course, no emergency.

PR #409, opened 2015-05-14
Removed more measurement glyphs and episemus glyphs from the documentation

### LINKED ISSUE #376: Excluding glyphs from the documentation
As we go forward, it looks like we will be adding more glyphs which are not accessible via gabc, such as the glyphs used for measurement.  I think these should be excluded from the documentation.  Do you agree?

henryso:
Fixes #376.


henryso:
@jakubjelinek, I got an error making the GregorioNabcRef:
```
! LuaTeX error [\directlua]:1: no string to print
stack traceback:
	[C]: in function 'sprint'
	[\directlua]:1: in main chunk.
\dosneume ...ll"]["\luatexluaescapestring {#1}"])}
                                                  \endgroup 
l.850 \sneume{qi!vs}
                   & \texttt{qi!vs} \textit{Cf.} \sneume{qi!vssu2} & \multic...
```

I changed the macro to print an empty space if nil, but you might want to check this.

eroux:
Thanx!

PR #413, opened 2015-05-14
Fixed make distcheck.  Fixes #412.

### LINKED ISSUE #412: make distcheck fails in develop
Upon running `make distcheck`, it fails with the following error:
```
ERROR: files left in build directory after distclean:
./doc/GregorioRef-3_0_0.pdf
./doc/GregorioNabcRef-3_0_0.pdf
Makefile:629: recipe for target 'distcleancheck' failed
make[1]: *** [distcleancheck] Error 1
make[1]: Leaving directory '/home/inxi/develop/gregorio/gregorio-3.0.0/_build'
Makefile:557: recipe for target 'distcheck' failed
```

However, these files should be rightly(?) left behind since they are only cleaned with `maintainer-clean`.  I think I'm at the limits of my autotools understanding to fix this issue.

henryso:
I know the rules look wrong, but if anything depends on something generated, it will be regenerated and `make distcheck` will fail.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -18,29 +18,27 @@
 # GregorioRef.tex must be the first source file
 SRCFILES = GregorioRef.tex Command_Index_gregorio.tex \
 		   Command_Index_internal.tex Command_Index_User.tex \
-		   Appendix_Font_Tables.tex GregorioRef.lua
-NABCSRCFILES = GregorioNabcRef.tex ../examples/FactusEst.gabc
+		   Appendix_Font_Tables.tex GregorioRef.lua factus.gabc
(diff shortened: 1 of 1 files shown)


jakubjelinek:
Isn't the make -C ../src gregorio in two different goals a problem though for parallel make?
I mean won't it be racy in that case?
Perhaps just adding
.NOTPARALLEL:
would be enough for that.
Is there any reason for supporting make dist* btw?  For git based project, one can just use git archive.

henryso:
I wonder if there's a way to prevent the raciness of the make.  Something to look into, I guess.

We need to support `make dist` because the git archive won't have the references or the fonts.

PR #416, opened 2015-05-15
Add background images to the font,

jakubjelinek:
with permission from libraries Stiftsbibliothek, St. Gallen and Kloster Einsiedeln.


eroux:
Did you check that the images are not in the final font? I'm not even sure it can, but it's just to be sure... Also, can you make the pull request against nabc branch? I'll document it a bit before merging (adding the libraries in the thanks part of CONTRIBUTORS, documenting the authorization in fonts/README.md etc.)

jakubjelinek:
Yes, I have verified the generated *.ttf is exactly the same length, a few bytes in the font have changed, but it was under 20 bytes, so I bet just timestamps.

eroux:
ok thanx!

PR #418, opened 2015-05-15
Images in gregall.sfd

eroux:


jakubjelinek:
resuse -> reuse
I wouldn't say the font looks precisely like the codices, the glyphs look similarly, but from the around 100x100 pixel images (often smaller, sometimes larger) it is hard to find out from the shades of grey where exactly to draw the curves, all have been drawn by hand, rather than using some automatic rendering from the images, and in some cases I've even tried to stress more somethign that used to be ambiguous in the original image (mainly, make it slightly more clear which glyphs do and which don't have an episema).

eroux:
Sure, I didn't realize my wording was misleading... Is it better now?

jakubjelinek:
LGTM, thanks.

PR #420, opened 2015-05-15
Changed the note position offset numbers into control sequences.

henryso:
Please review and let me know if you notice something I missed or if I should take a different approach somewhere.

### DIFF
--- a/src/gregoriotex/gregoriotex-position.c
+++ b/src/gregoriotex/gregoriotex-position.c
@@ -24,90 +24,156 @@
 
 #include "gregoriotex.h"
 
-static inline char number_note_before_last_note(
+// TODO: The numbers in the comments are a vestige of the older code; I'm
+// leaving them here for now, in case the refactor missed an instance
+// somewhere, but the numbers should evenually be removed
+
+// (loose) naming convention, employing camel case to be TeX-csname-compliant:
+// {specific-glyph-shape}{note-position}{note-shape}{first-ambitus}{second-ambitus}
+OFFSET_CASE(FinalPunctum); // 0
+OFFSET_CASE(FinalDeminutus); // 1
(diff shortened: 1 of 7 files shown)


eroux:
The C part is ok for me. I'm just wondering if the TeX part might slow things down a little bit, can you make a performance test?

I was thinking about keeping the `\ifcase`, having defined the new csnames to be numbers... but that still is a bit non-abstracted enough...

henryso:
This is the method I used for performance testing.
- Clear the local luatex cache directories used for testing
- Run the 1000 x PopulusSion test 6 times
- Ignoring the time of the first run, average the times

Before this change, average time is 3:22
After this change, average time is 3:24

Therefore, there is a performance impact.  There is always the fallback to ifcase, but how about having Lua generate the ifcase (once per run, of course) so that we don't have to worry about keeping the cases in order?

On a curious note, I'm getting an interesting result with this methodology.  The first run, with the luatex cache directories cleared, is considerably faster (about 85 seconds faster) than the following runs, which are within one second of each other.  One would think that it would be faster with the caches, so this is (to me, at least) a surprising result.  It could also be something specific to my configuration, as my computer is typically under some light load (background processes, etc.).

eroux:
The performance loss is ok, let's keep your solution, you can merge for me!

About the luatex caches, that's an interesting result, I admit I'm a bit surprised... Maybe if you can make a minimal example (with some lorem ipsum) showing that, you could report it on the [luaotfload tracker](https://github.com/lualatex/luaotfload/issues), it's certainly not just you... with modern CPUs, the CPU cost of parsing the font might be smaller than the additional cost of loading the font cache, which is bigger in size (IIRC) and thus requires mors disk I/O... So it might be worth making the cache an option, so that it can be optimized, as on your system the gain seems substantial!

PR #421, opened 2015-05-16
Fixed incorrect offset determination for scandicus and fixed bridges on "below" episemae.

### LINKED ISSUE #419: Episema beneath scandicus
I've been testing the developer build against some outstanding issues I've noted in my scores. The new episema code is working quite well!

Here is a case of the correct episema not appearing as expected, but perhaps this case wasn't foreseen, as I don't see it in any of the tests (I don't think the scandicus is completetest.gabc).

The expected result, (from [Scriptum est enim percutiam](http://gregobase.selapa.net/chant.php?id=2162) as in the Liber Usualis):
screen shot 2015-05-15 at 1 01 09 pm

The episema only partially appears:

    au(j_0k_0l)tem(k.)

screen shot 2015-05-15 at 12 58 57 pm

The correct episema appears, but with a strange duplication on the second note:

    au(j_0!k_0l)tem(k.)

screen shot 2015-05-15 at 12 43 06 pm

henryso:
Fixes #419.

Note: I based this code on #420, so merging this implicitly merges 9b1111641c07fc1f822a33c1fe243bdc61fd78b6 as well.  If we decide to retract #420, but keep this, then close both pull requests and I will make these changes against `develop`.


rarty:
Thanks! This fix does show the correct output for the last case shown here, `j_0!k_0l` which is what I would think is the best/correct way because it is most explicit.

But I still get the first note is missing an episema for `j_0k_0l` and `j_k_l` (which also seem to me like legitimate notations for the same output).

screen shot 2015-05-16 at 9 08 34 am

henryso:
@rarty What version of gregorio (executable) are you using?  These cases seem to work for me.

henryso:
@rarty I can't reproduce your problem.  Would you mind pasting your gabc file and the gtex file produced by gregorio?

rarty:
Ah sorry, you're right. I re-built your pr and ran my test again and it works perfectly. Thanks again.

PR #425, opened 2015-05-16
Moved the definition of cases to lua.

henryso:
@eroux I had most of this done by the time I read your comment in #420.  Feel free to close without merging if you prefer things without it.

With this, the performance improves to 3:23, right in between the two.  I don't know which way is clearer, so I'll go with your opinion, either way.


henryso:
Is this satisfactory?

eroux:
`\RequireLuaModules` does some additional checks in addition to just loading the file, so we would have to make another module for it (more or less separate from GregorioTeX). Ok for me!

PR #426, opened 2015-05-17
Added changelog entries for the horizontal episema improvements and fixes.

henryso:


eroux:
Seems good ot me! Are there some not-so-uncommon where the new system produces a different output than the old one? If so, I think it should be in UPGRADING, if not I can merge

henryso:
I don't know of any case (besides bug fixes like #419) where the new code should produce something different than before.  But then again, I'm not familiar with all the esoteric cases supported by the old code.  Maybe I should add some sort of catch-all phrasing in UPGRADING with a description of the new system?

henryso:
How does that look?

PR #427, opened 2015-05-17
Miktex

rpspringuel:
The installer is now compatible with MiKTeX.

Still need documentation before merging.

Also, need to update version number.  I think this is probably a bugfix version, but am willing to hear arguments to the contrary.


henryso:
If by "bugfix" version, you mean what http://semver.org/ calls PATCH, then I agree. bump the third number.

rpspringuel:
Yes.  That's what I meant.  Sorry about the terminology confusion.

eroux:
Fine for me!

rpspringuel:
Other than the notation in the CHANGELOG, is there any other documentation you think this needs?

eroux:
I don't think so...

PR #428, opened 2015-05-18
Added the ability to force a vertical episema to appear above/below the note.

### LINKED ISSUE #385: force vertical episemus placement top or below
It would be useful to have `'0` and `'1` forcing placement of vertical episemus on top or below the corresponding note (reported on [the ancient bugtracker](https://gna.org/bugs/?22699)).

henryso:
Fixes #385.

One thing I noticed is that the output looks a little funny when a horizontal episema appears above a vertical episema.  The horizontal episema is always pushed up into the next space.  This looks even stranger when there is no line above the vertical episema (i.e., `(m'1_1)`).  Somehow, things are handled properly (or at least better) when the horizontal episema is below the vertical episema (i.e., `(a'0_0)`).  Does anyone know where/how/why the below-case works or any other clues as to how I should proceed?


eroux:
These cases look weird, but users would certainly put one of the episemus on the other side... If you really want to fix them, I think they'd look better with the horizontal episemus close to the note, and then the vertical episemus, but you would have to add quite intrusing changes for these two cases only, so I'm not sure it's worth it (though I have nothing against it).

henryso:
Are you satisfied, then, with this feature as it is, or should I try to fix the weird cases?

eroux:
No, for me it's ok let's merge. If someone complains, we'll fix these, but that's unlikely...

PR #431, opened 2015-05-18
Switched to BUILT_SOURCES to list bison/flex generated files.

henryso:
This seems to work.  What do you think?


henryso:
@rpspringuel Would you be able to tell if this helps any?

Issue #59, opened 2015-02-27
Update each line's vertical spacing


rpspringuel:
Apparently the Solesmes scores only move the lyric text down only on lines where the a low element (note `a'`, `a`, `b'`, or `b`) is present.  Currently gregorio moves the lyrics down for the whole score when this happens.  A user would like to be able to imitate the Solesmes behavior.

This will probably require making two passes per line.  The first to figure out what the lowest note on the line is, the second to actually set the lines.  Perhaps the first pass populates two boxes (one for staff and neumes, the other for text), then the second can simply place them on the page.  However, there might be alignment issues in this approach.

eroux:
To develop a little bit: TeX has its own complex line-breaking algorithm, so we can't just do things at line level easily, though there are [some experiments](https://github.com/eroux/luaparbuilder) [to do it in Lua](http://tex.stackexchange.com/questions/200989/replacements-for-texs-paragraph-builder-using-luatex), but it's really not mature enough, and it will make everything noticeably slower. So let's not count on it and do what we can.

What I propose is the following solution:

 * for each syllable, make gregorio pass the lowest note value to each syllable
 * with this, set a LuaTeX attribute to each syllable
 * set another luatex attribute to the value of the current spacing (also to each syllable)
 * during the second pass on the score in gregoriotex.lua, read on each line if there is a discrepancy between the lowest note and the spacing used
 * if there is a discrepancy, shift all notes up or down (not the easiest part, but feasible)

Do you see what I mean?

eroux:
Note that this won't work in the case where a line breaks in the middle of a syllable and the lowest note is not well placed, but as it will fix 90% of the cases, it's already a good improvement. But if you or I work on it, it would be good to make gregorio also pass lowest note to each glyph, so that we can refine the algorithm later at glyph level, in order to avoid the case I described before.

rpspringuel:
Some discussions I was having today about the Gregorio inspired the following idea:

1. At the beginning of each line, check with the aux file to see how much space is required for this line.  On a first pass of the file, the lack of an aux file would mean that spacing defaults to assuming there were no low notes, and thus lead to overlapping text and notes.
2. Define a tracking variable that would know what the lowest note encountered so far is.  At the end of every line, write the current value of this tracking variable to a line (and score) specific variable in the aux file and then reset it to zero.
[...]

PR #440, opened 2015-05-21
Pass glyph high/low pitches.

henryso:
This change is for #59.

I piggybacked this on #439 because it touches the same code.  If #439 is rejected, close this and I will re-do these changes against develop.

This is ready for review.


eroux:
Ok for me too!

PR #444, opened 2015-05-21
Space around bars : reduce negative tolerance

jperon:
After previous change, there where other clashes yet. As default space around bars within syllables is already quite small, this PR suggests reducing negative tolerance to nearly nothing.


eroux:
merging, though spaces around bars may have some modifications soonish

PR #448, opened 2015-05-22
Modified to pass the heights (pitches) as numbers to TeX.

### LINKED ISSUE #447: Using numbers for pitch/height
I've been thinking about this as of late, and with #59, #429, and #446, it seems that it would be increasingly valuable to pass a number instead of a letter for the pitch or height of a sign, with `a` being passed as `3`.

This way, numbers can be compared more easily for the min/max height determination for #59, adjusted for the sign change computation for #446, and (as a byproduct) `\ifcase`..`\or`..`\or`..`\fi` should be theoretically faster than `\if`..`\fi\if`..`\fi\if`..`\fi`.

For the 5-line staff (#429), we would be skipping `o` in the gabc, and we already hack `z` and `y` to be the two pitches below `a`, so why not just pass numbers instead to eliminate potentially wrong association of letter with pitch?

If we agree to this, I can make the changes, if so desired, but it may disrupt work in progress, especially on #59.

henryso:
Fixes #447.

This is ready for review.  I *think* I've gotten all the spots in TeX, but it's possible I've missed something.  The PDF comparison tests pass, and the gabc-to-gtex tests fail as expected.

### DIFF
--- a/src/dump/dump.c
+++ b/src/dump/dump.c
@@ -698,6 +698,16 @@ static const char *dump_vposition(gregorio_vposition vpos) {
     }
 }
 
+static const char *dump_pitch(const char height) {
+    static char buf[20];
+    if (height >= LOWEST_PITCH && height <= HIGHEST_PITCH) {
+        snprintf(buf, 20, "%c", height + 'a' - LOWEST_PITCH);
+    } else {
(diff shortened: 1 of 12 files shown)


eroux:
ok for me!

PR #449, opened 2015-05-22
Forced the pitch into range for an automatic custos on clef change.

### LINKED ISSUE #446: error:pitch difference too high to set automatic custo (z0)
I encounter this error on a real case : the [*Univérsi*](http://gregobase.selapa.net/chant.php?id=1747) gradual. When changing from `f3` to `c4` with `z0::c4`, gregorio throws this error. A possible workaround (inspired by GregoBase) would be `f+!//::c4`, but if this could be automatically computed, this would be better.

henryso:
Fixes #446.  This pull request piggybacks on #448, so merging this will merge the other.

This is ready for review.


eroux:
So now an error is not possible anymore? If it's the case ok for me

henryso:
Yes, an error is no longer possible, unless we do something silly like make `HIGHEST_PITCH` less than `LOWEST_PITCH + 7`.

PR #450, opened 2015-05-23
Pass the first syllable and its first letter to macros for styling.

henryso:
This is something that I've needed for a while.  Feel free to reject/close this if you don't think it's a good idea.  Otherwise, it's ready for review.  In doing this, I also cleaned up the code a bit.


eroux:
ok for me (I thought it would add some warnings to clang, but it doesn't)

PR #452, opened 2015-05-25
Added verbatim and special character handling to the first syllable code.

henryso:
This is a bug fix for #450.

I have run the tests and they seem to be fine after adjustment for the bug fix.  I will merge this immediately so that @rpspringuel can continue testing, but please review as time permits.

@rpspringuel: This should fix the bug I mentioned in gregorio-test issue 2, but I don't know if it will solve your segfault issue.  This may just be a fix for a "nearby" bug to the segfault, but it does have the potential to correct the segfault because of the convoluted nature of `gregorio_rebuild_characters` in `characters.c`.  You will need updated test expectations will I will push immediately after merging this.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -1066,10 +1066,27 @@ void gregorio_rebuild_first_syllable(gregorio_character **param_character,
         gregorio_insert_style_before(ST_T_BEGIN, ST_FIRST_SYLLABLE,
                 current_character);
         do {
-            if (current_character->is_character && !marked_syllable_iniital) {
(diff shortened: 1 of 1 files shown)


rpspringuel:
Note: You don't need to merge a pull request in order for me to test it.  I've configured git so that it is aware of all of the fork repositories and can pull from anyone of them.  You just need to let me know what you need tested and I can pull it directly from your repo.

henryso:
I'll keep that in mind for the future.  Thanks!

Issue #383, opened 2015-05-07
TeX macro naming conventions


rpspringuel:
I figured I'd open in issue for this so that it can be listed in the milestones for 4.0 and so people can comment on it if the feel the need.

1. functions internal to gregoriotex are named `\gre@*`
2. functions which appear in the gtex output of gregorio, but which should not be accessed by users are named `\gre*`
3. functions which are meant for user access and use have no prefix

henryso:
How do we name functions for user access to minimize conflict with other packages?

rpspringuel:
Short answer: we don't, just like every other LaTeX package.

Long answer: since the internal and gtex functions are prefixed, we shouldn't have to worry about them.  For the user functions, we define them conditionally, raising an error if there is a name clash.  If we start getting lots of bug reports about name clashes with a specific function, then we consider renaming that function.

eroux:
I'm not against not prefixing user commands, but it should be explicit that they are intended for gregorio scores in their names... the reason why, for instance, luatexbase or fontspec doesn't prefix its commands is because it can't be compatible with another system of the same type, but I don't think gregorio should be incompatible with another gregorian chant style.

Besides, LaTeX is heavily criticized for its lack of coherence and is considered a huge mess by many (who prefer ConTeXt), so I'm not sure it's a very good model...

rpspringuel:
I don't disagree with the sentiment, I was just defaulting to the current standards for LaTeX because I'm familiar with them.

Perhaps the following system would be better:

1. User commands are all lowercase and are prefixed with `\gre`.  E.g. `\greincludescore` (this is similar to the LaTeX standards for user commands, except with the prefix added)
2. gtex commands use CamelCase for their names with a `\GRE` or `\Gre` prefix.  E.g. `\GreBarSyllable` or `\GREBarSyllable` (this is similar to the LaTeX standards for class and package commands, except with the prefix added)
3. internal GregorioTeX commands are prefixed `\gre@`  E.g. `\gre@warning` (this is the LaTeX standard for internal private commands)

In this fashion all of our commands are prefixed to identify them as ours, and the different prefix identifies which of the three groupings the command falls under.

This would require some changes to `gregoriotex-write.c`, but alternatively we could switch the convention for 1 and 2 to avoid that (yielding `\GreIncludeScore` or `\GREIncludeScore` and `\grebarsyllable`, which is what that last macro is currently called).

eroux:
Another proposal:

[...]

PR #453, opened 2015-05-25
Feature/texrenaming

rpspringuel:
The first part of the process of dragging the function names into order (addressing #383).  This commit renames all the Gregorio controls to have CamelCase names.  Functions which had CamelCase names but which were not Gregorio controls have been renamed to have names more appropriate to their class (though these are only draft names at this point).  User controls that have been renamed in this fashion have been given deprecation warnings on the CamelCase name.

This phase can be merged (if desired) as all tests pass except the gabc-gtex ones (which are expected to fail since we're renaming the functions which get written to the gtex files) and the output test for symbols (which has `\def\grestar{\greheightstar}` in it which will have to be replaced by `\def\GreStar{\greheightstar}`).


doc/Command_Index_internal.tex
```diff
@@ -430,6 +430,42 @@ \subsection{Note Alignment Type}\label{notesalign}
   12 & divisio finalis
  \end{tabular}
 
+\subsection{Episemus Special}\label{EpisemusSpecial}
```
eroux: To be confirmed, but the signature is changed, it doesn't use numbers anymore

eroux:
Looks globally good apart from a comment, but I also think these conventions should be documented

doc/Command_Index_internal.tex
rpspringuel: I'll admit I was just copying what I had done before and hadn't looked at the code to see if the meaning of the arguments had changed.

rpspringuel:
What sort of documentation are you thinking of?  I assume you mean something beyond GregorioRef, but am not quite sure what form that should take.

eroux:
I just meant a few sentences about this in GregorioRef (if you did this already, sorry, I have skipped that)

rpspringuel:
Looking through the file, I don't have anything explicit.  I'll add something.

rpspringuel:
I'd like to incorporate the changes from #460 and #462 to this before considering merging it.

rpspringuel:
I don't have time to test right now, but I think that should take care of deprecating the `gregoriotex-font` header.  Please double check my changes to `gabc-score-determination.y`

rpspringuel:
I fixed the changes and started testing.  The test suite fails where it's expected to fail and no where else.

However, since I don't see any gregorio warning messages when doing that, I manually ran the following gabc file through:

```
% !TEX root = Main.tex
% !TEX encoding = UTF-8

name:Abraham pater vester;
office-part:Antiphona;
mode:2;
book:The Liber Usualis, 1961, p. 577;
transcriber:Andrew Hinkley;
centering-scheme: english;
gregoriotex-font: gregorio;
language: english;
%initial-style: 0;

def-m1:\greblockcustos;
def-m2:\greblockcusto=0\relax;

%%


%%
(f3) test(b) test(G) (;g+) test(b) (z0::f2) test(a) (::z)
test(a)
```

Now this should trigger both the `centering-scheme` and `gregoriotex-font` warnings.  However, my command line output, looks like this:

```
gregorio -vW test.gabc 
 Looking for Latin in /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregorio-vowels.dat
 Could not find Latin in /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregorio-vowels.dat
 in voice 1 the first element is a key definition, considered as initial key
```

It appears that neither deprecation warning is getting triggered.  Also, shouldn't it be looking for `english` in `gregorio-vowels.dat`?  Is there a problem with the message or with what the program is doing?

rpspringuel:
Hold on, I was in the wrong directory.  Everything is working fine.

henryso:
What's left to do for this?

rpspringuel:
Nothing, so far as I can tell.  This phase is ready for merging.

henryso:
The fix 347 test is failing.
```
! Undefined control sequence.
<argument> \gresmallcaps 
              {\greboldfont {VI}}
l.6 \GreMode{6}
             %
```

henryso:
Based on the changes for #476, is the mode number for fix-347 supposed to be so high above the initial?

rpspringuel:
Yes.  #476 pushed the default position of the annotations up so that the baseline of the top annotation would align with the top line of the staff.  I.e. I took the behavior of `\gresetfirstannotation` (or calling `\gresetfirstlineaboveinitial` with the same thing in both arguments) as the expected behavior.

henryso:
What is the replacement for `\setgrefactor`?  Should the replacement be in the UPGRADE.md?

rpspringuel:
That's a user command and not covered by this pull request.  It will be in the next phase (renaming the user commands).  My plan is to replace it by `\grechangestaffsize`.

henryso:
Is there something I can put temporarily in the `tex-output/sample/sample.tex` and `tex-output/translation/translation.tex` which will set grefactor to 15 as before or would you just accept the current output for the time being?

rpspringuel:
`\setgrefactor` should still work.  I haven't changed that function yet.  If there's a mismatch due to a change in grefactor, then there's a bug that needs to be worked out.

henryso:
For those two tests, then, it seems that `\setgrefactor{15}` is not doing what it should.  The score is larger by a tiny bit with this branch.

rpspringuel:
Okay, that should fix that problem.  Rechecking the tests and I find the following failures:

All `gabc->gtex`: expected since this PR renames all `gtex` commands
`gabc->output` fix-347: due to moving of annotation
`tex->output` PopulusSion, sign-styles, and symbols: due to moving of annotation.

Those are all expected so unless there's something we're not testing for, we should be good to go.

henryso:
`sign-styles` is also failing because `\greusemedicaeastyle`, `\greusehufnagelstyle`, and `\greusemensuralstyle` and not pulling in the appropriate glyphs any more.

rpspringuel:
There was a hidden dependency which my recent changes broke.  I've now removed that dependency so that the appropriate glyphs should be pulled in just fine.

henryso:
Looks good to me.  Merging.

PR #455, opened 2015-05-25
Added the ability to use nabc for a choral sign.

henryso:
Submitted for your approval and/or review.

This adds the ability to use nabc easily in a choral sign by using `[cn:nabc]` where you would previously have had to use `[cs:\gregorionabcchar{nabc}]`.

Feel free to reject it if you don't like the idea.

### DIFF
--- a/src/gabc/gabc-notes-determination.l
+++ b/src/gabc/gabc-notes-determination.l
@@ -149,6 +149,7 @@ static inline void add_alteration(const gregorio_type type) {
 %x texverbglyph
 %x texverbelement
 %x choralsign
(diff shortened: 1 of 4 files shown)


eroux:
do you have a small image example of the result? Just for the issue log

henryso:
Example:
```
name: example;
initial-style: 0;
%%
(c4) ( ) ...(;) a(f)le(fh/ih)lu(gh[cn:to>]g)ia,(e) ...(,)
```
produces
example

eroux:
ok!

PR #456, opened 2015-05-25
Changing redlines to use gregoriocolor

rpspringuel:
Addresses [#21787 on old tracker](https://gna.org/bugs/index.php?21787)
This is one of a bunch of changes which I had rolled into texrenaming.  I'm now pulling those changes out individually to separate the functionality changes from the application of the naming conventions.


eroux:
Ok for me

PR #459, opened 2015-05-26
Enabled 'z' to force-justify the last line of a score.

### LINKED ISSUE #43: Last line of Score alignment
Currently the last line of a score is always set ragged.  It should be possible to manually override this in either the gabc or TeX so that the last line is justified.  It is possible to get this effect by messing around in the gtex file currently.

1) End your score with just `(z)`.

2) Run the score through gregorio and open the generated gtex file.

3) At the end of the file you'll see something like this (I've removed empty lines to save space):

> \grebarsyllable{}{}{}{1}{\gresetnextsyllable{}{}{}}{}{0}{%
> \grelastofline %
> }{%
> \gredivisiofinalis{}%
> }%
> \grenewline %
> \endgregorioscore %
> \endinput %

Modify this by changing `\grelastofline` to `\grelastofscore` and moving `\grenewline` to be right after `\gredivisiofinalis{}`.  You should end up with something like the following:

> \grebarsyllable{}{}{}{1}{\gresetnextsyllable{}{}{}}{}{0}{%
> \grelastofscore %
> }{%
> \gredivisiofinalis{}%
> \grenewline %
> }%
> \endgregorioscore %
> \endinput %

4) Now typeset the main document that includes this score and you should get the divisio finalis all the way to the right hand side of the score and the neumes spaced to fill the line.

I think that there should either be a TeX switch that adds code to `\grefinaldivisiofinalis{0}` (what you usually get for the last syllable you end your gabc with `(::)`) or gregorio should recognize `(::z)` as requiring that code.

Note: Adding `\grenewline` to between lines 1534 and 1536 of `gregoriotex-signs.tex` has the desired effect.  All that's needed is a switch. 

Also, are there ever occasions when `\grefinaldivisiofianlis{0}` doesn't end the score?  My quick testing didn't reveal any.  If this is true, then `\grelastofline` in this function should be changed to `\grelastofscore` (though it doesn't seem to have any visible effect in my testing).

henryso:
Fixes #43.

In implementing this, I found a bug in the way elements are "cut", which I think I've fixed.  The tests pass, except for one which ended with a 'Z', and whose result is acceptable.

This is ready for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - The dot in a space directly above a punctum in a descending neume is now placed slightly higher in the space (see [#386](https://github.com/gregorio-project/gregorio/issues/386) and [Gna! bug 21737](https://gna.org/bugs/?21737)).
 - Choral signs are now positioned correctly around porrectus and torculus resupinus (see [#387](https://github.com/gregorio-project/gregorio/issues/387) and [Gna! bug 22025](https://gna.org/bugs/?22025)).
 - Gregorio will now try harder to select an appropriate pitch for an automatic custos (`z0`) on a clef change (see [#446](https://github.com/gregorio-project/gregorio/issues/446)).  If results are not satisfactory, use a manual custos (`+`) to select a pitch manually.
+- The final line of a score may now be forced to be fully justified (rather than ragged) by appending a `z` to the final `:` or `::` of the score (see [#43](https://github.com/gregorio-project/gregorio/issues/43)).
 
 ### Changed
 - The spacing algorithm has completely changed, expect your scores to look quite different (better we hope).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - The dot in a space directly above a punctum in a descending neume is now placed slightly higher in the space (see [#386](https://github.com/gregorio-project/gregorio/issues/386) and [Gna! bug 21737](https://gna.org/bugs/?21737)).
 - Choral signs are now positioned correctly around porrectus and torculus resupinus (see [#387](https://github.com/gregorio-project/gregorio/issues/387) and [Gna! bug 22025](https://gna.org/bugs/?22025)).
 - Gregorio will now try harder to select an appropriate pitch for an automatic custos (`z0`) on a clef change (see [#446](https://github.com/gregorio-project/gregorio/issues/446)).  If results are not satisfactory, use a manual custos (`+`) to select a pitch manually.
-- The final line of a score may now be forced to be fully justified (rather than ragged) by appending a `z` to the final `:` or `::` of the score (see [#43](https://github.com/gregorio-project/gregorio/issues/43)).
 
 ### Changed
 - The spacing algorithm has completely changed, expect your scores to look quite different (better we hope).
@@ -46,6 +45,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Support for `lualatex -recorder`.  Autocompiled gabc and gtex files will now be properly recorded so that programs like `latexmk -recorder` can detect the need to rebuild the PDF when a gabc file changes.
 - A vertical episema may now be forced to appear above or below a note.  In gabc, use `'0` for the vertical episema to appear below and `'1` for the vertical episema to appear above (see [#385](https://github.com/gregorio-project/gregorio/issues/385)).
 - The first syllable and first letter of the first syllable that is *not* interpreted as the initial of the score are now passed to macros that allow them to be styled from TeX.  The first syllable is passed to `\GreFirstSyllable#1` and the first letter of the first syllable is passed to `\GreFirstSyllableInitial#1`.
+- The final line of a score may now be forced to be fully justified (rather than ragged) using `\grejustifiedlastline` before including the score.  Use `\greraggedlastline` to switch back to a ragged last line.  See [#43](https://github.com/gregorio-project/gregorio/issues/43).
 
 ### Removed
 - GregorioXML and OpusTeX output
--- a/src/gabc/gabc-elements-determination.c
+++ b/src/gabc/gabc-elements-determination.c
@@ -62,13 +62,15 @@ static inline bool is_puncta_descendens(char glyph)
  */
 
 static void close_element(gregorio_element **current_element,
-        gregorio_glyph *first_glyph)
+        gregorio_glyph **const first_glyph,
+        const gregorio_glyph *const current_glyph)
 {
-    gregorio_add_element(current_element, first_glyph);
-    if (first_glyph && first_glyph->previous) {
-        first_glyph->previous->next = NULL;
-        first_glyph->previous = NULL;
+    gregorio_add_element(current_element, *first_glyph);
+    if (*first_glyph && (*first_glyph)->previous) {
+        (*first_glyph)->previous->next = NULL;
+        (*first_glyph)->previous = NULL;
     }
+    *first_glyph = current_glyph->next;
 }
 
 /*
@@ -82,7 +84,8 @@ static inline void cut_before(gregorio_glyph *current_glyph,
                               gregorio_element **current_element)
 {
     if (*first_glyph != current_glyph) {
-        close_element(current_element, *first_glyph);
+        close_element(current_element, first_glyph, current_glyph);
+        // yes, this is changing value close_element sets for first_glyph
         *first_glyph = current_glyph;
         *previous_glyph = current_glyph;
     }
@@ -110,7 +113,7 @@ static gregorio_element *gabc_det_elements_from_glyphs(
     // first_glyph
     gregorio_glyph *previous_glyph = current_glyph;
     // a char that is necessary to determine some cases
-    char do_not_cut = 0;
+    bool do_not_cut = false;
     // a char that is necesarry to determine the type of the current_glyph
     char current_glyph_type;
 
@@ -125,7 +128,7 @@ static gregorio_element *gabc_det_elements_from_glyphs(
         if (current_glyph->type != GRE_GLYPH) {
             if (current_glyph->type == GRE_MANUAL_CUSTOS) {
                 first_element = current_element;
-                close_element(&current_element, first_glyph);
+                close_element(&current_element, &first_glyph, current_glyph);
                 current_glyph = current_glyph->next;
                 continue;
             }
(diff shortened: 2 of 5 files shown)


eroux:
Just to be sure; did you check that the (vertical) space after the score is the same with this change? I don't think it changed, but it's something that is rarely tested and not obvious (sometimes the change in spacing is very small).

henryso:
There is a difference in the vertical space, about the height of one character.  I'll see if I can fix it.

henryso:
I'm stumped.  I don't know what is generating that extra space.  Any hints?

eroux:
There's a code in gregoriotex.lua that removes empty lines, there is maybe something it doesn't manage to remove here...

henryso:
The line that *both* adds the extra space *and* enables the justification is `\grepenalty{-10001}` which evaluates (in this case) to `\penalty-10001`.

I took a look at the Lua code, and for the case *without* the `z`, the `process` function doesn't remove any lines.  In the case *with* the `z`, it *does* remove a single line.

I don't claim to understand the underlying TeX structure, so I don't feel confident messing with this function.

eroux:
Debugging this is not easy, but I think you'll learn a lot... What I'd probably do is inspect the number of lines in the `post_linebreak` lua hook, it's relatively easy, and you'll be able to see what the lines contain, and if there is something that should be removed...

henryso:
That's the thing...  The number of lines is correct (after the empty line is removed) and the last non-empty line contains, apparently, the content of the last line.  Are you saying I should look to remove something from within the last line?

eroux:
Would you have a minimal example showing the problem?

henryso:
###### fix-43.gabc
```
name: test for force justify of last line via 'z';
%%
(c3)
g(g) g(g) g(g) g(g) g(g)
g(g) g(g) g(g) g(g) g(g)
g(g) g(g) g(g) g(g) g(g)
g(g) g(g) g(g) g(g) g(g)
g(g) g(g) g(g) g(g) g(g)
g(g) g(g) (::f+z)
```
###### fix-43.tex
```
\documentclass[11pt]{article}
\usepackage[debug=all]{gregoriotex}
\usepackage{libertine}
\begin{document}
\includescore[f]{fix-43}
*****
\end{document}
```

eroux:
A clue:

    diff --git a/tex/gregoriotex.lua b/tex/gregoriotex.lua
    index e61ee1b..3020140 100644
    --- a/tex/gregoriotex.lua
    +++ b/tex/gregoriotex.lua
    @@ -160,6 +160,9 @@ local function process (h, groupcode, glyphes)
           currentshift=0
         end
       end
    +  for n in node.traverse(h) do
    +    texio.write_nl('found a '..node.type(n.id))
    +  end
       -- due to special cases, we don't return h here (see comments in bug #20974)
       return true
     end 

Another one:

    \setlength{\baselineskip}{0pt plus 0pt minus 0pt}

eroux:
BTW, setting `baselineskip` to 0 (or a user defined value) for the time of a score is probably a good idea...

henryso:
So it looks like there's an extra glue at the end.  I'll look into this more closely tonight.

eroux:
I think TeX adds `\baselineskip` after each line, and when we remove one we forget to remove the `\beaselineskip` afterwards. So the bug certainly appears in other circumstances.

henryso:
The glue subtype appears to be 1 (n.subtype, using the code above, unless I'm doing this wrong), which, from the documentation, should be `\lineskip`, unless I'm reading that wrong, too.

eroux:
baselineskip might be a LaTeX wrapper around lineskip yes

henryso:
The extra glue is actually before the empty line that we remove.  I tried removing this glue when the empty line is removed, but this only shrinks the vertical size very slightly, and certainly not a character's height.

henryso:
The line prior to the glue before the empty line looks like the last line produced with no `z`, except that there is no `hfill` before the `rightskip` at the end.  There's some other magic going on here.

henryso:
Setting `baselineskip` to 0 made no difference.  However, setting `lineskip` to 0 did tighten things up.  Not only did the space below the score shrink, but the space between lines within the score also shrunk.  I don't know if this is the right thing to do, though.

henryso:
That said, even with `lineskip` set to 0, it is still tighter when you leave out the `z`.  I'm out of ideas, at least for now.  Maybe I'll think of something as time goes by.

eroux:
Strange, in my tests setting baselineskip to 0 did work... I'm trying with

    \documentclass[11pt]{article}
    \usepackage{gregoriotex}
    \usepackage{libertine}
    \begin{document}
    
    \setlength{\baselineskip}{0pt plus 0pt minus 0pt}
    
    \includescore[a]{fix-43}
    *****
    
    \includescore[a]{fix-43-noz}
    ****
    \end{document}

What's your test?

henryso:
I actually set baselineskip within `begingregorioscore` and set it back in `endgregorioscore`  Perhaps the key is there somewhere.

eroux:
if you reset it in endgregorioscore, you must do it after `\par`, otherwise it won't work

henryso:
I'm resetting it right before `\relax` at the end.  If I *don't* reset it, then it does *appear* to work.

rpspringuel:
Be sure to have a bunch of text (multiple lines) after the score to make sure that it appears right.  You can use `\lipsum` from the [lipsum package](https://www.ctan.org/pkg/lipsum?lang=en) to create such a block of text really quick.

henryso:
Yes, I am testing with Lorem Ipsum text (though not `\lipsum`).  That said, I don't have a good solution yet.

henryso:
An interesting observation: If the last line is *incidentally* justified by TeX (when the notes just happen to line up that way, even *without* the `z` at the end), then I see the extra space.  In this case, also, the callback is removing an empty line.

henryso:
I think I've gotten to the bottom of the cause, but I don't know how to fix it correctly.

In both the `z` case and the incidentally justified case, there is a negative penalty before the `\rightskip` at the end of the last line.  In the case of `z`, this value is fixed at -10001, and comes from the `\grenewlinecommon` macro.  In the case of the incidentally justified case, it's just some negative value.

If I remove the penalty in `\grenewlinecommon`, the space below the score magically goes away, so one possibility is to call the newline macro differently from within the "finaldivisio" macro which would cause it to omit the penalty.  However, this does not fix the incidental case.  Is there a proper way to deal with this, or should I muck around with the final penalty of a score in Lua (how would I know a given line is *really* the final line of a score)?

henryso:
After thinking about this for a while, I'm thinking the best thing to do would be to put a `\grepenalty{-10001}` in \endgregorioscore.  Then you always get the space and everything remains consistent regardless of how the score ends.  However, this may just be my frustration talking.  Anyone have any other ideas?

eroux:
That would make things consistent indeed... I'll try to inspect that later today.

eroux:
Here is another approach: use

    \parfillskip 0pt plus 0pt minus 0pt

before your score, and this patch (which fixes bugs, so it should be applied anyway):

    diff --git a/tex/gregoriotex-syllable.tex b/tex/gregoriotex-syllable.tex
    index 9d82edb..3cfd7d1 100644
    --- a/tex/gregoriotex-syllable.tex
    +++ b/tex/gregoriotex-syllable.tex
    @@ -729,32 +729,33 @@
         \fi %
         #9\relax %
         \grepenalty{\greendafterbaraltpenalty }% TODO: isn't it a bit buggy?
    -    \gre@debugmsg{ifdim}{ temp@skip@two < -wd(GreSyllablenotes)}
    -    \ifdim\gre@skip@temp@two < -\wd\GreSyllablenotes %
    -      \gre@debugmsg{ifdim}{ nextbegindifference > 0pt}
    -      \ifdim\gre@skip@nextbegindifference > 0 pt%
    -        \gre@skip@temp@one = \gre@skip@interwordspacetext%
    -        \grehskip\gre@skip@temp@one %
    -      \else % \ifdim\gre@skip@nextbegindifference > 0 pt
    -        \gre@skip@temp@one = \gre@skip@interwordspacetext%
    -        \grehskip\gre@skip@temp@one %
    -      \fi %
    -    \else 
    -      %\ifdim\gre@skip@temp@two < -\wd\GreSyllablenotes
    -        %\gre@skip@temp@two=\wd\GreSyllablenotes 
    -        %\divide\gre@skip@temp@two by 2\relax 
    -        %\kern -\gre@skip@temp@two 
    -      \gre@skip@temp@two=\gre@skip@temp@three %
    -      \gre@debugmsg{ifdim}{ nextbegindifference < 0pt}%
    -      \ifdim\gre@skip@nextbegindifference < 0 pt%
    -        \advance\gre@skip@temp@two by \gre@skip@nextbegindifference %
    -      \fi %
    -      \gre@debugmsg{ifdim}{ temp@skip@two > -wd(GreSyllablenotes)}%
    -      \ifdim\gre@skip@temp@two > -\wd\GreSyllablenotes %
    -        \kern\gre@skip@temp@two %
    -      \else % \ifdim\gre@dimen@temp@five > -\wd\GreSyllablenotes 
    -        \gre@skip@temp@one = -\wd\GreSyllablenotes %
    -        \grehskip\gre@skip@temp@one%
    +    \gre@debugmsg{ifdim}{ temp@skip@two < -wd(GreSyllablenotes)}%
    +    % end of same code as syllable
    +    \ifnum\grelastoflinecount=1\relax %
    +      \global\grelastoflinecount=2\relax %
    +    \else %
    +      \ifdim\gre@skip@temp@two < -\wd\GreSyllablenotes %
    +        \gre@debugmsg{ifdim}{ nextbegindifference > 0pt}%
    +        \ifdim\gre@skip@nextbegindifference > 0 pt%
    +          \gre@skip@temp@one = \gre@skip@interwordspacetext %
    +          \grehskip\gre@skip@temp@one %
    +        \else % \ifdim\gre@skip@nextbegindifference > 0 pt
    +          \gre@skip@temp@one = \gre@skip@interwordspacetext %
    +          \grehskip\gre@skip@temp@one %
    +        \fi %
    +      \else %
    +        \gre@skip@temp@two=\gre@skip@temp@three %
    +        \gre@debugmsg{ifdim}{ nextbegindifference < 0pt}%
    +        \ifdim\gre@skip@nextbegindifference < 0 pt%
    +          \advance\gre@skip@temp@two by \gre@skip@nextbegindifference %
    +        \fi %
    +        \gre@debugmsg{ifdim}{ temp@skip@two > -wd(GreSyllablenotes)}%
    +        \ifdim\gre@skip@temp@two > -\wd\GreSyllablenotes %
    +          \kern\gre@skip@temp@two %
    +        \else % \ifdim\gre@dimen@temp@five > -\wd\GreSyllablenotes 
    +          \gre@skip@temp@one = -\wd\GreSyllablenotes %
    +          \grehskip\gre@skip@temp@one%
    +        \fi %
           \fi %
         \fi %
       % then the most simple : the case where there is something to write under the bar. We just need to adjust the spaces.
    @@ -787,7 +788,7 @@
         \fi %
         %and that's it !!
       \fi %
    -  \grefirstsyllableofword=1 %
    +  \grefirstsyllableofword=1\relax %
       \global\gre@dimen@notesaligncenter= 0 pt\relax % very important, see flat and natural
       \greunsetfixedtextformat %
       \ifnum\greblockcusto=1\relax\ifnum\greinsidediscretionary=0\relax %

eroux:
That might remove the need for the final z, with a TeX-only solution... It might be better, what do you think?

eroux:
I don't know what's best to document it or make it more accessible to users, maybe setting `parfillskip` to a user defined value in `begingregorioscore` according to some `\greAlignLastLineRight` or something?

henryso:
I certainly do like the TeX approach better, but as this was proposed by @rpspringuel in #43, I'd like his opinion as well.

So if I understand the complete fix, I should merge in #461, get rid of the C code here that allows the `z` on the final bar, and create some sort of macro that instructs `begingregorioscore` to set `parfillskip` to zero.  If this is right, then it sounds good ti me.

I don't think it makes much sense to use a user-defined value here.  I think you either want ragged or justified.  However, I can be easily convinced otherwise.  Is there a reason for a user-defined value for `parfillskip`?

How about `\grejustifiedlastline` and `\greraggedlastline` (I don't think we have agreed to camel-case here, at least not yet) to switch between the behaviors?  I suggest "justified" because this doesn't do right-alignment in the traditional sense.

eroux:
Ok, let's wait for @rpspringuel 's answer, but otherwise you're right about what to do yes. User defined value might be useful, for instance if you don't want a blank of more than x cm at the end of your score, etc. But `grejustifiedlastline`, etc. are ok too, the two systems can live together.

rpspringuel:
While I may have been the author of the Issue here, it arose out of [a discussion on the user list](http://www.mail-archive.com/gregorio-users%40gna.org/msg02145.html) where Joei wanted to have the last line justified like the first.

Personally, I've no problem with an entirely TeX solution as it makes the gabc file more versatile.  As for the possibility of a user defined value for `parfillskip`, there are already mechanisms for changing that value.  If the user wants that level of control, then they should be able to make those changes themselves.  For most users ragged and justified should be enough.

eroux:
ok then, let's do that

henryso:
I made the changes specified here.  Please review.

eroux:
looks ok to me!

rpspringuel:
I see not problems either, but haven't done any real testing myself.

PR #461, opened 2015-05-27
no space after final bar, fix missing %

eroux:


henryso:
Thanks for finding this, I was beginning to get rather frustrated with my inability to find the cause of the extra space.

eroux:
I fixed the debug message. The `\relax` might not be necessary here, but I've become very cautious not to end a line with `1%` (or any number). The reason is that sometimes, you have a macro afterwards that will expand to another number, and these might get merged, while with `\relax` you're sure the expand ends where you want, and if another number appears (because of a bug in most cases), you'll see it in the output, and won't get some confused output. So this is mainly to prevent bugs which are sometimes extremely difficult to find.

PR #462, opened 2015-05-27
Feature/stylingenvironments

rpspringuel:
This creates a unified system for handling formatting of score elements and provides the user with an easy to use interface for changing those formats.  I've retained backwards compatibility, at the expense of redundant, and therefore slightly bloated, code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,18 +35,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
   - `lowchoralsignshift` is now `choralsigndownshift`.
   - `highchoralsignshift` is now `choralsignupshift` and its sign is now inverted.
 - `\grecoloredlines` now takes a single argument, a named color, instead of the three components of an RGB color.  As a result, `\redlines` can now use `gregoriocolor`, making the red staff lines consistent with the text, even when the user teaks `gregoriocolor`.  Addresses [#21787 on the old tracker](https://gna.org/bugs/index.php?21787).
-- All formats can now be changed via the `\grechangeformat` command.  This command takes two or three arguments.  The first argument, required, is the format to be altered.  The defined formats are:
-    - `initial`: for formatting normal initials
-    - `biginitial`: for formatting big (2-line) initials
-    - `translation`: for formatting translation text
-    - `abovelinestext`: for formatting above line text (`alt` in gabc)
-    - `normalstafflines`: for formatting the full length staff lines
-    - `additionalstafflines`: for formatting the short lines behind notes above or below the staff.  This format defaults to inheriting changes to `normalstafflines`.
-    - `lowchoralsign`: for formatting low choral signs
-    - `highchoralsign`: for formatting high choral signs
-    - `firstsyllableinitial`: for formatting the first letter of the first syllable of a score which is not the score initial
-    - `firstsyllable`: for formatting the balance of the first syllable of the score
-    The second argument, also required, is the code necessary to turn on the formatting.  The third argument, optional and enclosed in square braces (`[` and `]`), is the code necessary to turn off the formatting (e.g. if the code to turn on the formatting contains a `\begin{environment}` then the code to turn it off must have the matching `\end{environment}`.  The third argument is optional because not all formatting commands have explicit off switches.
+- All formats can now be changed via the `\grechangeformat` command.  This replaces the mixed system of styling commands which could be redefined for some elements and specialized commands for applying formatting to others.  See GregorioRef for details.
 
 ### Added
 - With thanks to Jakub Jelínek, St. Gallen style adiastematic notation is now handled through [nabc syntax](http://gregoriochant.org/dokuwiki/doku.php/language) (see GregorioNabcRef.pdf for details and [the new example](examples/FactusEst.gabc)). Only one line above the notes is currently handled. This is a preview, backward incompatible change are possible in future releases.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
     - `initial`: for formatting normal initials
     - `biginitial`: for formatting big (2-line) initials
     - `translation`: for formatting translation text
-    - `abovelinetext`: for formatting above line text (`alt` in gabc)
+    - `abovelinestext`: for formatting above line text (`alt` in gabc)
     - `normalstafflines`: for formatting the full length staff lines
     - `additionalstafflines`: for formatting the short lines behind notes above or below the staff.  This format defaults to inheriting changes to `normalstafflines`.
     - `lowchoralsign`: for formatting low choral signs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,18 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
   - `beforechoralsignspace` is now `beforelowchoralsignspace`.
   - `lowchoralsignshift` is now `choralsigndownshift`.
   - `highchoralsignshift` is now `choralsignupshift` and its sign is now inverted.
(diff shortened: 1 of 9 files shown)


henryso:
Maybe it's for a different pull request, but I don't understand, from the changes, how you use this new feature correctly.

eroux:
ok for me

rpspringuel:
@henryso, I've added some examples to the documentation.  Let me know if that makes things clearer or if I need to make additional edits.

henryso:
I think the examples are good.  It makes it much easier to understand.

Are the styles isolated in all cases?  For example, is:
```
\grechangeformat{abovelinestext}{\bfseries}
```
good enough, or do I need to do the following:
```
\grechangeformat{abovelinestext}{\begingroup\bfseries}[\endgroup]
```
This should also be made more clear in the documentation.

One last comment: It feels like "style" would be a better word than "format" here, at least from the connotations that are coming to my mind.  What do you think?

rpspringuel:
In reverse order:

I debated the use of "format" or "style" myself because we had used both previously (e.g. it was `\greinitialformat` but `\greabovelinetextstyle`).  In the end I chose format simply because there were more elements which had previously used that word (both the initials, the translation, and both staff lines as opposed to the above line text and the choral signs).  I'm not adverse to switching to "style".  It would also be trivial to support both `\grechangeformat` and `\grechangestyle`.

The first option is sufficient because I've built the `\begingroup` and `\endgroup` into the format changing function.  It is my hope that by doing this I've made the function somewhat "idiot-proof" so that users who aren't familiar with the fact that many of the font commands have "on-switches" but no "off-switches" wouldn't end up with something unexpected.  

That said, there is nothing wrong with the second option, it will just doubly encapsulate the bold command to make sure it doesn't carry over into other elements of the score.  However, the main purpose of the third argument is to support the use of environments as in:

   \grechangeformat{abovelinestext}{\begin{bfseries}}[\end{bfseries}]

You can see an example of this sort of thing in `additionalstafflines` which is defined to inherit its settings from `normalstafflines`.

henryso:
I think it's worth a quick mention that the styles will be used in an isolated group (or whatever the right term is in TeX) in the documentation.

rpspringuel:
Is that sufficient?

henryso:
That looks fine to me.

rpspringuel:
Checking the tests there are four failures:

- fix-347 and test98: These appear to be spacing related.
-PlainTeX: These aren't typesetting, but I'm not familiar enough with this variant of TeX to determine why right away.

I have to close up for the day, but will look at these closer tomorrow if no one else gets there first.

tex/gregoriotex-main.tex
```diff
@@ -660,6 +679,7 @@
       \vskip\gre@dimen@additionalbottomspace %
       \vskip\gre@dimen@spacebeneathtext %
       \vskip\gre@dimen@currenttranslationheight %
+      \endgre@format@normalstafflines
```
henryso: Missing `%` at the end?

tex/gregoriotex-main.tex
```diff
@@ -1202,23 +1222,25 @@
 }%
 
 \def\GreSetStaffLinesFormat#1{%
-  \global\def\greadditionalstafflinesformat{#1\relax }%
-  \global\def\grenormalstafflinesformat{#1\relax }%
+  \gre@warning{\protect\GreSetStaffLinesFormat\space is deprecated.\MessageBreak Use \protect\grechangeformat{normalstafflines} instead.}
```
henryso: Missing `%` at the end?

tex/gregoriotex-signs.tex
```diff
@@ -351,13 +352,14 @@
...
   \hbox to 0pt{%
-    \greadditionalstafflinesformat %
+    \gre@format@additionalstafflines %
     \kern\gre@dimen@temp@three %
     \gre@dimen@temp@five=\wd\GreTempsign %
     \advance\gre@dimen@temp@five by \gre@dimen@additionalcustoslineswidth %
     \kern-\gre@dimen@temp@five %
     \vrule width \gre@dimen@temp@five height \gre@dimen@stafflineheight%
     \hss %
+    \endgre@format@additionalstafflines
```
henryso: Missing `%` at the end?

henryso:
fix-347 and test98 are not spacing-related.  It's actually a difference in font.  I think what's common about those two tests is that they have above-lines-text.

tex/gregoriotex.tex
```diff
@@ -44,4 +44,81 @@
 
 \input gregoriotex-main.tex
 
+%% The following two functions are defined for backwards compatibility.  They will eventually be removed.
+\def\greinitialformat#1{%
+  {\grefontofinitial #1}%
```
henryso: should be `gre@fontofinitial`

tex/gregoriotex.tex
```diff
@@ -44,4 +44,81 @@
...
+  \relax %
+}%
+
+\def\grebiginitialformat#1{%
+  {\grefontofgrebiginitial #1}%
```
henryso: should be `gre@fontofbiginitial`

tex/gregoriotex.tex
```diff
@@ -44,4 +44,81 @@
...
+  \begingroup%
+  \relax%
+}%
+\def\endgre@format@lowchoralsign{\endgroup}%
+
+\def\gre@format@highchoralsign{%
+  \begingroup%
+  \relax%
+}%
+\def\endhighlowchoralsignformat{\endgroup}%
```
henryso: should be `\def\endgre@format@highchoralsign{\endgroup}%`

henryso: Missing:
```
\def\endgre@format@firstsyllableinitial{\endgroup}%

\def\gre@format@firstsyllable{%
  \begingroup%
  \relax%
}%
\def\endgre@format@firstsyllable{\endgroup}%

\def\gre@changeformat#1#2[#3]{%
  \gdef\csname gre@format@#1\endcsname{\begingroup#2}%
  \gdef\csname endgre@format@#1\endcsname{#3\endgroup}%
}%
```

henryso:
If there's nothing more to add, then I think this is ready to merge.

CHANGELOG.md
```diff
@@ -35,6 +35,18 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
   - `lowchoralsignshift` is now `choralsigndownshift`.
   - `highchoralsignshift` is now `choralsignupshift` and its sign is now inverted.
 - `\grecoloredlines` now takes a single argument, a named color, instead of the three components of an RGB color.  As a result, `\redlines` can now use `gregoriocolor`, making the red staff lines consistent with the text, even when the user teaks `gregoriocolor`.  Addresses [#21787 on the old tracker](https://gna.org/bugs/index.php?21787).
+- All formats can now be changed via the `\grechangeformat` command.  This command takes two or three arguments.  The first argument, required, is the format to be altered.  The defined formats are:
```
eroux: Reading it again, I think the entry in the Changelog should have a length similar to the others, with a pointer to the documentation... The changelog for v 4.0 is already very big, let's not make it unreadable...

rpspringuel:
Shortened the CHANGELOG entry.

Should I add change `format` to `style` or add support for a `\grechangestyle` entry as was mentioned above?  If not, then I agree that we're ready to merge.

henryso:
I still feel like style is more appropriate than format, as style has a more extrinsic connotation and format a more intrinsic one, so I'd vote to change it, but in the end it's just stuff people have to type, and inertia seems to favor format.

Whatever we decide, I don't think it's necessary to create aliases.  One way to do it is fine.

eschwab:
Style may be the better term. For me at least style is an artistic thing while format belongs to the backend programmer.

rpspringuel:
Made that change.  No change to the tests. fix-347 and test98 still fail, but as @henryso indicated above, that's because they exhibit a bug which is fixed by this PR.

Anything else that needs to change prior to merge?

henryso:
I think this is good to merge.

henryso:
With no comments to the contrary, let's merge.

Issue #457, opened 2015-05-25
The score starts too close to the clef if initial-style is 0


henryso:
I'm not entirely sure this is a bug, but it seems to me like the first note starts too close to the clef if `initial-style` is set to 0.

Compare...
```
initial-style: 0;
%%
(c3) g(g) g(g)
```
... which produces...
![image](https://cloud.githubusercontent.com/assets/421318/7802759/cfb663a4-030a-11e5-89c3-4f8330f97403.png)
... with ...
```
initial-style:1;
%%
(c3) g(g) g(g)
```
... which produces...
![image](https://cloud.githubusercontent.com/assets/421318/7802763/f97f0b32-030a-11e5-9438-3397b5086d2d.png)

I can work around it by putting a space after the clef like so:
```
initial-style: 0;
%%
(c3 ) g(g) g(g)
```
... which produces...
![image](https://cloud.githubusercontent.com/assets/421318/7802768/25741ee4-030b-11e5-836d-ed8fe1f325be.png)

Is this a bug?  Should I try to fix it?

rpspringuel:
I'd need to check the spacing code to be sure, but it certainly looks like a bug to me, given that the punctum and the clef are touching.

eschwab:
Probably a spacing bug, but it is different behaviour than before (before when, I do not know).

henryso:
I think this has been happening since I created the individual tests from `completetest.gabc`, from the looks of the historical PDFs in gregorio-test.

eroux:
This might be due to the change that aligns notes left (using `\gre@compute@bolshift`)

henryso:
This is not due to `\gre@compute@bolshift`.  It still happens if I take that computation out.

henryso:
The culprit is this code in `\gresetinitialclef` (in `gregoriotex-signs.tex`):
```
  \ifnum\greremoveclefcount=0\relax %
    \ifnum\grelastoflinecount=2\relax % we must not type a space if there is no initial
      \gretypekey{#1}{#2}{0}{0}{#3}%
    \else %
      \gretypekey{#1}{#2}{0}{1}{#3}%
    \fi %
  \fi %
```

This explicitly says "we must not type a space if there is no initial".  Why not?

If I just call `\gretypekey{#1}{#2}{0}{1}{#3}` always, then, to no surprise, I get a nice space between the clef and the first note.

henryso:
Any objections to making that change to add the space?

eroux:
[...]

PR #463, opened 2015-05-29
fix first line when no initial

eroux:


henryso:
(From #457)

That patch appears to break your first syllable changes when I run the PopulusSion test. Everything (well, for the four lines of that chant) seems to be aligned on the first letter of the words rather than the notes.

eroux:
fixed

PR #467, opened 2015-06-01
Fixes leaking of small font size from alt text to lyrics

rpspringuel:
This fix is valid for 3.0 only. The styling environments change has already supplanted it on develop.

### DIFF
--- a/.gregorio-version
+++ b/.gregorio-version
@@ -1,4 +1,4 @@
-3.0.1
+3.0.2
 
 *** Do not modify this file. ***
 Use VersionManager.py to change the version.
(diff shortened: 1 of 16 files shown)


henryso:
When is the intended date for this hot fix?

rpspringuel:
What do you mean?

henryso:
Is the intent to release 3.0.2 immediately upon merge?

rpspringuel:
Yes, that's what I was thinking.

That way your average user who only deals with stable releases doesn't have to wait until the 4.0 release cycle to complete to see this fixed.  Said cycle won't start for another couple of weeks and then is likely to run for at least a month (if 3.0 is any indication).   That means it'll be at least 6 weeks before 4.0 goes to general release and I didn't feel that users should have to wait so long for such a simple bugfix.

henryso:
Ok, then this looks fine to me.  Merging.

PR #470, opened 2015-06-03
Deprecation of centering-scheme gabc header.

henryso:


eroux:
maybe something could be added into UPGRADE.md?

PR #472, opened 2015-06-03
Centering scheme deprecation.

henryso:
In preparation for language-based centering, I've deprecated `centering-scheme` as discussed in #383.  Currently, this generates `\englishcentering` or `\defaultcentering` (contrary to what is documented) since @rpspringuel in currently in the process of updating the TeX, but we should obviously update this before release.

When/if this is merged, I will continue to use this branch for the language-based centering.

All tests currently pass except for the dump test that uses "english" centering.


henryso:
I have more changes ready for language centering.  Should I add them to this pull request or should I create a new pull request?

eroux:
Let's make a new one

Issue #473, opened 2015-06-03
Deprecate or remove preamble headers?


henryso:
Should we remove or deprecate the following gabc headers?

- lilypond-preamble
- opustex-preamble
- musixtex-preamble

eroux:
Agreed

henryso:
Deprecate then remove in the future or just remove now (since support for these has already been dropped)?

eroux:
I think it's quite safe to remove them

rpspringuel:
I agree.  Since support for those output types is already dropped it should be safe to simply remove the associated headers.

PR #475, opened 2015-06-04
Language-specific centering

henryso:
I have pulled in a dependency on PCRE (not PCRE2).  The logic to center syllables is now based on regular expressions and can be read from a `.vwl` file that is in the gregoriotex directory in `TEXMF`.  I'm not completely sure how to document this.

Also, I'm not sure I put the dependency into `configure.ac` correctly, and need help verifying that this builds under Windows and Mac.

This change also addresses #473.

~~This probably needs some review and revision before it can be merged.~~


henryso:
Note that I purposely created this branch in the main repository so we can collaborate on it.

eroux:
What about an argument to the `gregorio` command? This would allow to specify it in TeX (and passed automatically to the compilation command) instead of gabc, I think it would be better... What do you think?

In fact, I didn't think about that, but maybe this could be done in pure lua, at compilation time? It might eat resources though...

henryso:
On your first point, the header actually identifies the language of the lyric text in the gabc file.  This is really an attribute of the gabc file itself as opposed to a styling concern.  The user can substitute whatever expressions they want for the various languages, even for Latin, overriding the in-built rules.  Besides, having an argument to the include command and passing that to gregorio would only work if compilation is driven by Lua (i.e., auto- or force-compiled) rather than done outside (never-compile), and for the case of auto-compilation would have to figure into the algorithm that figures out if a score needs to be recompiled.

I *did* think about doing it purely from (Lua)TeX after gregorio runs, but then you have the complication of dealing with styles and other weird stuff like verbatims and special characters which make it even less appealing (than even regular expressions) and more error-prone for a non-developer to attempt to create themselves.  I suppose it is possible to build a framework which passes only the string to a Lua function, but that is some pretty hefty Lua coding to maintain position.  In addition, Lua only supports UTF-8 well enough for this kind of thing from version 5.3, which is still relatively new.

eroux:
I see the logic better now, the language field is indeed a real improvement, not only for centering rules!

Styles indeed make a pure Lua solution almost impossible... but LuaTeX supports UTF8 well enough (throuh a third party library), though it's not in lua 5.3 yet.

So this seems ok!

henryso:
@rpspringuel When you have a moment, would you be able to try compiling this branch under Windows and MacOS and let me know if you run into any issues?  You will need the pcre library (version 8.x) built with UTF-32 support.

rpspringuel:
### Results on Mac

Despite already having the pcre library installed via MacPorts, I had to download and install the library from pcre.org to get things built.  It appears that the MacPorts version doesn't have UTF-32 support.  After some playing, I found that I needed to install as follows to get things working:

```
./configure --enable-pcre32 --enable-utf
make
sudo make install
```

Without `--enable-utf` I could build, but not run gregorio (I'd get an error about pcre not having UTF support).  You'll need to note these new dependencies in the appropriate spots in the documentation.

Once that was done, I could build, install, and run gregorio.  Having worked that out, I ran the battery of tests just to make sure there were no problems and had the following results:

```
Gregorio = /usr/local/bin/gregorio
GregorioTeX = /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.tex

gabc-dump/bugs/fix-20.gabc : FAIL - Files gabc-dump/bugs/fix-20.gabc.out- and gabc-dump/bugs/fix-20.dump- differ

real	10m48.767s
user	9m40.905s
sys	0m32.237s

SOMETHING FAILED
```

The dump that failed appears to be failing on centering stuff, which is probably to be expected.

I'll do the Windows testing a bit later today.

henryso:
Thanks for testing on Mac.

I wonder if there's a way to probe if the library has utf support in `configure.ac`.  The dump should be updated in gregorio-test as of about an hour ago.

henryso:
Thinking about this more, when we release binaries (to Mac/Windows/Debian), I suppose we need to ensure the dependency or link the library statically.  I wonder if there's a better way.

rpspringuel:
It would be better, I think, if we link the library statically, rather than have an external dependency.  This is especially true for Mac users who make use of the installer as they're almost certainly going to be less comfortable with the command line install of pcre.

I'm still working on the Windows testing, but I foresee similar problems for Windows users.

jakubjelinek:
It would be better if it is configure time decidable, whether to link it statically or not.
E.g. on various Linux distributions linking statically of system provided libraries goes against various packaging guidelines for various reasons (especially security).

henryso:
I don't disagree with the static linking for distribution to Mac/Windows, and it'll definitely be configurable (to @jakubjelinek's point) from the configuration script.

@rpspringuel I've committed a change that I believe removes the need for `--enable-utf` when building pcre.  Can you try this out with the MacPorts version of pcre?

jakubjelinek:
Also, not relying on a single regex library, but allowing wider choice, might help users.  E.g. using POSIX regcomp/regexec if detected (which is provided in system C library on various OSes).

henryso:
The problem with POSIX regcomp/regexp is shaky support for 32-bit characters and seems to be locale-specific.  PCRE support for 32-bit characters is stable and predictable.

eroux:
There need to be an option for static linking indeed, especially for TeXLive. See the "lltxplatform integration" thread on [the LuaTeX list](http://tug.org/pipermail/luatex/2015-June/thread.html)

rpspringuel:
The MacPorts version of pcre doesn't have 32-bit support enabled, and so fails to configure with the error `configure: error: unable to find the pcre32 library` (same behavior as before).

With pcre installed from pcre.org, the need for `--enable-utf` in order to run gregorio is removed by the latest commit.

henryso:
I will definitely need help putting the static linking option in configure.ac.  My google search on this topic so far yields confusing information.

rpspringuel:
I'm trying to build the Windows executable on Ubuntu 12.04 and am running into problems.  Despite having installed pcre from pcre.org in the same way that I did do so on my Mac, I'm getting the error about configure not being able to find the pcre32 library.  Any ideas as to what I could do differently?

henryso:
I suspect pcre is not cross-compiled for Windows on that system.  Unfortunately, I don't have any experience with cross-compiling for Windows.  I'm starting to wonder if it's best to just come up with a simple library for this code to get around library issues.  This would eliminate problems not building the libraries in the exactly right way, cross-compile problems, and issues with inclusion into TexLive.  Does anyone have any thoughts on a good approach?

eroux:
I'm using [mxe](http://mxe.cc) for another project... It's kind of gentoo-like (you compile the entire compile chain) so it's slow, but it's quite good, and pcre is packaged, maybe it's worth a try. But this makes things much more complex of course...

rpspringuel:
Hunting around on the web it seems that there might be a pcre library for mingw on Fedora.  Absent steps which make that obsolete, I'll try downloading Fedora and getting a VM of that distro up and running and try again.

> I'm starting to wonder if it's best to just come up with a simple library for this code to get around library issues. This would eliminate problems not building the libraries in the exactly right way, cross-compile problems, and issues with inclusion into TexLive. Does anyone have any thoughts on a good approach?

I can't comment on the feasibility of this approach, but from a portability stand point, having all the necessary resources built into our code base will make supporting Windows much easier.  It will also make things a bit easier for the novices on Mac, though, since I could get it working, a user with some comfort on the command line (or a willingness to learn) could be coached through it.

henryso:
One way to do get rid of the regular expression engine, which works for Latin and English is to define a set of characters which may be vowels, a set of possible prefixes (exclusive) to a string of vowels, and a set of possible suffixes (inclusive) to a string of vowels.

Language | Always Vowels | Prefixes | Suffixes
-- | -- | -- | --
Latin | a, e, i, o, u, ae, oe, y | i, u | *none*
English | a, e, i, o, u, y | y, qu | w

However, I only have two data points here, so I don't know if this is general enough a set of restrictions.

eroux:
That seems reasonable! Thinking "out loud", for French that would be something like:

* vowels, eau, ou, où, oue, oué, éé, ée, eue, au, eu, é, è, à, i, o, u
* prefixes: y, qu, i, ill
* suffixes: none

rpspringuel:
I'm afraid I don't understand the terminology of "prefix" and "suffix" in this context enough to comment intelligently.  Can you provide and explanation, with examples, of how you would use this to recognize which letters in a syllable constitute the vowel?

henryso:
The rule is basically that you match from the start of the syllable

1. Skip any non-vowels.
2. Upon finding the first vowel, see if a prefix rule applies and skip the first vowel if it is part of a prefix and is followed by another vowel
3. The vowel string starts at this point
4. Keep reading past vowels
5. If the first character(s) after the vowels is(/are) a suffix, include them
6. The vowel string stops at this point

Using English (since that's what I know):

vowels: a, e, i, o u, y
prefixes: y, qu
suffixes: w

- vent - would identify "e" as first string of vowels, no prefixes or suffixes apply
- type - would identify "y" as the first string of vowels.  Although it is a prefix, since there is no vowel following it and "y" is a vowel, "y" is identified
- yes - would identify "e" as the first string of vowels. "y" is a prefix and is followed by a vowel.
- queen - would identify "ee" as the first string of vowels. "qu" is a prefix and is followed by a vowel.
- qursh - would identify "u" as the first string of vowels.  Though "qu" is a prefix, it is not followed by a vowel.
- say - would identify "ay" as the first string of vowels, no prefixes or suffixes apply
- saw - would identify "aw" as the first string of vowels.  Though "w" is *not* a vowel, it is a suffix, so it is included.
- cool - would identify "oo" as the first string of vowels, no prefixes or suffixes apply.

I don't have an international keyboard, so I can't easily type certain characters, but for Latin, what I meant was the ae and oe ligatures.

@eroux's French example would simplify to
- vowels: a, e, i, o, u (and accented varieties)
- prefixes: qu, i (y and ill are not needed because they don't end in vowels and so are not ambiguous with them)
- suffixes: none

jakubjelinek:
But in latin i isn't always a vowel, depending on the style (j vs. i consonant).

henryso:
Right, so i is one of the prefixes in Latin

ius - identifies "u" since "i" is a prefix and is followed by a vowel.

rpspringuel:
Okay, I think I've got it.

A quick Google search for vowels in another language ([Spanish](http://www.spanishdict.com/topics/show/115)) reveals a potential problem.  "y" can be a suffix in Spanish (indicating a diphthong), but only after "a", "e", and "o" (the "strong" vowels).  I thus conclude that after the "weak" vowels ("i" and "u") "y" should always be a consonant.  How would you handle something like that?

henryso:
In Spanish, as far as I can tell, "y" is always considered as a consonant even when it sounds like a diphthong and would be part of the following syllable:

tuyo: tu - yo
mayor - ma - yor

So it wouldn't actually be a problem.  I suppose you could consider loan words, in which case, you can make is a prefix, and there still wouldn't be a problem.

rpspringuel:
The website I linked to gives "hay", "hoy", and "rey" as examples where the "y" is part of the diphthong.  Still, I have to admit that my knowledge of Spanish is limited.  I'll check with one of my confreres who is a Spanish teacher on the issue.

henryso:
Ok, that makes sense.  So make it a vowel and prefix.  It should still work.

rpspringuel:
I just tried installing the pcre build on my Mac test boxes using the installer.  In both cases the Installer failed during the post-install script.  The executable did get put into place, but calling `gregorio --version` resulted in an error because the pcre32 library couldn't be found.  Clearly, then, even when using pre-built binaries, there is still a dependency on pcre being installed before hand.

henryso:
Even more encouragement to get rid of the dependency.  The algorithm seems to work for Latin, English, Spanish, and probably French.  I guess that's good enough to start with.  It can be further tweaked as the exceptions arise.

rpspringuel:
Okay, after talking to my confrere "y" is always a consonant in modern Spanish with the following 2 exceptions:

1. "y" (the word for "and").  Here it's the complete word and is considered a vowel.
2. When it occurs after a single vowel to complete a diphthong.  Turns out this isn't just for the strong vowels, but can also occur on the weak ones (though it is much rarer).  Basically, it functions like "w" in English (i.e. as a suffix).

However, he pointed out that this is modern Spanish, which has been highly regularized by the Royal Academy.  Older forms of the language would have more irregularities.

Since I imagine that other languages would also have such irregularities, perhaps we should have some gabc notation which could be used to override the automatic vowel detection in a given syllable.  In this fashion, if the user doesn't like the way the system is working for a particular syllable, they can force something else.

I'd propose using the pipe (`|`) character for this (since brackets, parentheses, and braces are already in use), with one occurring on either side of the vowel: e.g. `s|ee|n`.

henryso:
We already have a way to do that in gabc.  Curly braces will force centering on whatever you want.

rpspringuel:
Huh... I thought those were for something else, but looking at the gabc notation pages I can clearly see that usage but not any other.

henryso:
These latest changes should eliminate the dependency on PCRE at the expense of a probably more buggy hand-coded simple DFA solution, but it does seem to work at this point against the gregorio-test project.  I still need to figure out how best to document this, especially since the `.vwl` files now have a specific syntax.

rpspringuel:
Perhaps some comments in the `.vwl` files themselves?  I would assume that most people would create such a file by starting with a copy of one of the existing files so having comments right there that explain the three lines would probably be a good start on documentation.

henryso:
That's a very good idea.  I'll do that.

henryso:
Please review and let me know if I should change anything.

henryso:
(I deleted the previous comment about review-readiness, in case someone is wondering after reading e-mails)

I just realized there is an issue with this implementation.  I am changing the value of the `language` header to lower-case for the purposes of finding the vowel file.  However, without a considerably more sophisticated solution, this only works for ASCII characters.  There is also a possibility for apostrophes and spaces (and who knows what else) in file names, which is probably something we don't want to deal with.

There are a few solutions that come to mind:
1. Stop changing the value to lower-case.  This assumes the file system is UTF8-encoded, and the possibility of strange characters in file names still exists.
2. Have multiple languages in the vowel file, which is already UTF8-encoded.  Then the system can load the rules for the selected language, matching the language name case-sensitively.
3. Have a "directory" file which maps UTF8-encoded language names with ASCII file names.
... and I'm sure there are others

I'm currently leaning towards the second, but I am open to suggestions and opinions.

eroux:
I think 1 is a bit simpler, but why not 2... Also, I realized that in French, "gu" is also a prefix, but in English too right? (guilt, guide, etc.)

henryso:
Indeed, `gu` is also a prefix.  English is also a great mish-mash of exceptions to every rule, so it's good we have the curly braces.

eroux:
Yes, English is know to be by far the most difficult language to pronounce (French certainly comes second, it has much more complex rules, but far less exceptions).

BTW, "gu" is also a prefix in Latine (guidare). "ge" is also a (rare) prefix in French, and it would be good to add ô, î, â, ê, û to the vowels (they are quite common in French). Also, maybe just adding the vowels would be enough for Vietnamese (ă, ơ, ư).

henryso:
"gu" in Latin is covered by having "u" as a prefix in Latin.

henryso:
@eroux Did you want me to add the circumflex vowels to the Latin list so that it would cover French a little better, or did you want me to create a French language section in the data file?

eroux:
I think adding circumflexes in Latine would be enough for French yes...

henryso:
I think this one is finally ready for review.  I've added the French and Vietnamese vowels to the Latin vowel set.  Please let me know if there is anything else I should change.

eroux:
ok for me!

rpspringuel:
A quick look over doesn't reveal anything that I think needs changing.  I assume that you've tested it to make sure it works, so I think it's ready to merge.

There should probably be tests for each of the supported languages added to gregorio-test, but I don't think merging this needs to wait on that.

henryso:
Yes, the tests are passing, and I will be adding a test for the English rules, probably later today.  Merging.

Issue #44, opened 2015-02-10
Set lines above initial in the gabc file.


eschwab:
The `gabc` file should be able to set the two lines above the initial.

I have been looking into adding this feature. Currently there can be two `annotation: :` fields in the `gabc` header but they are associated with `voice_info`. Rather than hijack these, perhaps rename `mode: ;` to `modelineone: ;` and add `modelinetwo: ;` For `\gresetfirstlineaboveinitial` and `\gresetsecondlineaboveinitial` respectively.

What is the intended/future use for `annotation: ;`?

eroux:
The best solution is to make annotation global (and not voice_info dependant), and use it. For nabc, there is also a annotation_nabc that will most certainly appear.

rpspringuel:
I agree that the annotation fields should be fixed.

One thing that might make things easier is a rewrite of the setannotation interface that I've been thinking about.  Right now there are two different annotation boxes, one for the first and one for the second, which are filled by separate commands.  My idea was to create a single annotation box to which lines could be added by successive calls of the same function.  In this fashion there's less to keep track of when writing information from the gabc into the gtex.

The idea came to me some months ago while working on `texrenaming` in gregoriotex.tex and I think I left myself some notes there in the form of a comment labeled TODO.

eschwab:
First there should be an improved annotation setting macro in TeX. See [this idea](https://mail.gna.org/public/gregorio-devel/2014-05/msg00027.html) from @rpspringuel 

Then however that works out, the gabc should be able to set the annotation with full formatting control i.e. `annotation:\color{red}\emph{H}.\color{black}II;`

eroux:
Apart from the fact that `\color{red}H\color{black}` is anti-pattern and should be replaced by `{\color{red}H}`, I admit I'm not particularly keen on having too much TeX in gabc... But I also admit it's just practical, and I don't think it would be reasonable (from a development resource point of view) to have some kind of markdown parsing in the gabc headers, so I don't disagree on this.

jperon:
Just a thought about that : to help one separate style and content, I think that approach should remain optional, and, if `modelineone` and `modelinetwo` are added, `mode` should remain (without any effect on the final output).

rpspringuel:
The revamp to a single annotation function that I talked about is now in #476.  This should make it much easier to implement this carry-over.

PR #476, opened 2015-06-04
Support for arbitrary number of lines in the annotation.

rpspringuel:
Annotation functions are consolidated into a single function `\greannotation` which can, through multiple calls, build up an annotation with an arbitrary number of lines.
There are now two distances associated with the annotation `annotationseparation` (a renaming of `aboveinitialseparation` for consistency) and `annotationraise`.  The first controls the distance between lines in the annotation, the second raises or lowers the annotation from it's default position (baseline of the first line aligned with the top line of the staff).  This second distance should be more transparent than the old method of measuring the height of a second argument to the annotation functions and more consistent with how other spaces are manipulated.

Three tests fail, all involving the placement of an annotation (as expected, since the new default is slightly higher than the old one).


rpspringuel:
No comments on this?

henryso:
I don't feel qualified to comment on the TeX, but the documentation looks OK.

eschwab:
I have not had the chance to test. I'll try to make time later today.

eschwab:
Sorry for the delay. Works wonderfully. This will make #44 much easier!

PR #485, opened 2015-06-11
Fix #44 Annotation Header

### LINKED ISSUE #44: Set lines above initial in the gabc file.

eschwab:
This is my first real attempt at C. Review and feedback is most welcome.

Addresses #44


src/struct.h
```diff
@@ -608,6 +608,8 @@ typedef struct source_info {
  * 
  */
 
+#define NUM_ANNOTATIONS 2
```
henryso: I would rename this `MAX_ANNOTATIONS`.

src/gabc/gabc-write.c
```diff
@@ -820,6 +813,12 @@ void gabc_write_score(FILE *f, gregorio_score *score)
     if (score->mode) {
         fprintf(f, "mode: %d;\n", score->mode);
     }
+    for (annotation_num = 0; annotation_num < NUM_ANNOTATIONS; ++annotation_num) {
+      if (score->annotation[annotation_num]) {
```
henryso: Should be indented 8 spaces, not 6.

src/gabc/gabc-write.c
```diff
@@ -820,6 +813,12 @@ void gabc_write_score(FILE *f, gregorio_score *score)
...
+	fprintf(f, "annotation: %s;\n",
```
henryso: I think this should be 12 spaces in, no tab.

henryso:
Stylistically, I think we are using 4-space indentation, no tabs, for C.

henryso:
Question: Is it possible to override the annotations in the gabc file from the tex file?

eschwab:
I think I found all the tabs, I did not have my editor configured properly.

Yes, by default the tex file overrides any annotations in the gabc. The hierarchy is:

 1 `\greannotation` in the main tex file
 2 `annotation` field in gabc file
 3 `mode` field in gabc file

If a higher level is already defined, then the lower ones are ignored.

eschwab:
This is ready for review.

src/struct.c
```diff
@@ -1037,6 +1042,11 @@ static void gregorio_free_score_infos(gregorio_score *score)
     if (score->user_notes) {
         free(score->user_notes);
     }
+    for (annotation_num = 0; annotation_num < MAX_ANNOTATIONS; ++annotation_num) {
```
henryso: Indentation appears to still be 2 spaces here.

henryso:
Besides my indentation comment, this looks fine to me.  Tests check out as well.

henryso:
Looks good. Merging.

PR #486, opened 2015-06-12
Corrected function declaration warnings.

henryso:


henryso:
Merging because this is just fixes for warnings.

PR #489, opened 2015-06-13
Fixed change log for vowel rule file name.

henryso:


eroux:
Sorry to report it here, but I think we should also add Czech and Polish vowels to latine (there are Gregorio users in these areas)... the two lessons of Polish I took are clearly not enough to know if it's ok, but I think just adding

    ůŮěĚąĄĘę

to the vowels would be enough... @jakubjelinek what do you think?

jakubjelinek:
Yeah, the first 4 should be enough to cover Czech.
Generally, for Czech I think there should be no prefixes nor suffixes,
but if the current algorithm will handle i and u words as {i} and {u}
(i.e. if it is listed in prefix, but there are no letters after it in the
syllable), then perhaps the latin vowel handling with the above
letters add could be good enough.

jakubjelinek:
BTW, not that I know anything about the languages, but I think in Hungarian you also want:
őŐűŰ
for Norwegian perhaps:
åÅøØ
etc.

eroux:
You're right, there are hungarian Gregorio users, in Norway I'm not quite sure...

jakubjelinek:
Also, forgot to mention, in Czech there are also syllabic consonants lLrR (and in Slovak those plus
ĺĹŕŔ), where if gregorio wanted to handle those, what could work perhaps if there are no vowels in a syllable and there is a syllabic consonant in there, center on the syllabic consonant.  Examples of syllables without vowels, but with syllabic consonants (used the {} syntax to highlight the intended centering):
st{r}č() p{r}st() sk{r}z() k{r}k() v{l}k()
Rarely in Czech mM also acts as syllabic consonant (e.g. {o}()s{m}() or s{e}()d{m}()).
https://en.wikipedia.org/wiki/Syllabic_consonant
mentions also Slovene, Serbo-Croatian, Macedonian.

henryso:
Let's move this discussion to #491.

PR #492, opened 2015-06-13
Added the ability to alias language rule sets to multiple names.

henryso:
The syntax is `alias [name] to [language];`.  Hopefully I documented this clearly.

Ready for review.  This also fixes a memory leak in vowel parsing.


eschwab:
Looks good.

PR #494, opened 2015-06-13
Fix problem when openout_any = p

jperon:
Since Gregorio refuses to write to `/tmp/lua_XXXX` when `openout_any = p`, it wasn't possible any more to use `\gabcsnippet` without `openout_any = a`.
As LuaLaTeX accepts to write to `/tmp/lua_XXXX` event with `openout_any = p`, writing gabc content into the tmpfile, then reading directly from gregorio process makes it possible again.

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -311,16 +311,16 @@ end
 
 local function direct_gabc(gabc, header)
   tmpname = os.tmpname()
-  local p = io.popen('gregorio -s -o '..tmpname, 'w')
+  local f = io.open(tmpname, 'w')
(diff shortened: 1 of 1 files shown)


henryso:
Looks good to me.

eroux:
There is something that doesn't make sense here: if LuaLaTeX can write to something and Gregorio not with the same kpse settings, there is something wrong with either LuaLaTeX or Gregorio... I'll try to inspect that.

henryso:
For LuaLaTeX, the file being written is a temp file.  For gregorio, it's just a filename passed at the command line.  It might have something to do with that.

Issue #491, opened 2015-06-13
Centering features


henryso:
Continuing the misplaced discussion in #489:

I am not at all familiar with any of the languages mentioned, so I will take everything said there an canon.  I can certainly add all the vowels to the in-built Latin ruleset.  However, they will have to "suffer" Latin's "i" and "u" prefixes.  So if "i" or "u" is followed by a vowel, the "i" or "u" will not be part of the center.

An alternative would be, of course, to create another language in the data file.

Supporting syllabic consonants will not be possible in the current code.  English also has syllabic consonants ("l", "r", "n", probably more), but, in English, there will be a vowel present that is not the center (i.e., c{a}t-t{l}e), so I'm not going to worry about them.  The rule for syllablic consonants proposed by @jakubjelinek is excellent, and can be handled by a second pass if no center is found.  I will add a keyword "secondary" to the vowel file to handle this case.  My reason for choosing "secondary" is that I wanted a single word, and neither "syllabic" nor "consonant" is correct here, and this rule is not general enough to cover all syllabic consonants, anyway, so I named it technically: it will be done in a second pass.

Thoughts?

eroux:
That sounds good! I don't think really know if Polish or Czech have think like ie, ue, uo, etc. I don't think so, but @jakubjelinek will know better!

jakubjelinek:
secondary keyword is fine.  Czech doesn't have ie, ue, uo (for two vowels only au or ou).
Though, do you want to add
secondary lLĺĹmMnNrRŕŔ;
for Latin too (won't that break anything for e.g. Latin itself), or just add Czech (and other Slavic languages that need this) to the vowels data file?

henryso:
Adding the secondaries to Latin will not break it, but it could conceivably break other languages that currently work fine with the Latin rules. The "safe" thing to do would be to add this to the vowel data file, but I will do whatever you tell me to do, since I don't know these languages.

eschwab:
It might be better to have a separate section in the vowal.dat for Slavic languages. If we fill the latin section with characters that don't occur in latin then people who look into the file in the future may be scratching their heads trying to figure out what is going on.

jakubjelinek:
But then it might be useful to have some syntax for language aliases.  Because there is no reason to write a separate section for Czech, another one for Slovak (perhaps the same can be useful also for other Slavic languages).  So perhaps allow:
language [Czech, Slovak];
etc.?

henryso:
[...]

PR #495, opened 2015-06-14
Added support for secondary (non-vowel) centering and Slavic languages.

henryso:
Addresses #495.

This branch is based on #492, so merging this merge changes for #492.  For the Slavic languages, I put in the letters from #489 and the suggestions from #491.

Please review.


henryso:
Added explicit Hungarian support per #495.

tex/gregorio-vowels.dat
```diff
@@ -55,7 +63,46 @@ prefix y Y;
...
+alias [Czech] to [Slavic];
+alias [slovak] to [Slovak];
+alias [Slovak] to [Slavic];
+
+language [Slavic];
+
+vowel aàáâąAÀÁÂĄ;
+vowel eèéêëěęEÈÉÊËĚĘ;
+vowel iìíîIÌÍÎ;
+vowel oòóôőOÒÓÔŐ;
```
eroux: I don't think slavic languages have ő nor ű (but I'm not 100% sure)

henryso:
@eroux Thanks, I forgot to remove the Hungarian vowels from the Slavic list. That's done now.

jakubjelinek:
There are many other vowels that aren't used in Slavic languages, but I actually think it shouldn't hurt to list them some obvious vowels not present in the languages.  E.g. if in say a slavic hymn there are foreign words (latin, greek written in latin, etc.), there could be vowels the language does not have etc.
So I actually think with some comment it wouldn't hurt even to have the slavic vowels in the latin vowel set, etc. (after all, it right now has many french ones that aren't used in latin).

henryso:
Right, this is actually the reason I left the Latin æ and œ as well as the accent and circumflex vowels in English, for loanwords and the likely use of some Latin in English text.  Should I add the Slavic vowels into the in-built Latin set?  I won't add the syllabic consonants.  People can always select the appropriate Slavic language should they need them.

jakubjelinek:
If @eroux is not against it, I'm all for that.  Even without the secondary, with the u/U with ring above and e/E with caron in the set, the latin centering would be good enough for most of the slavic words, so e.g. if somebody forgots to request slavic centering, or if in a mostly latin song there are a few slavic words, the result will be usable.

henryso:
The original request from @eroux in #489 was to add Slavic vowels to the Latin set, so I don't think he will be against it.  I'll add them.

jakubjelinek:
Haven't read the centering code yet, what does it do for latin syllable "in"? Does it treat i as prefix even when it isn't followed by vowels?

henryso:
Since the "i" in "in" is *not* followed by a vowel, then "i" is considered just a vowel (and not a prefix).  The same would be true for the "u" in "un".

henryso:
I'm satisfied with how this works.  Anyone willing to review it?

eroux:
ok for me

PR #496, opened 2015-06-15
Fixed memory leaks identified by valgrind.

henryso:
I ran gregorio through valgrind, and it found some more memory leaks, which I've fixed here.  I ran this check against my `secondary-center` branch, so merging this will pull in the changes for #492 and #495.

That said, this is ready for review.


henryso:
These are mechanical fixes, so I'll merge them.

PR #500, opened 2015-06-15
Integrate hour within debian-git version

jperon:
Without that, it's sometimes impossible to build two different commits in
the same day.


henryso:
Looks OK to me from its context, but I don't use debian.

PR #502, opened 2015-06-16
Add package option to make deprecated commands raise an error.

eschwab:
Added a new package option `deprecated=false` that makes deprecated macro raise an error instead of a warning.

Ready for review.


rpspringuel:
I think the definition of `\gre@deprecated should actually go inside `gregoriotex-main.tex`.  This way it should work for both LaTeX and PlainTeX.  This is also why I favor keeping the conditional inside the definition.  If the conditional were outside the definition the you would have to define the boolean value yourself before loading the package in PlainTeX so that the conditional can be evaluated as the package loads. As it has been done, someone using PlainTeX doesn't have to do that and we can set a default value for the boolean.

henryso:
Makes sense.

eschwab:
I still need to pay closer attention to the division between PlainTeX and LaTeX, and the associated files. And yes the bool was defined after `\gre@deprecated` so the if needed to go inside the macro. One less thing to loose track of (a manual define of the bool).

rpspringuel:
With the change in place I would expect all tests to pass.  If that is true , then I would say this is good to merge.

henryso:
Tests pass.  Merging.

Issue #487, opened 2015-06-12
Use the style system for \gre@writemode.


henryso:
Rather than hard-coding `\GreSmallCaps` and `\GreBold` for `\gre@writemode`, use a new style.

eschwab:
Are you thinking to wrap all annotations (`\greannotation`) with this style or just `\gre@writemode`? Personally I would not wrap all annotations. Doing `\gre@writemode` would be fine, since I think the amount of cases where a user wants the `mode` header to output is fairly slim.

henryso:
Just \gre@writemode.  I should have named the issue differently.

PR #503, opened 2015-06-17
Make a new formatting style for `\gre@writemode`.

eschwab:
Adds the new style environment `modeline`.
Addresses #487


rpspringuel:
Everything looks like it should work as expected, but I can't test on my iPod.

henryso:
This works for the default setting, but there is a bug in #462 (fixed by #504) that prevents the user from setting a custom style.  Since this works fine with #504 applied, I will merge this.

PR #504, opened 2015-06-17
Corrected the call to @ifnextchar.

henryso:
Related to #462.

`\@ifnextchar` was being called incorrectly by `\grechangestyle`.


henryso:
Please review and merge this if it is satisfactory.

eschwab:
This is good. `\ifnextchar` has a different syntax than `\if`

Issue #505, opened 2015-06-17
error: unknown type name ‘rulefile_parse_status’


jperon:
Here is what I get while compiling last develop : 

~~~
make[3]: Entering directory '/root/greg/gregorio-dev/src'
gcc -std=gnu99 -DHAVE_CONFIG_H -DBRANCH_VERSION='"develop-59821c7-2217"' -I.  -I../src -I../src/gabc  -I../src/dump  -I../src/gregoriotex -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2   -g -O2 -fstack-protector-strong -fPIE -Wformat -Werror=format-security -Wstrict-prototypes -Wall -Wextra -MT vowel/gregorio-vowel-rules-l.o -MD -MP -MF vowel/.deps/gregorio-vowel-rules-l.Tpo -c -o vowel/gregorio-vowel-rules-l.o `test -f 'vowel/vowel-rules-l.c' || echo './'`vowel/vowel-rules-l.c
In file included from vowel/vowel-rules.l:29:0:
vowel/vowel-rules-y.h:82:87: error: unknown type name ‘rulefile_parse_status’
 int gregorio_vowel_rulefile_parse (const char *const filename, const char **language, rulefile_parse_status *const status);
~~~

jperon:
Fixed by #506

PR #506, opened 2015-06-17
Fix for unknown type name ‘rulefile_parse_status’.

henryso:
Possible fix for #505.

@jperon Please try out this fix for #505 and merge it if it works for you.  I don't get that error, so maybe it's a differing version of bison.


jperon:
Thank you !

Issue #441, opened 2015-05-21
unique id of score


eroux:
It would be necessary for #59 (and possible other future features) to have a unique identifier of the score, to write some info in the .aux file. I thought about a few things:

- using lua md5 library (packaged in LuaTeX) to get a md5 hash on the gabc file. The problem is that if you include the same score twice, you'll get twice the same id, which is not always what you want (if you include it in different `\hsize` formats. There could be a check when computing the md5 hash, to know if there was already one though...
- make Gregorio write a random ID in the `.gtex` file, that would change at each compilation. This would be a bit simpler, but wouldn't work if the score is included with the "force compile" option. But that would be more efficient, and we can make #59 not compatible with the force option

What do you think?

jperon:
What about using first option, but adding an optional argument (`id` for example) to \includescore, that would "salt" the md5 ? It would then be possible to have #59 compatible with the force option, while letting users with more advanced needs change `\hsize`.

eroux:
that seems a good idea, thanx!

jakubjelinek:
Wouldn't just a simple counter, bumped at the end of each includescore, be enough for this?
Those should be the same between two consecutive lualatex invocations on the same source.

eroux:
yes, but this would mean you would have to recompute the infos at each pass, whereas with this solution you don't have to... That may look like misplaced optimization, but I'm not sure about the performance penalty of the pass gathering information on scores...

rpspringuel:
@eroux, can you explain in a bit more detail what you mean?  Wouldn't you have to recompute the md5 hash each run in order to match score specific variables in the aux file to their respective scores?

jakubjelinek:
You could use that counter as kind of array index when caching that computation in the *.aux file and again when reading it back from there.  So pretty much the same as the md5/sha or any other hash, or uuid, just it wouldn't be a constant you could put directly into the *.gtex file, but instead you'd use \the\some_counter or what's the exact syntax to query the counter value.

eroux:
No, wouldn't have to recompute the informations on the lowest and highest notes for each line of the score

eroux:
@jakubjelinek : suppose the user compiles a doc with score A then score B, we have score A with counter == 1, score B with counter == 2. Now users inserts score C between scores A and B, we have counter == 1 for A, counter == 2 for C, counter == 3 for B, so all spaces are mixed up... That means each compilation must recompile all infos of each score in order to be consistent...

[...]

PR #515, opened 2015-06-24
Rough implementation of per-line spacing adjustment.

henryso:


eroux:
That seems good to me! A few things missing:

- conditional to make it happen or not
- maybe better glyph IDs, the reason is that if a user inserts a score or adds a glyph in a score, with the current implementation, it might cause height changes in the middle of a line of another score, and this will look really wrong... I think the idea is to make things as good as possible at each run, though 2 runs would be necessary to get the final result
- maybe a message telling the user that Gregorio had to change some height (if it did), and that the score might look weird, but it should be ok after a second compilation

A few tests to make:

- include the same score twice at two different places (won't matter with current glyph ID implementation, but might with another
- performance: did you notice a change?

henryso:
Addressing your points in order.

- I will work on conditional once I'm satisfied with the general implementation.  I had assumed that this would not be switchable, so this means fewer things will need to be removed as dead code.  I'm glad you cleared that up before I got too far.
- Regarding glyph IDs, attribute values are 32-bit integers, so I suppose we'll need an indirection table.  I'll have to think about this one a bit more.
- There is already a message generated by `gregoriotex.lua` in `write_greaux`, unless you mean you want to change the wording of this message (currently the message is `Line heights may have changed. Rerun to fix heights.`)
- I will test including the score in multiple places
- I didn't notice a performance change on a single pass, but I was only using small test scores and not testing it explicitly.  The big performance killer is the necessary second pass.

Thanks for reviewing.  I'll work on your suggestions and be back with a new cut in a few days.

eroux:
the message is ok, you can leave it like this. In real life, most people won't have to compile twice all the time, only when they see some ugly things, so the most important is not to make one compilation considerably slower... Maybe there could be a mode where the data is only read, not written? This way if someone just works on some text and not on the scores (in a big document such as a book), he can turn off most of this processing?

henryso:
How should this read-only mode be activated? TeX command? Environment variable?

eroux:
I don't think most users will know how to handle environment variables, so I had a TeX command in mind... I'm affraid this won't be used a lot, but for very big projects it will certainly matter.

henryso:
My only "concern," if you can even call it that, is that a TeX switch means you need to change your sources to make it work, but if that's standard practice in the TeX world for this kind of thing, then there's no problem.

eroux:
This is of course something only for "experts" working on very large documents, so they should know what they are doing... See for instance the "draft" mode of LaTeX...

henryso:
Please see the changes in 710f5db1d45f6f5ee2228bca183aa42915af6fb6 and let me know if this is a good solution to the glyph id problem.

eroux:
it will do yes... I tend to like my initial idea of having a random ID per score, allowing insertion of a score between two other scores without destroying all the spacings afterwards, but this is already significantly better than the previous solution! Maybe a `\greRandomId{xxx}` in the `.gtex` file would be enough, with just a small mechanism to add `.1`, `.2`, etc. to the ID if the score is included several times... What do you think?

henryso:
If I seed the random number generator by `time(NULL)`, there's a very good chance for collision if all the gabc files are compiled at once, up front.  Additionally, this causes a problem (as noted in #441) with force-compiled scores.

I pulled in the GPL SHA-1 digest code from mbed TLS and hooked it into the score determination lexer.  This way, I generate an SHA-1 digest while parsing the gabc file and use it as the unique score ID.

Please see #517 for these changes.  Does this suffice for score ID?

eroux:
I think so, thanks a lot!

PR #517, opened 2015-06-25
Generated an SHA-1 digest score ID for indexing computed score line height data.

henryso:


jakubjelinek:
I must say I'm a little bit surprised by the choice of sha1 source, I'd have picked the one from GNU coreutils instead, that one is already GPLv3, is very portable and used widely.

henryso:
Frankly, the mbed TLS SHA-1 source code came up first on a Google search.  Should I switch?

jakubjelinek:
Guess depends on how much work it would be.  I'd trust the GNU version more.

henryso:
Looking at the source code, I'd have to figure out how to properly set `WORDS_BIGENDIAN`.  Other than that, it shouldn't be too difficult.

jakubjelinek:
Best using standard autoconf macro AC_C_BIGENDIAN .

PR #518, opened 2015-06-25
Switched SHA-1 to the GNU coreutils implementation.

henryso:
In response to @jakubjelinek in #517.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,6 +95,7 @@ auto/
 #gregorio specific
 doc/_minted-*
 fonts/*.ttf
+*.greaux
 *.gz
 *.bz2
 *.zip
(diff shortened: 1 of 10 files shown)


jakubjelinek:
I'm worrying about the stdalign.h include, that is a C11 header and thus fairly recent thing (e.g. in GCC only added in 4.7 in November 2011).  Coreutils provides its own stdalign.h if system does not have one.  So, I wonder either if you shouldn't add stdalign.h to AC_CHECK_HEADERS in autoconf and do something like:
{{{
#ifdef HAVE_STDALIGN_H
#include <stdalign.h>
#else
#define alignof(x) sizeof(x)
#endif
}}}
Also, in unicode.h you already use
{{{
#ifdef HAVE_STDINT_H
#include <stdint.h>
#else
#include <inttypes.h>
#endif
}}}
supposedly the #include <stdint.h> should be replaced with that even in sha*

henryso:
Ok, will do.  Thanks.

PR #523, opened 2015-06-27
Implemented variable line height computation within a score.

### LINKED ISSUE #59: Update each line's vertical spacing

henryso:
Fixes #59 and #441.

Please review and let me know if there's anything I missed.  I believe this is ready for merge.


henryso:
@eroux The substantive differences since you last reviewed are f04e49a, d317009, 2e39da2.

UPGRADE.md
```diff
@@ -113,6 +113,28 @@ Full TeX markup is also accepted:
...
+
+This new behavior requires two passes (two runs of lualatex) in order to do the calculation correctly.  On the first pass, lualatex will warn you with the message
+
+    Module gregoriotex warning: Line heights may have changed. Rerun to fix heights.
+
+As you edit your document, lualatex will only show this warning if an extra run is required.
+
+Alternately, you can use something like latexmk to automate this activity.  Instructions for using latexmk are outside the scope of this document.  Please consult the latexmk documentation for more information on using it.
+
+It should also be noted that this new feature, besides requiring two passes, also imposes a small performance penalty on each pass to do the additional computation.  This is negligible for small documents, but may have more noticeable impact as documents get larger.
```
eroux: Did you manage to measure something?

eroux:
Apart from a few minor remarks, everything seems fine, an excellent job, thanks a lot!

UPGRADE.md
henryso: The individual difference in time for a single pass was within the margin for error, but the average time was slightly higher (within two seconds for the 1000x PopulusSion).  The biggest impact is that a second pass is necessary.  The computation of SHA-1 also has a slight impact on score compilation. but this one is within milliseconds, which, given the size of gabc files, is not really surprising.

eroux: Great! I admit I expected the slowdown to be much bigger, this is a very good news!

henryso: We are piggybacking on the existing loop, so the big-O complexity (in Lua) hasn't changed.  There are only a few comparison and table manipulation operations within the loop, so the result doesn't really surprise me.  Reading/writing the file .greaux could slow down the operation variably depending on the operating system, but if the OS caches writes, it won't be noticeable, especially given the size of the file.

henryso:
Merging.

PR #530, opened 2015-06-29
Added the ability to generate LilyPond-like textedit links.

### LINKED ISSUE #528: Support generation of lilypond-like textedit links
This idea comes from http://www.mail-archive.com/gregorio-users%40gna.org/msg02769.html

The format of the URL is `textedit://{filename}:{line}:{char}:{column}`.  Line and column appear to be 1-based and char is the character-offset from the beginning given line, 0-based.  Column expands tabs.

Links would be created for score glyphs and for syllable text.

This feature would be disabled by default.  You would need to enable it both when calling `gregorio` to generate additional information in the `gtex` files (because this "leaks" filenames into the gtex file) as well as in the TeX code (which is possible to drive strictly based on whether the additional information is in the `gtex` file, but I like the idea of toggling it from TeX as well).

Obviously, it would be disabled if `gregorio` is reading from stdin since it can't know the filename.

https://www.tug.org/pipermail/pdftex/2006-March/006475.html shows how to generate links without LaTeX and hyperref.

henryso:
Fixes #528.


henryso:
There are probably some obscure cases where this gets weird, but this seems to be working for the most part.

Please review.

henryso:
Is it OK to include this in 4.0?  Anyone willing to review?

eroux:
Ok for me!

eschwab:
I can get links in the pdf but cannot easily test that they are accurate. Good for me.

henryso:
I double-checked the accuracy of the column numbers and found a few bugs:

- discrepancy with tabs in the source, compared to LilyPond
- incorrect computation of the character offset (the second number), compared to LilyPond
- missing point-and-click links on bar syllable text
- incorrect zero character offset on syllable text links

The numbers should be more-or-less correct now.  There is a slight weirdness against the text on the first syllable of the score due to when Bison calls the macros to update the counters, but since this is a debugging aid, I don't think it's much of a problem to find the first syllable text.

I think this is good to merge.

Issue #535, opened 2015-07-03
Consider drawing braces using metapost or tikz


henryso:
We currently draw braces by stretching out glyphs in the font.  This is fine as long as the aspect ratio is fairly close to the original, but when it's stretched far, the glyph begins to look funny, being thicker horizontally than vertically.  This is true also when the glyph is compressed, but less obvious since this condition is less likely and less noticeable given the original aspect ratio.

This is a low priority item stemming from a thought in #454, but since that issue is closed, I'm opening this to track it.

eroux:
Note that apart from basic Metapost learning, this is quite straightforward with [luamplib](https://github.com/lualatex/luamplib)

henryso:
TeX Live comes with cmarrows, which looks like it produces pretty good-looking (curly) braces.  Is it too much of a dependency to use that?  We'd still have to implement round braces.

eroux:
Why not, I admit I didn't know the existence of metapost packages, but I think it's a good way to go!

eroux:
(well, I admit I find cmarrows curly braces good-looking for math, but I find them too flat for music)

eroux:
There's a better example page 9 of [metaflow manual](http://www.utdallas.edu/~hamlen/metaflow.pdf)

jakubjelinek:
Yeah, the current gregorioTeX ones are nice if they aren't too stretched.  Would be nice to get similarly looking ones for the non-stretched case.

henryso:
The metaflow braces do look much nicer for music.  However, metaflow does not appear to be part of TeX Live, and metaflow is licensed under the LPPL, which, as far as I know, is not compatible with the GPL.

eroux:
It was mainly an example, I'm sure it's quite easy to draw it by hand...

PR #536, opened 2015-07-05
Added MetaPost rendering of braces.

henryso:
Implements #535.

While testing, I found and fixed an issue with multiple scores inclusions and variably-sized braces.  The fix uses a `\directlua` call to generate the `\luatexlatelua` command with the score identifier carved in so that it will be correct on ship out.

This is ready for review and critique, especially on the look of the braces as they change size.  They should be very close to the greciliae figures when at their design widths but change in various, non-linear ways as they stretch out in an attempt to keep horizontal line weight more-or-less the same throughout the range of widths.


eroux:
Patch looks good, can you post a picture showing the curly braces at different widths?

henryso:
image

henryso:
Generating that image caused me to think further about the tuning of compressed round braces (not much I can do for curly braces, though).  Here are the salient parts of the image (still at 300 dpi):
image

eroux:
Ok for me to merge!

PR #539, opened 2015-07-05
Develop

rpspringuel:
This is a version number update for the new release.

Since the patch implemented in the release is already in develop (in a different form), the only visible changes should be to the version number.

Please note that I've added a comment to `gregoriotex.tex` which states the version number (said file is already on the list of `VersionManager.py` because it didn't have anything declaring the version.  I'm not sure if a comment is the best way to do this but it's all I have time to do right now.

### DIFF
--- a/tex/gregoriotex.sty
+++ b/tex/gregoriotex.sty
@@ -19,8 +19,8 @@
 
 \NeedsTeXFormat{LaTeX2e}%
 \ProvidesPackage{gregoriotex}%
-    [2015/06/01 v3.0.2 GregorioTeX system.]% PARSE_VERSION_DATE_LTX
-\RequirePackage{xcolor}%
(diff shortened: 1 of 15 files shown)


tex/gregoriotex.sty
```diff
@@ -19,8 +19,8 @@
 
 \NeedsTeXFormat{LaTeX2e}%
 \ProvidesPackage{gregoriotex}%
-    [2015/06/01 v3.0.2 GregorioTeX system.]% PARSE_VERSION_DATE_LTX
-\RequirePackage{xcolor}%
+    [2015/07/01 v3.0.3 GregorioTeX system.]% PARSE_VERSION_DATE_LTX
+\RequirePackage{xcolor}
```
eschwab: Missing `%`

henryso:
I'll merge this and fix the issues in develop.

PR #542, opened 2015-07-09
Added support for Dominican-style neumes.

henryso:
Implements #1.

This is a bit rough, but it works.  I still have a lot of testing to do before this can be merged, but please advise me if you notice anything I should do here.


eroux:
Seems good to me, but I'm in no position to judge properly...

henryso:
I've run into an issue while testing. A note at `d` typically doesn't have a long queue (unlike most notes which are on lines), but with the `dV>` figure in its Dominican form with inverted stem, it runs into the line.  I think I can do something with a new TeX command for this particular case, but if anyone has any other ideas, I'm all ears.

henryso:
As far as this change goes, I think this is ready for a final review/merge.  If someone familiar with Dominican glyph usage comes forward, we can address their concerns at that time.

eroux:
ok for me!

henryso:
Hearing nothing to the contrary, I'll merge this.

PR #543, opened 2015-07-11
Improve styled barred symbols (LaTeX-only)

henryso:
From @jakubjelinek on gregorio-devel:

> I've yesterday noticed a problem with \Vbar, \Rbar and \Abar when used in \textit{} contexts looks ugly (the bar character does not have an italic version).  Is it possible to find out from inside of these macros whether it is used in italic context and use more slanted version of the bar and/or have separate adjustments for non-italic and italic?

I've taken the ideas I presented in response to that email and developed them a bit, hence this pull request.

A few notes:

1. No changelog entry because this is a new feature in 4.0, but I did update `UPGRADE.md` to make mention of it
2. Unlike how we do `\f@size`, I couldn't figure out a way to implement `\f@shape` and `\f@series` outside LaTeX, so this new feature is basically LaTeX-only.
3. The reason I defined it in `gregoriotex-symbols.tex` in a `gre@iflatex` block rather than in `gregoriotex.sty`, is because this code is needed by both `gregoriotex.sty` and `gregoriosyms.sty`, and I didn't want to repeat all of it in both files.  Let me know if I should do this in a different way.

Please review and let me know if I should change anything or if there is better TeX to handle this kind of thing.


doc/Command_Index_internal.tex
```diff
@@ -325,6 +325,26 @@ \section{Gregorio\TeX{} Controls}
   \#3 & \MP{} commands & \MP{} commands to draw the brace outline.
 \end{argtable}
 
+\macroname{\textbackslash gre@iflatex}{\#1}{gregoriotex.sty \textup{and} gregoriotex.tex}
+Evaluates to \verb=#1= if running under \LaTeX{}.
+
+\begin{argtable}
+  \#1 & \TeX{} code & the width of the brace.
```
eschwab: There probably is not a brace involved here...

eschwab:
Works for me.

PR #544, opened 2015-07-14
Minor tweaks to changelog and upgrade.

eschwab:
- Also found two deprecated macros to convert to the new deprecated
  system.

- Revert `GreDagger` to `gredagger` since it does not appear in `gregoriotex-write.c`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,20 +20,19 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Lines are now aligned on the leftmost note instead of the leftmost letter.
 - Some glyph shapes are improved a little in greciliae, it should be noticeable especially for porrectus.
 - The `O` modifier in gabc now has expanded uses beyond the salicus `(egOi)`.  A stemmed oriscus will appear on a lone pitch `(gO)` or a followed by a lower pitch `(gOe)` (see [#76](https://github.com/gregorio-project/gregorio/issues/76)).  A virga strata will appear on the second note of two ascending pitches `(giO)`.
-	- With thanks to *Abbazia Mater Ecclesiae (IT)* for drawing the new shapes, the strophicus, liquescent strophicus, liquescent oriscus, and liquescent punctum inclinatum in greciliae have changed to better differentiate them from other shapes.  Use
-
-	\grechangeglyph{Stropha}{greciliae}{.caeciliae}
-	\grechangeglyph{StrophaAucta}{greciliae}{.caeciliae}
-	\grechangeglyph{OriscusDeminutus}{greciliae}{.caeciliae}
-	\grechangeglyph{PunctumInclinatumAuctus}{greciliae}{.caeciliae}
-
-	if you prefer the old shapes.
+- With thanks to *Abbazia Mater Ecclesiae (IT)* for drawing the new shapes, the strophicus, liquescent strophicus, liquescent oriscus, and liquescent punctum inclinatum in greciliae have changed to better differentiate them from other shapes. If you prefer the old shapes use:
+```
+    \grechangeglyph{Stropha}{greciliae}{.caeciliae}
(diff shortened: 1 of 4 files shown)


henryso:
GreDagger *does* appear in gregoriotex-write.c, twice.

eschwab:
Yes it does. Looks like I was grepping `gabc-write` instead. Since `GreDagger` can be both a gregorio and user level macro, would it be useful  to define both `GreDagger` and `gredagger` similar to how we have `GreStar` and `gresixstar`?

henryso:
The difference is that `\GreStar` is an alias for `\gresixstar`, which has an alternate, `\greheightstar`.  So you would issue `\let\GreStar\greheightstar` to select the alternate, whereas the same is not true of `\GreDagger`, which is the "one-and-only" dagger figure.  Of course the waters are very muddy here because of the new alternates system.

It really doesn't matter to me one way or the other, so if people are in favor of defining `\gredagger` and aliasing `\GreDagger` to it, I don't have any objection.

CHANGELOG.md
```diff
@@ -32,7 +42,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
   - It is now possible to suppress the automatic horizontal episema "bridging" across spaces within a syllable.  To use it, append `2` to the `_` on the note before the space you do not want bridged (see [#72](https://github.com/gregorio-project/gregorio/issues/72)).
   - A horizontal episema will no longer overlap with a vertical episema on the same note (see [#395](https://github.com/gregorio-project/gregorio/issues/395)).
-- The dot after a torculus whose last note is on a line will now appear below the line.  If you do not like the automatic placement of a dot after any note on a line (not just the torculus), you may specify a `0` or `1` after the `.` to force the dot to appear below or above the line, respectively.  See [#386](https://github.com/gregorio-project/gregorio/issues/386).
+- The dot after a torculus whose last note is on a line will now appear below the line (see [#386](https://github.com/gregorio-project/gregorio/issues/386).)  If you do not like the automatic placement of a dot after any note on a line (not just the torculus), you may specify a `0` or `1` after the `.` to force the dot to appear below or above the line, respectively.
```
henryso: The period should come after the right parenthesis.

tex/gregoriotex.sty
```diff
@@ -125,12 +125,12 @@
 
 
 \def\grecoloredlines#1#2#3{%
-  \gre@warning{\protect\grecoloredlines\space is deprecated.\MessageBreak Use \protect\gresetlinecolor\space instead.}%
+  \gre@deprecated{\protect\grecoloredlines}{\protect\gresetlinecolor{red}}%
```
henryso: I'm not sure if "red" should be mentioned here.  The command is not exactly equivalent in any case, so perhaps just leave the argument out, like the original warning?  I can't think of a good, terse way to do this here.

henryso:
Besides the last comment, I'm fine with this pull request being merged.

tex/gregoriotex.sty
```diff
@@ -123,14 +123,13 @@
...
-
 \def\grecoloredlines#1#2#3{%
-  \gre@warning{\protect\grecoloredlines\space is deprecated.\MessageBreak Use \protect\gresetlinecolor\space instead.}%
+  \gre@deprecated{\protect\grecoloredlines}{\protect\gresetlinecolor}%
   \gre@changestyle{normalstafflines}{\color[RGB]{#1,#2,#3}}%
 }
 
 \def\greredlines{%
-  \gre@warning{\protect\greredlines\space is deprecated.\MessageBreak Use \protect\gresetlinecolor{gregoriocolor}\space instead.}%
+  \gre@deprecated{\protect\greredlines}{\protect\gresetlinecolor{gregoriocolor}%
```
henryso: Sorry for being piecemeal, but I didn't notice it before.  The braces are unbalanced in #2.  Same below.

henryso:
Looks good.

PR #545, opened 2015-07-15
Allowed programs other than cp to be used for copying.

henryso:
This is an improvement primarily for helping gregorio-test use rsync instead of cp to avoid unnecessary copies.

### DIFF
--- a/install-gtex.sh
+++ b/install-gtex.sh
@@ -135,7 +135,7 @@ function install_to {
     dir="$1"
     shift
     mkdir -p "$dir" || die
-    cp "$@" "$dir" || die
+    $CP "$@" "$dir" || die
 }
 
 echo "Installing in '${TEXMFROOT}'."


install-gtex.sh
```diff
@@ -135,7 +135,7 @@ function install_to {
     dir="$1"
     shift
     mkdir -p "$dir" || die
-    cp "$@" "$dir" || die
+    $CP "$@" "$dir" || die
```
eschwab: This may be a bashism I don't know, but dosen't `CP` need to be defined?

henryso:
It does, and it is, if you look at the whole file.

install-gtex.sh
henryso: It does, and it is, if you look at the whole file. It just wasn't used before.

eschwab: Ok, I just was not expecting there to be a previously unused variable there.

eschwab:
Fine for me.

PR #546, opened 2015-07-26
Added punctum cavum inclinatum glyphs and support.

henryso:
As mentioned on gregorio-devel, this adds support for punctum cavum inclinatum figures (hollow diamonds), which can be used in a more distinguishable form of trigonus without changing the overall shape.  I also added the augmented form.  These are accessed via gabc `(Gr)` and `(Gr<)`.

Implementation was a little more complex than I thought because of the punctum inclinatum positioning on descent, but I think I got everything.

This is ready for review.


henryso:
The tests pass, but I will commit a modified test which includes the new shape once this is merged.

eschwab:
Looks great to me. Thank you for adding this.

Issue #551, opened 2015-08-04
TeXLive integration


eroux:
Putting here everything for TeXLive integration into one issue.

- https://mail.gna.org/public/gregorio-devel/2015-08/msg00005.html
- https://mail.gna.org/public/gregorio-devel/2015-08/msg00004.html

Also, as `fonts/Makefile` is supposed to go in the TDS, `install-gtex.sh` is not usable without running the autotools, `./configure`, etc. which makes no sense in the context of the source tarball. So I think the best solution is to remove `Makefile` from `install-gtex.sh` if it's ok for you.

eroux:
Also, `greg-book.engine` should be executable. @rpspringuel could you also convert the different `.engine` script to posix `sh`? This would allow them to be included in TeXLive directly, thus avoiding user configuration.

henryso:
@eroux When you say "remove `Makefile` it from `install-gtex.sh`", do you mean to stop `install-gtex.sh` from installing `Makefile` or to replace `install-gtex.sh` with `Makefile` functionality?

eroux:
The first option

henryso:
The patches look fine on initial review.  I will apply them tonight or tomorrow night if someone else doesn't beat me to it.

rpspringuel:
Both of the `.engine` files are obsolete, so there is no point in fixing them for TeXLive integration.  We'd be better off moving up the removal time table and just deleting them.

rpspringuel:
Actually, checking the changelog, they should be scheduled for removal in 4.0.0 since they were deprecated in 3.0.0.  I'll submit a pull request to that effect.

rpspringuel:
On the issue of upgrading flex, based on what I can find, flex 2.5.39 won't build on a Mac.  I've seen a couple of different proposed patches (one on the [MacPorts site](https://trac.macports.org/ticket/46378) and one on the [flex site](http://sourceforge.net/p/flex/bugs/182/)), but as of yet, no fix has been implemented.  As a result, I can't upgrade flex on my Mac at the moment (I'm currently running 2.5.37 there, the most recent version which builds on a Mac).

However, I should be able to upgrade it on my Ubuntu VM.  I'll see what I can do about that.

rpspringuel:
Okay, after some further research, I discovered that Homebrew does have flex 2.5.39 available in it's system and so I've been able to install it on my Mac.  I still have to work out the details of forcing the use of that version of flex rather than the MacPorts version.

[...]

PR #554, opened 2015-08-05
Fixes for TexLive integration

henryso:
Addresses some of the items in #551:

I applied patches 1, 2, 4, and 5 from Peter Breitenlohner.  Some hunks were left out and all of patch 3 was left out for (flex/bison-generated) files not checked into git.  These changes should pass once flex 2.5.39 is used to generate these files for the distribution archive.

I also took Peter's advice and removed the `const` modifier from pointers which are freed to remove the warnings about casting away const.  Note that Peter is wrong about language never being changed, so the `free` in `character.c` is, in fact, necessary.

Finally, I implemented @eroux's suggestion to not install the Makefile into the TDS.

Tests pass.  This is ready for review and merge.


eroux:
Thanx!

PR #555, opened 2015-08-05
Fixed self-intersection and wrong-direction errors.

henryso:
I upgraded fontforge to version 20150612 and got the `Internal Error (overlap) in {some character}: The end of the monotonic does not match that of the containing spline.` error when running `squarize.py` that has been mentioned by @eschwab and @jakubjelinek on the mailing list.

In trying to find a fix, I discovered the fontforge validation feature, which flagged multiple self-intersecting and wrong-direction glyphs.  This pull request fixes those errors.

The validator also flagged missing points at extrema, but I don't want to "fix" that until I understand the impact of the fix.

Unfortunately, these things don't seem to make any appreciable dent into the "monotonic...containing spline" errors.

That said, I do think this somewhat improves the quality of the fonts.  The generated fonts still have some remaining self-intersection and wrong-direction problems, so something else is additionally at play here.

Using the older version of fontforge, this change causes minor test failures, which I expected given earlier experience.  However the results are visually fine.

It's also interesting to note that, using the newer version of fontforge, there are many more test failures.  While still visually fine, the detected differences are at the join points of the glyphs combined by `squarize.py` (which may or may not be a coincidental thing).

I think this change is ready for review/merge.  I probably won't have much time to do more investigation into the other font issues until at least this weekend.

### DIFF
--- a/fonts/greciliae-base.sfd
+++ b/fonts/greciliae-base.sfd
@@ -19,7 +19,7 @@ OS2Version: 0
 OS2_WeightWidthSlopeOnly: 0
 OS2_UseTypoMetrics: 1
 CreationTime: 1176402534
-ModificationTime: 1437872559
+ModificationTime: 1438737067
 OS2TypoAscent: 0
 OS2TypoAOffset: 1
 OS2TypoDescent: 0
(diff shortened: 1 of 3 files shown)


eroux:
In my experience this kind of errors might be due to pasting glyphs with a common edge but no overlap, this seems to trouble fontforge... shifting it 0.1 unit may fix it (but the errors you experience might be different).

henryso:
Based on a spot check I did yesterday, at least some of these glyphs definitely have an overlap.  I suspect that this overlap may be troubling some new spline handling code in fontforge.  But that's really not much more than a guess.

PR #556, opened 2015-08-05
Fixed spaces after abbreviations.

henryso:
An additional fix for #551.

I also found a bug in the `Makefile.am` which was not finding not-installed fonts for building the PDFs.  This is also fixed.

This is ready for review/merge.


eroux:
Thanx a lot!

Issue #558, opened 2015-08-09
_3 does not correctly align the horizontal episemus in some cases


asbloomf:
for example `(eg/F'_E_3/fv_)`

Really, it seems it extends too far in this case all the time, for example `(eg/F'_E_/fv)` still seems to extend to far to the right.

From Henry So Jr ( https://mail.gna.org/public/gregorio-users/2015-08/msg00011.html ): The problem is that the algorithm is not taking into account the "kerning" done to handle descending puncta inclinata.

henryso:
More information: I was wrong about the cause.  This is actually related to bridging.  `(eg/F'_2E_3/fv_)` gives the correct result.

henryso:
I understand the problem a bit better, but I don't have a good solution in mind yet.

Bridging simply adds a fixed-sized horizontal episema extension to the current horizontal episema.  This extension happens to be longer than the shortened punctum inclinatum "Hepisemus" glyph on the second punctum inclinatum.  The problem is that this same extension is not long enough when there is a greater distance to the second punctum inclinatum.  To wit:

![image](https://cloud.githubusercontent.com/assets/421318/9155448/84e91c64-3e87-11e5-8bf8-243a2b7df55d.png)

As I think through the solution, I'll take any suggestions anyone has to offer.

PR #559, opened 2015-08-09
Added argument to GreHEpisemusBridge to handle adjacent puncta inclinata.

henryso:
This is my first attempt at fixing #558.

It works well enough, and tests pass, except that the overlap of the horizontal episemas over puncta inclinata is understandibly different and causes the "diff" image to catch something which is not visible when printing (and which may be visible on screen, depending on PDF viewer).

However, I am completely open if anyone has a better way of doing this.

Barring that, I think this is ready for review and merge.


eroux:
Fine for me. There is not obvious solution, I think this one is quite good!.

PR #560, opened 2015-08-14
More TeXLive-related fixes.

henryso:
More fixes for #551 as per recent messages posted to gregorio-devel, with contributions from Akira Kakuto and Peter Breitenlohner.

I broke these out into individual commits.

Peter's explanation of Akira's changes map as follows:
1. e1e4459cf580920c7b7b462577cd5ad8aa1e37a2
2. already done
3. 501eef9c07e33960bfa7a976107c33f402ad0276
4. already done
5. c0ece89a664928f5ac46d5535959d0b7ff65b9d3
6. 802764f5a616b2f84756c5e914ffcb3384416e2a
7. 50663290032ccd2f616995e2180514b6d90b72a6
8. ad131d989bd8397f74b23bd4ec27e4c32cc0ed90
9. c0ece89a664928f5ac46d5535959d0b7ff65b9d3
10. 9579a37899f419b151f3782ccfd4dc1b13bb8f52

This compiles without warnings under both gcc and clang, and all tests pass.

Please let me know if I've missed anything or if you'd rather I do something differently.

This is ready for review and merge if there are no issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,8 @@ Makefile
 config.status
 lib/gregorio.pc
 src/gregorio
+src/utf8strings.h
+src/encode_utf8strings
 
 #emacs auto gen files
 auto/
--- a/configure.ac
+++ b/configure.ac
@@ -28,16 +28,17 @@ AC_CONFIG_MACRO_DIR([m4])
 
 AM_INIT_AUTOMAKE([-Wall -Wno-portability subdir-objects foreign dist-bzip2 no-dist-gzip])
 
-AC_PROG_CC
+AC_PROG_CC_C89
 dnl AM_PROG_CC_C_O is deprecated since Automake 1.14, to be removed in the future
 AM_PROG_CC_C_O
-AC_PROG_CC_STDC
 AC_PROG_CPP
 AM_PROG_LEX
 AC_PROG_YACC
 AC_CHECK_TOOL([RC], [windres], [no])
 AM_CONDITIONAL([HAVE_RC], [test x$RC != xno])
 
+AX_CHECK_COMPILE_FLAG([-std=gnu89], [CFLAGS+=" -std=gnu89"])
+AX_CHECK_COMPILE_FLAG([-pedantic-errors], [CFLAGS+=" -pedantic-errors"])
 AX_CHECK_COMPILE_FLAG([-fstack-protector-strong], [CFLAGS+=" -fstack-protector-strong"])
 AX_CHECK_COMPILE_FLAG([-fPIE], [CFLAGS+=" -fPIE"])
 AX_CHECK_COMPILE_FLAG([-Wformat], [CFLAGS+=" -Wformat"])
--- a/configure.ac
+++ b/configure.ac
@@ -53,6 +53,8 @@ AX_CHECK_LINK_FLAG([-pie], [LDFLAGS+=" -pie"])
 
 AC_HEADER_STDC
 
+AC_CHECK_HEADER_STDBOOL
+
 # linux has integer types in stdint.h, solaris, vms in inttypes.h
 AC_CHECK_HEADERS([stdint.h])
 AC_CHECK_HEADERS([stdalign.h])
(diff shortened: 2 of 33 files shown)


henryso:
Note: I (obviously?) didn't test this in VS2010, so the conditionally compiled code is based on faith in Akira Kakuto's patch and some studying of documentation on the MSDN.  I think we should roll out another beta and ask Akira to try building it again.

eroux:
Before merging, I know this is quite some additional work, but would you consider using gnulibs for [stdbool](https://www.gnu.org/software/gnulib/manual/html_node/stdbool_002eh.html) and [snprintf](https://www.gnu.org/software/gnulib/manual/html_node/snprintf.html)? This seems a cleaner solution to me... I don't know if gnulib files would go into the git or only in the distributed tarballs with something in the install script (there's a [discussion about that](https://www.gnu.org/software/gnulib/manual/html_node/VCS-Issues.html)... What do you think? I've wanted to use gnulibs for quite a long time, but now that we have actual portability issues, I think the time is right...

henryso:
I quote from https://www.gnu.org/software/gnulib/manual/html_node/Target-Platforms.html#Target-Platforms :

Native Windows, with MSVC as compiler, is rarely tested and low priority.

Based on that statement, it doesn't sound like a good idea, at least not for this.

henryso:
I wish I had VS2010 available to verify if it works.  That would reduce my fears about the rarely tested low priority support of that platform.

eroux:
Also, I just realized that gnulibs hooks into configure, which is not used by VS users, so it's not really relevant...

BTW, another mail (that doesn't change much):

```
Hi Élie, Akira,

correction. This should be completely independent from the use of Kpathsea.
If "wb" is undesirable for non-TeXLive builds, perhaps

+#ifdef TEXLIVE
+        output_file = fopen(output_file_name, "wb");
+#else
         output_file = fopen(output_file_name, "w");
+#endif

with -DTEXLIVE when compiling for TeXLive/W32TeX.

Regards
Peter
```

@jakubjelinek do you have some thoughts about all this?

jakubjelinek:
b is a nop on Unix-like OSes, it only matters for Windows, which I have only aging experience of (stopped using it 21 years ago).  But making this dependent on TEXLIVE define is really strange, either you want it always, or never.

henryso:
@eroux I assume you've checked that adding a library (via gnulib) will not prevent us from entering TeXLive?

eroux:
gnulibs does not add libraries, just additional source code files (it's quite an original system)

henryso:
As far as I can tell, it creates a library in a subdirectory (default = `lib`) of your project that builds library that you then link into your source.  Or maybe this is a use-case I am not aware of?

eroux:
I don't see how you could turn "stdbool.h" into a library...

henryso:
"stdbool" would indeed only add the header, but "snprintf" would add a library module.

eroux:
Sure, but that would be linked statically, like if it was part of gregorio's source code, so it doesn't matter at all... unless I'm missing something? But anyway, as I said, gnulibs can only be used through the autotools, which are obviously not supported in VS2010, so there's no point in adding gnulibs...

eroux:
BTW, "library module" is misleading: this would add `snprintf.c` and `snprintf.h` that would be integrated in gregorio's build process.

eroux:
So, as gnulibs doesn't solve the problem, if you have no objection I'll merge

henryso:
Autotools is not so much a problem as is running configure.  From what I can tell, the TeX Live folks seem to be starting with the distribution tarball.  Assuming Akira Kakuto does not run configure (a shell script) on Windows, they must be creating a config_.h manually and pulling source files into a Visual Studio project.  In addition, even though it is "rarely tested" and "low priority", MSVC *is* on the list of gnulib-supported targets, so I think it's worth a try since linking a library statically will not be an issue.  Do you know how TeX Live programmers pull code into VS2010 for building?

eroux:
I've sent a mail to you and TeXLive builders... I won't have access to the Internet until monday morning, but do not hesitate to take actions on this without my opinion!

henryso:
Akira has responded and asked that we not change too much from what we have in the 4.0.0 beta if possible, and that they are not particularly familiar with gnulib.  That said, I think what we have here should work, so perhaps we should merge this.  When Akira gets a new beta, I would guess that he'll find any problems and let us know how he fixed them.

henryso:
Seeing as there are no objections, I'll merge this.

PR #565, opened 2015-08-18
Add `\gre@obsolete`

rpspringuel:
Similar to `\gre@deprecated` this function makes the obsolescence messages consistent and makes sure they always raise an error.  It also makes moving down the removal path easier as a function which is moved from deprecated to obsolete now simply needs `\gre@deprecated` to be replaced by `\gre@obsolete` (and a removal of the bypass code, if necessary).
All functions which currently had individual obsolescence messages have been adapted to use the new function.

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -60,6 +60,12 @@
   \relax%
 }%
 
+\def\gre@obsolete#1#2{%
+  \greerror{#1\space is obsolete.\MessageBreak Use #2\space instead}%
+  \relax%
+}%
+
+
 \def\gre@nothing{}
 
 \AtBeginDocument{\IfStrEq{\gre@debug}{}{}{\typeout{GregorioTeX is in debug mode}\typeout{\gre@debug\space messages will be printed to the log.}}}%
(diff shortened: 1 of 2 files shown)


henryso:
Is the plan to leave the "obsolete" messages in forever, or do we intend to remove them after some number of releases?

rpspringuel:
They'll get removed at the next major or minor release.

A reminder of the deprecation process:

1. Deprecated function names raise a warning which points to the new name, but otherwise work as follows.
2. Obsoleted function names no longer work, but raise an error which points to the new name.
3. The function name no longer exists in the code base.  Anyone still using it gets the generic "Undefined control sequence" error.

Each of these steps should be in a major or minor release, with step 1 or 2 being a mandatory major (i.e., if step 1 is a minor release, then step 2 must be major, if step 1 is a major release, then step 2 is allowed to be a minor release).

In this scheme, anything in step 1 uses `\gre@deprecated`.  Those in step 2 use `\gre@obsolete`.

henryso:
Once these details are all worked out, I think it would be good to add this to the wiki.

henryso:
Seems fine.

rpspringuel:
Added to the wiki: https://github.com/gregorio-project/gregorio/wiki/Deprecation-Process

PR #566, opened 2015-08-19
Refactoring old styles deprecation check

rpspringuel:
I've been shown a better way (which doesn't require the use of `\protected@edef`) to do the check for the use of the old style syntax (See [TeX StackExchange] (http://tex.stackexchange.com/questions/261830/testing-for-macro-change/261834#261834)).  I implement that here and remove the definition for `\protected@edef`.
I also pulled the definitions of `\greinitialformat` and `\grebiginitialformat` out of gregoriotex.tex and gregoriotex.sty and put them in gregoriotex-main.tex since they are identical.
Fixed some consistency problems with when styles were being applied relative to the checks (all styles which existed in the old system should now be applied within the checks)
Added checks for `additionalstafflines`, which was lacking the backwards compatibility code completely (despite being a legacy style)

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -330,8 +330,7 @@
   \gre@dimen@temp@five=-\gre@dimen@textlower %
   % if it is a big initial we print it on the second line
   \ifnum\grebiginitial=0\relax %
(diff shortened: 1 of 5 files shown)


henryso:
Seems fine.

PR #571, opened 2015-08-20
Considered automatic custos when setting the next custos.

### LINKED ISSUE #569: Incorrect auto-custos on clef change at the end of a line
The automatic custos between the two pictured lines was generated with `(z0::c3)`.  Because of the clef change, the custos should be showing on the top line, but it is not.

image

The generated GregorioTeX code is correct:
```
\GreDiscretionary{%
\GreSyllable{}{}{}{1}{\GreSetNextSyllable{}{}{}}{}{0}{}{%
\GreInDivisioFinalis{}%
}%
%
}{%
\GreSyllable{}{}{}{1}{\GreSetNextSyllable{}{}{}}{}{0}{}{%
\GreEndOfElement{1}{1}%
\GreCusto{12}%
\GreInDivisioFinalis{}%
\GreChangeClef{c}{3}{0}{3}%
}%
%
}%
```

Notably, the `\GreCusto{12}` does indicate that the custos should be on the top line.  It would appear that the "localrightcontent" supplants this setting with one based on the physical position of the next note.

henryso:
Considered automatic custos when setting the next custos.
Recomputed the next custos if the clef change discretionary is printed.
Fixed the manual custos by reusing the automatic custos code.
Fixes #569.

This is ready for review and merge if acceptable.

Several dump/gtex tests (which use the explicit custos features in various ways) fail, but the new results are appropriate.  Because the manual custos is now rendered at the element (rather than glyph) level, there is additional space before it, but this is consistent with automatic custos.  This causes the `divisio_finalis` output test to fail in an acceptable manner.  All other tests pass.  Once this is merged, I will add a test for these fixes themselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+### Fixed
+- Corrected the rendering of explicit automatic and manual custos at the end of lines when the clef change that follows it is pushed to the next line (see [#569](https://github.com/gregorio-project/gregorio/issues/569)).
 
 
 ## [4.0.0-beta] - 2015-08-01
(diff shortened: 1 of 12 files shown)


henryso:
A message on the mailing list reminds me that there are some other cases I need to test.  I'll work on them tomorrow.

henryso:
This is *now* ready for review and merge.  I've updated the description to reflect the changes.

eroux:
Ok !

PR #573, opened 2015-08-21
Texrenaming Phase III

rpspringuel:
This completes the renaming and documenting of all the macros to be consistent with the standards we discussed in #383.

The only exception are the names of the fonts.  These are currently a bit of a hodge-podge: `\gregoriofont`, `\gregoriostylefont`, `\gre@nabcfont`, `\gre@fontofbiginitial` (PlainTeX only), `\gre@fontofinitial` (PlainTeX only), and some whose names are less immediately clear because they are defined by Lua based on the value of some string variable.  I haven't done anything with these because of just how much of a mess changes here could make (especially for the ones defined by Lua).  My thoughts are to name them in the pattern `\gre@font@...` but I'm also willing to leave well enough alone if that's what others would prefer.

These changes require some changes to the tests in order to get them all to pass.  I'll post a matching pull request on gregorio-test which indicates those changes.


doc/Command_Index_User.tex
```diff
@@ -565,28 +568,40 @@ \subsubsection{Glyph Alteration}
   \#5 & dimension & vertical shift of the bar glyph.\\
 \end{argtable}
 
+\macroname{\textbackslash greseperator}{\{\#1\}\{\#2\}}{gregoriotex-symbols.tex}
```
henryso: Misspelled; should be greseparator (the spelling is correct in the description below).

doc/Command_Index_User.tex
```diff
@@ -1099,6 +1126,108 @@ \subsection{Distances}\label{distances}
...
+Space between elements in ancient notation.
+
+Default: \unit[0.06927]{cm} plus \unit[0.00182]{cm} minus \unit[0.00363]{cm}
+
+\macroname{nabclargerspace}{}{gsp-default.tex}
+Larger space between elements in ancient notation.
+
+Default: \unit[0.10938]{cm} plus \unit[0.01822]{cm} minus \unit[0.00911]{cm}
+
+\macroname{clivisalignmentmin}{}{gsp-defualt.tex}
```
henryso: Misspelled; should be gsp-default.tex

doc/Command_Index_User.tex
```diff
@@ -1099,6 +1126,108 @@ \subsection{Distances}\label{distances}
...
+
+Default: 10001
+
+\macroname{\textbackslash grenewlinepenalty}{}{gsp-default.tex}
+Penalty to force a line break.
+
+Default: -10001
+
+\macroname{\textbackslash grelooseness}{}{gsp-default.tex}
+The looseness within a score.
```
henryso: What exactly is "looseness"?

doc/Command_Index_User.tex
```diff
@@ -1099,6 +1126,108 @@ \subsection{Distances}\label{distances}
...
+
+Default: \verb=\looseness=
+
+\macroname{\textbackslash gretolerance}{}{gsp-default.tex}
+The tolerance within a score.
```
henryso: What exactly is "tolerance"?  Tolerance to what?

doc/Command_Index_User.tex
```diff
@@ -1099,6 +1126,108 @@ \subsection{Distances}\label{distances}
...
+
+Default: 9000
+
+\macroname{\textbackslash grepretolerance}{}{gsp-default.tex}
+The pretolerance within a score.
```
henryso: What exactly is "pretolerance"?

doc/Command_Index_internal.tex
```diff
@@ -40,24 +40,27 @@ \section{Gregorio\TeX{} Controls}
...
+  \#1 & string & name of the obsolete macro\\
+  \#2 & string & name of the correct macro to use\\
+\end{argtable}
+
 \macroname{\textbackslash gre@loadgregoriofont}{}{gregoriotex-main.tex}
 Loads the chosen font for the neumes at the correct size.
 
 \macroname{\textbackslash gre@calculate@constantglyphraise}{}{gregoriotex-spaces.tex}
-%\verb=\gre@calculateconstantglyphraise=%tex.tex
 Macro to caluclate \verb=\gre@constantglyphraise=
```
henryso: I know you didn't put this line in but caluclate is misspelled.  Should be calculate.

doc/Command_Index_internal.tex
```diff
@@ -345,6 +353,400 @@ \section{Gregorio\TeX{} Controls}
   \#4 & dimension   & the amount to shift the bar to the left from the end of the base symbol.
 \end{argtable}
 
+\macroname{\textbackslash gre@additionalbottomcustolineend}{}{gregoriotex-signs.tex}
+Macro to place a bottom custo with an additional line (positions \texttt{a} and \texttt{b}) at the end of a line.
```
henryso: "Custo" in all these statements should really be "custos".  "Custos" is a Latin word meaning guard.  No declension of "custos" is "custo".

doc/Command_Index_internal.tex
```diff
@@ -345,6 +353,400 @@ \section{Gregorio\TeX{} Controls}
...
+  \#3 &  0 & this is for the current syllable\\
+  & 1 & this is for the next syllable\\
+\end{argtable}
+
+\macroname{\textbackslash gre@hepisorline}{\#1\#2\#3\#4\#5}{gregoriotex-signs.tex}
+Macro to typeset a horizontal line (either an additional staff line or an episemus).
+
+\begin{argtable}
+  \#1 & character & The letter of the height of the episemus (not the height of the note it corresponds to).\\
+  \#2 & 0 & go back to the beginning of the previous glyph; this starts the         episemus at the beginning of the previous glyph\\
```
henryso: Why are there a bunch of spaces between "the" and "episemus", or is that a tab?

doc/Command_Index_internal.tex
```diff
@@ -378,45 +780,415 @@ \section{Gregorio\TeX{} Controls}
...
+Alias for \verb=\luatexlocalleftbox=.
+
+\macroname{\textbackslash gre@localrightbox}{}{gregoriotex-main.tex}
+Alias for \verb=\luatexlocalrightbox=.
+
+\macroname{\textbackslash gre@resizebox}{}{gregoriotex-main.tex}
+Alias for \verb=\resizebox=.
+
+\subsection{Auxiliary File}
+Gregorio\TeX\ creates it's own auxiliary file (extension \texttt{gaux}) which it uses to store information between successive typesetting runs.  This allows for such features as the dynamic interline spacing.  The following functions are used to interact with that auxiliary file.
```
henryso: `it's` should be `its`

doc/Command_Index_internal.tex
```diff
@@ -378,45 +780,415 @@ \section{Gregorio\TeX{} Controls}
...
+The vertical episemus character.
+
+\macroname{\textbackslash gre@fontchar@brace}{}{gregoriotex-signs.tex}
+The (rounded) brace character.
+
+\macroname{\textbackslash gre@fontchar@curlybrace}{}{gregoriotex-signs.tex}
+The curly brace character.
+
+\macroname{\textbackslash gre@fontchar@custobottomlong}{}{gregoriotex-chars.tex}
+The custo character with a long upwards directed vigra.
```
henryso: Again, this and the next few should say "custos" instead of "custo".

doc_check.sh
```diff
@@ -0,0 +1,114 @@
...
+# This script is used to compare the code base to the documentation and make
+# sure that the two match up.
+# It does this by creating two lists: one of all the macros defined in the TeX
+# code, and one of all the macros documented in the documentation (plus all the
+# macros in the code which are marked as deprecated or obsolete, and thus don't
+# need documentation.  It then compares these two lists and outputs res.tex
+# which contains a list of undocumented macros (preceded by >) and macros which
+# are documented but which no longer appear in the code (preceded by <).
+
+# The scirpt is not perfect, and especially has trouble with classes of macros
```
henryso: Not particularly important, but "scirpt" should be "script".

tex/gregoriotex-symbols.tex
```diff
@@ -182,7 +182,7 @@
 %lines
 %#1 is the type of line (1-5)
 %#2 is the number of pt of the font it will use
-\def\gresep#1#2{%
+\def\greseperator#1#2{%
```
henryso: Should be `\greseparator`.

tex/gregoriotex-symbols.tex
```diff
@@ -191,3 +191,8 @@
   \or\greLineFive{#2}%
   \fi%
 }%
+
+\def\gresep#1#2{%
+  \gre@deprecated{\protect\gresep}{\protect\greseperator}%
+  \greseperator{#1}{#2}%
```
henryso: Should be `\greseparator`.

henryso:
I'm done with my review.  Comments are inline.

doc/Command_Index_User.tex
eroux: I think it would be way too complex to explain here, but I admit I don't know which website or doc it should point to... Maybe https://en.wikibooks.org/wiki/TeX/pretolerance ?

doc/Command_Index_User.tex
eroux: (previous comment should have gone here). The idea of having a difference between version of LuaTeX is due to [this bug](http://tracker.luatex.org/view.php?id=842) (pointed in the comments of gsp-defaults). The idea is that we use discretionaries (explicit hyphens, though more than hyphens in our case) for clef changes, and we need to give them a special penalty, which is not taken into account if `pretolerance` is > -1 on LuaTeX < 0.80. The reason is exposed [here](http://tug.org/pipermail/luatex/2013-July/004516.html). But I think the doc is already complex enough...

doc/Command_Index_User.tex
henryso: OK, so looseness, tolerance, and pretolerance are TeX jargon (or maybe publishing jargon), so perhaps say "The \TeX{} tolerance (or loosenesss or pretolerance) within a score." for these three lines.

doc/Command_Index_User.tex
henryso: Actually, I think your explanation here is very good, and it would be fitting to put that explanation, with links, in gsp-default.tex where the bug 842 comment is mentioned, and have something in this document that says "See gsp-default.tex for the reason this is needed".

rpspringuel:
Edited.  Thanks for the proofreading.

henryso:
Something is causing extra space in the `plain-tex/PopulusSion/PopulusSion.tex` test.  I'm not sure what it is; maybe the pretolerance value?  I don't see any obvious extra spaces from the diff.

eroux:
I doubt pretolerance can have an effect except on a clef change... (but who knows)

henryso:
Maybe it's a manifestation of that bug, but with line 55 of `gsp-default` being:
```
\ifnum\the\luatexversion < 80\relax %
```
the result is:
image

If, on the other hand, I change line 55 back to
```
\ifnum\the\luatexversion < 78\relax %
```
the result is:
image

I happen to be on LuaTeX 0.79.1

If the new result (the first one) is correct, then fine.  I just want to be sure this is acceptable.

rpspringuel:
The rule about which pretolerance level to use was in the code in two different places with two different checks as to when to apply it (<78 and <80).  When I removed the duplication, I preserved the <78 value because that was the one that was ruling in the original code (given the order of the duplication).  After reading the links that @eroux posted, it appeared that the fix to LuaTeX first appeared in version 0.80, so I changed the rule in my latest update.  I can easily change it back, if desired.

henryso:
Aesthetically speaking, either result looks fine to my eyes.  Since the bug report states that it is fixed in 0.80, then I say that the new result is the acceptable result.  Merging.

eroux:
Theoretically, the first one is indeed the good one... But that's really weird... I honestly see no reason why there would be such a space difference here... Anyway, 0.80 fixed quite a few bugs (some on dimension computations that may have an impact), I recommend using it

henryso:
@eroux @rpspringuel New information has come to light. See https://github.com/gregorio-project/gregorio-test/commit/a25effb45d3654222e26558624d291f671516a2a. Suggestions?

PR #579, opened 2015-08-24
Texrenamingiii

rpspringuel:
This renames the fonts, and thus should complete the texnaming process.


henryso:
I'm not sure I like "neumebase". There's more stuff in there than just neumes. Though in the end, it doesn't make much difference to the user, I'd be happier with "score" (i.e., `\gre@font@score`). What do you think?

rpspringuel:
I realize there's more than just the neumes, but I don't like "score" because that to me would include stuff like the lyrics, which most certainly aren't typeset with this font.  What about "music"?

henryso:
I would be happy with "music".

henryso:
Looks good.

PR #580, opened 2015-08-25
Changing which versions of LuaTeX the pretolerance fix applies to

rpspringuel:


henryso:
We've discussed this one enough.  Merging.

PR #581, opened 2015-08-25
Feature/illumination

rpspringuel:
This is a feature which has been playing around in my head for a couple of months, but which I kept putting off as I was busy working on the TeX renaming.  Since that is now essentially complete, I had the time to implement this and it turned out to be fairly simple.


henryso:
Neat feature.

PR #583, opened 2015-08-25
Distinguishing end of line shifts from beginning of line shifts

rpspringuel:
End of line shifts (`eolshift`) are for allowing the lyric text to stretch under the custos
Beginning of line shifts (`bolshift`) are to align the lines on the notes rather than the text
The code previously conflated these two and used the same flag to determine whether to apply each one.  I've distinguished the two here, giving them each their own flag (both default to true) and user command for changing the behavior.


henryso:
All current tests pass, but I don't think there's a test for this specifically.  I'll merge; would you provide a test when you are able?

rpspringuel:
I'll see what I can come up with.

PR #586, opened 2015-08-31
Allowed the first word to be styled from TeX.

henryso:
Similar to the styling of first syllable and first syllable initial, this allows styling of the whole first word.

All of the gabc->gtex and gabc->dump tests fail because of the new code, but that is expected.  All PDF tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+### Added
+- The first word of the score is now passed to a macro that allow it to be styled from TeX.  The first word is passed to `\GreFirstWord#1` and is styled by changing the `firstword` style.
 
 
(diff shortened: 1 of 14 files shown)


jakubjelinek:
gregoriotex-syllble.tex is a typo?

henryso:
Yes.  Fixed.

eschwab:
If I try `\grechangestyle{firstword}{\color{red}}` the text changes accordingly but hyphens remain black. Should the hyphens change color or not? I'm not sure what would look best.

henryso:
Interesting. I think the best thing here would be to have another style for the first word hyphens.  That way the user can choose.  Now I just need to figure out how those work so I can make the style work there.

henryso:
On second thought, having a separate style would make things very confusing if you decide to force hyphens by typing them into the gabc.  I guess I'll just apply the same style to the hyphens, once I figure out where that is.

henryso:
3cfff63 applies the `firstword` style to the hyphens.  I had to rejig the `GreSyllable` macro in order to free up some arguments, but I think I've refactored everything as necessary.

eschwab:
...and that works perfectly. Thank you for this.

PR #590, opened 2015-09-02
Latest fixes for TeX Live integration.

henryso:
Fixes issues found by Peter Breitenlohner and Akira Kakuto from the TeX Live team.  For #551.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -71,15 +71,15 @@ EXTRA_DIST = encode_utf8strings.c utf8strings.h.in utf8strings.h \
 					vowel/vowel-rules.l \
 					vowel/vowel-rules-l.h vowel/vowel-rules-l.c \
 					vowel/vowel-rules.y \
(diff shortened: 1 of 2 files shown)


henryso:
Ok, now gregorio-resources.o is dependent on both gregorio-resources.rc and gregorio.ico, and both are in the tarball.  I think this is done.

henryso:
I left the PIC warning unfixed.

eroux:
Thanks a lot! Do you think the warnings of MSVC (const argument different between definitions) are worth fixing (I have no strong opinion about that)?

henryso:
The differences reflect interface vs implementation details, which is why GCC (or clang for that matter) doesn't flag these things.  It shouldn't matter to the caller if I modify the variable containing a value passed (prevented by `char *const`), but it does matter if I modify something that has a side effect (prevented by `const char *`).  Therefore, my opinion is to leave them alone and ask Akira Kakuto to use the `-wd4028` option.

eroux:
ok!

Issue #589, opened 2015-09-01
New parameter `initialraise`


jperon:
When the initial style was defined by `\greinitialformat`, I used to type `\raisebox{HEIGHT}{#1}` to adjust height (which is required with some fonts, for example EB Garamond Initials). As it isn't possible any more with `\grechangestyle{initial}`, I've found a workaround, but it seems quite dirty : 

```
\let\oldGreSetInitial\GreSetInitial
\def\GreSetInitial#1{\oldGreSetInitial{\raisebox{HEIGHT}{#1}}}
```

Would it be possible to have a `initialraise` parameter, so that one might type `\grechangestyle{initialraise}{HEIGHT}` ?

rpspringuel:
Something like that should be easy to implement, but I'd implement it as a distance and so it would be changed with `\grechangedim{initialraise}{HEIGHT}{1}`.  Would that be acceptable?

henryso:
This is slightly off-topic, but it seems "fragile" constructs like raisebox are not possible using the new styling system.  What other limitations are there?  I was trying to use `\MakeUppercase` in style and was trying stuff like `\grechangestyle{...}{\MakeUppercase\begingroup}[\endgroup]`, which didn't work, but was able to use `\grechangestyle{...}{\MakeUppercase}` instead, which takes advantage of the internal behavior of placing a `\begingroup` after the start-style and `\endgroup` before the end-style.  This seems fragile from a future-proofing point-of-view (it would break if the internal behavior is changed in the future).  How would someone do something like this correctly with the style system?

rpspringuel:
I think the best way of doing that would be with [the answer to this question](http://tex.stackexchange.com/questions/100336/towards-a-ucshape-and-textuc-command-for-uppercase-text-xetex-luatex) which implements a `\ucshape` command using Lua.  Then you'd simply need `\grechangestyle{...}{\ucshape}`.

Do you think it would be worth incorporating the `\ucshape` command into GregorioTeX to make it more readily available to users?

henryso:
I don't want to stray too far from the topic of this issue, so to answer your question briefly, I think `\ucshape` probably strays too far from the subject matter of GregorioTeX to be included.

That said, the actual point of my question was not how to do uppercase per se (though I very much thank you for that link and will use that command myself), but rather if the style system is flexible enough for people to use (perhaps third-party) styles that are macro-based (i.e., `\textit{...}`) but not switch-based (i.e., `\itshape`).

[...]

PR #591, opened 2015-09-03
Added the ability to manually raise or lower the initial

rpspringuel:
Addresses #589
I also flipped the sign on the dimension that gets used to determine the raise which positions the initial so that positive is up and negative is down


henryso:
Seems fine.

PR #593, opened 2015-09-03
Emptied the annotation box at the end of the score.

### LINKED ISSUE #592: Mode annotation should be suppressed when initial-style is 0
From Sr. M. Ruth OSB (Abbazia Mater Ecclesiae):

If a .gabc file with initial-style:0; has the “mode” field filled with something, and this file is included without commenting the "mode" field, that string appears over the first initial letter of the following score.

henryso:
Per @rpspringuel; fixes #592.

As this is a fix for a 4.0 feature, I did not add it to `CHANGELOG.md` since I am assuming all the -beta and -rc change logs will be sublimated into the 4.0.0 change log upon release.

I tested this and it works.  Ready for review/mege.  I will add the new test once this is merged.

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -1078,6 +1078,7 @@
   \gre@skip@temp@two=0pt%
   \gre@skip@temp@three=0pt%
   \gre@skip@temp@four=0pt%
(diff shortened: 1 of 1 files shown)


eschwab:
Indeed it does work.

Issue #396, opened 2015-05-10
<eu> block and finalis


jperon:
#371 gives `<eu>E(i) u(i) o(i) u(h) a(h) e</eu>(fe..) (::)` as the good syntax for `<eu>` blocks ; but in some rare cases, this gives bad output where the final bar is on a new line ; the bug isn't present when not closing the `<eu>` block at the end of the piece.
![euouae](https://cloud.githubusercontent.com/assets/4775231/7553558/f1d372a0-f6fe-11e4-85e7-666e4f5754a5.png)

*mwe.tex* : 

    \documentclass[a3paper,42pt,DIV=17]{scrartcl}
    
    \usepackage[autocompile]{gregoriotex}
    \usepackage{libertine}
    \pagestyle{empty}
    \setgrefactor{40}
    
    \begin{document}
        \includescore{test}
    \end{document}

*test.gabc* : 

    name:Cantémus Dómino;
    initial-style:1;
    %%
    (c3) CAN(df~)té(fef)mus(d) Dó(fh)mi(f)no :(ff) *(;)
    glo(h)ri(g)ó(hih)se(fe) e(hhh)nim(h.) (,) ma(hf)gní(hji)fi(h_g)cá(hji)tus(h_g) est.(hhh!/fgf___.) (:)
    
    For(f)ti(ef)tú(d!e'f)do(f) me(fefd)a,(d.) (,) et(d) laus(efe) me(f!g'h)a(hie'!f) Dó(fe)mi(h)nus,(hif___.) (;)
    et(f) fa(hh)ctus(h) est(hi) mi(igh)hi(gf..) (,) in(d) sa(ef)lú(fhF'EfvED'e)tem.(ed..) (::)
    
    <i><c>T.P.</c></i>
    Al(df~)le(fef)lú(hf!/hhh)ia,(hiHF.) (,) al(ef)le(gxfhe___!fw!gvFE)lú(de!fvED'e)ia.(ed..) (::)
    
    <i><c>Ps.</c></i> Can(d)tá(f)te(h) Dó(h)mi(h)no(h) cán(h)ti(h)cum(i) no(i)vum :(h.) *(:)
    qui(f)a(h) mi(h)ra(h)bí(i)li(g)a(g) fe(h)cit.(f.) (::)
    
    Gló(d)ri(f)a(h) Pa(h)tri.(h) (::) <eu> E(h) u(h) o(i) u(g) a(h) e.</eu>(fefvEDe.) (::)

[...]

PR #594, opened 2015-09-04
Fixing End of Line Shifts

rpspringuel:
After considerable experimentation, I discovered that part of the problem was that the documentation for end of line shifts was backwards: they were not needed to allow lyrics to stretch under the custos at the end of the line (that was happening even though they were never being applied under the old code), but rather to **prevent** it.

They also needed to be called a bit earlier in the stream so that they were applied at the end of the line being ended rather than at the beginning of the next line (where they get swallowed by TeX's space ignoring rules).

This PR moves the call so that they have an effect and fixes the documentation to explain what they are for.

### DIFF
--- a/.gregorio-version
+++ b/.gregorio-version
@@ -1,4 +1,4 @@
-4.0.0-beta
+4.0.0-beta2
 
 *** Do not modify this file. ***
 Use VersionManager.py to change the version.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
-
+### Fixed
+- End of line shifts now have an effect on the score.  When activated they prevent lyrics from extending under the custos at the end of a line.  These shifts are now disabled by default.  **NOTE:** This reverses the behavior formerly described in the documentation.
 
 ## [4.0.0-beta2] - 2015-08-26
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Fixed
 - End of line shifts now have an effect on the score.  When activated they prevent lyrics from extending under the custos at the end of a line.  These shifts are now disabled by default.  **NOTE:** This reverses the behavior formerly described in the documentation.
 
+### Added
+- New distance, `initialraise`, which will lift (or lower, if negative) the initial.
+- The first word of the score is now passed to a macro that allow it to be styled from TeX.  The first word is passed to `\GreFirstWord#1` and is styled by changing the `firstword` style.
+
 ## [4.0.0-beta2] - 2015-08-26
 ### Fixed
 - Corrected the rendering of explicit automatic and manual custos at the end of lines when the clef change that follows it is pushed to the next line (see [#569](https://github.com/gregorio-project/gregorio/issues/569)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Added
 - New distance, `initialraise`, which will lift (or lower, if negative) the initial.
 - The first word of the score is now passed to a macro that allow it to be styled from TeX.  The first word is passed to `\GreFirstWord#1` and is styled by changing the `firstword` style.
+- A new type of lyric centering, enabled with `\gresetlyriccentering{firstletter}`, which aligns the neume with the first letter of each syllable.
 
 ## [4.0.0-beta2] - 2015-08-26
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+
+
+## [4.0.0-beta2] - 2015-08-26
 ### Fixed
 - Corrected the rendering of explicit automatic and manual custos at the end of lines when the clef change that follows it is pushed to the next line (see [#569](https://github.com/gregorio-project/gregorio/issues/569)).
 - Distinguished between `eolshift` and `bolshift` giving each their own flag and user commmand for turning them on and off.  `\seteolshift{enable}` allows the lyric text to stretch under the custos at the end of the line.  `\setbolshift{enable}` aligns the beginning of each line on the notes instead of the text.  Both are on by default, but can be turned off with `\seteolshift{disable}` and `\setbolshift{disable}`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 
 ## [Unreleased][unreleased]
 ### Fixed
-- End of line shifts now have an effect on the score.  When activated they prevent lyrics from extending under the custos at the end of a line.  These shifts are now disabled by default.  **NOTE:** This reverses the behavior formerly described in the documentation.
+- Deactivating the end of line shifts now prevents lyrics from stretching under the custos at the end of the line.
 
 ### Added
 - New distance, `initialraise`, which will lift (or lower, if negative) the initial.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -3,6 +3,11 @@
 This file contains instructions to upgrade to a new release of Gregorio.
 
 ## 4.0
+
+### Font changes
+
+The music fonts (greciliae, gregorio, and parmesan) have changed with this release.  If the neumes do not appear or look strange in the output, you may need to clear your LuaTeX font cache by using `luaotfload-tool --cache=erase`.
+
 ### Command Name Systemization
 
 A naming scheme for GregorioTeX commands has been established and so most commands have had their names changed in order to bring them into line with the new scheme.  Some have also had their syntax changed.  Briefly, user commands all now have a `\gre` prefix (to prevent name colisions with other packages) and groups of commands which altered a single setting have been replaced by a single command which takes an argument specifying the value of the setting.  The notable exception to this are the two main commands: `\gregorioscore` (replaces `\includescore`) and `\gabcsnippet`.  See GregorioRef for the complete list of new command names and their syntax.
@@ -15,7 +20,7 @@ Additionally a new package option has been added. The option `deprecated=false` 
 
 ### Barred letters
 
-If you use barred letters and a font other than libertine in your document, you must adjust the horizontal placement of the bar on the letter.  To do so, use the `\gresimpledefbarglyph` macro.  For example, use `\gresimpledefbarredsymbol{A}{0.3em}` in your preamble, tweaking the second argument to have a good result (same for R and V).  If using LaTeX, you can alternately use the `\grelatexsimpledefbarredsymbol` macro, which allows control over bold and italic versions of the barred letter.  See the documentation of `\gresimpledefbarredsymbol` and `\grelatexsimpledefbarredsymbol` in the PDF documentation for more details.
+If you use barred letters and a font other than Linux Libertine in your document, you must adjust the horizontal placement of the bar on the letter.  To do so, use the `\gresimpledefbarglyph` macro.  For example, use `\gresimpledefbarredsymbol{A}{0.3em}` in your preamble, tweaking the second argument to have a good result (same for R and V).  If using LaTeX, you can alternately use the `\grelatexsimpledefbarredsymbol` macro, which allows control over bold and italic versions of the barred letter.  See the documentation of `\gresimpledefbarredsymbol` and `\grelatexsimpledefbarredsymbol` in the PDF documentation for more details.
 
 If you were using `\Vbarsmall`, `\greletterbar`, and `\greletteraltbar`, you must use `\gresimpledefbarredsymbol` to redefine your barred letters (see PDF documentation for details).
 
@@ -37,7 +42,7 @@ can become
 
     <eu>E(i) u(i) o(i) u(h) a(h) e</eu>(fe..)
 
-This will prevent line breaking, so if you were using so called *no linebreak areas* (with `{` in gabc) just for Euouae blocks, you can switch to this new tag, it will make things clearer and allow further spacing customization.
+This will prevent line breaking, so if you were using so-called *no linebreak areas* (with `{` in gabc) just for Euouae blocks, you can switch to this new tag, which will make things clearer and allow further spacing customization.
 
 ### Horizontal episema improvements
 
@@ -56,11 +61,11 @@ Note: `3`, `4`, and `5` encompass a new feature and are listed here only for com
 
 - `beforechoralsignspace` has been renamed to `beforelowchoralsignspace`.
 - `lowchoralsignshift` has been renamed to `choralsigndownshift`.
-- `highchoralsignshift` has been renamed to `choralsignupshift` and its sign inverted.
+- `highchoralsignshift` has been renamed to `choralsignupshift` and its sign (direction of shift) inverted.
 
 ### Styling score elements
 
-Changing the styling of text elements of the score (the initial, translations, etc.) formerly required the user to redefine a command which took an argument while changing the styling of the staff lines had a command specialized to that purpose.  All formats can now be changed via the `\grechangestyle` command.  This command takes two or three arguments.  The first argument, required, is the format to be altered.  The defined formats are:
+Changing the styling of text elements of the score (the initial, translations, etc.) formerly required the user to redefine a command which took an argument, while changing the styling of the staff lines had a command specialized to that purpose.  All formats can now be changed via the `\grechangestyle` command.  This command takes two or three arguments.  The first argument, required, is the name of the format to be altered.  The defined formats are:
 
 - `greinitial`: normal initials
 - `grebiginitial`: big (2-line) initials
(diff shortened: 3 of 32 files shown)


rpspringuel:
See [gregorio-test#17](https://github.com/gregorio-project/gregorio-test/pull/17) for a test showing the effect of the settings.

henryso:
Reviewing the code and documentation, this looks fine, but a lot of the tests appear to be breaking because of spacing.  Can you check that they are correct and accept the new results as appropriate into gregorio-project/gregorio-test#17?

rpspringuel:
@eroux 
> I admit I don't really understand this code, and I'm affraid it doesn't correspond to the first intention...
What, then, is the intention of the end of line shifts?  Part of the problem for me, in writing this fix, is that the original intention of the end of line shifts is unclear.  This fix is based on my best attempt at understanding what they were for based on the comments in the code.  Seeing as the old code which attempted to implement the eol shifts wasn't working, it's probably best, for the moment, to forget about implementation details.  Can you just give me a description of what the desired effect of the eol shifts is?

eroux:
The desired effect is in fact the same as bolshift: to allow text to go beneath the custo. This doesn't make a difference as big as bolshift because custos are relatively small, but in my experience (at least what I remember of it), it sometimes allow TeX to push an additional syllable on a line. 

According to me, it should be implemented the same way bolshifts are, meaning it should be done on all syllables, whatever the value of `lastoflinecount`, with the same mechanism of `\kern -x\hbox{}\kern x`. The main blur zone for me is to know how this will get along with spaces at the end of syllables...

Is it clearer for you? Don't hesitate to tell me it isn't... Also, if you think it's difficult (because it is), I can work on this, I don't think it should be more than a few lines of TeX.

rpspringuel:
Okay, that matches my understanding of what they were for, so at least we're on the same page there.

On implementation, there are two problems with implementing the bol strategy directly.

1. TeX does not ignore spaces at the ends of lines the way it does at the beginnings.  As a result what's needed between the two kerns is not a `\hbox{}`, but rather something which tells TeX that the line break, if needed, should go here.  After some searching online I believe that's `\allowbreak`.  Once the line break occurs, the second kern will be pushed to the beginning of the next line, where it will be ignored.
2. Manual line breaks are still going to need their own special code because printing them inserts a line break.  We thus need to make sure the shift happens before we get to that line break, and thus has to happen as part of the actual printing process.  This means that when a manual line break occurs, both of the "normal" shifts will be pushed to the start of the next line so we have to make sure that there is nothing between them (like a `\hbox{}`) which would break TeX's ignorance of spaces, so that they both can be ignored.

I think I've been able to solve both of these issues and will be uploading a new try soon.  As you indicated the change is very small.  As a result, I'm double checking that I've included all the factors and am trying to refine my example test so that it shows the results more clearly.

eroux:
You're right, I didn't have in mind the difference with beginning of lines... 

What's handling the line breaking are penalties, `allowbreak` is just a zero penalty (it's defined in [ltplain.dtx](http://latex-project.org/svnroot/latex2e-public/trunk/base/ltplain.dtx). Gregorio has its own set of penalties, inserting a zero penalty would have a lot of bad consequences... 

What is needed here is to compute the eolshift, and then, when gregorio inserts the penalty at the end of syllable (it should be easy to find, but there are many possibilities of penalties, so maybe some things will need to get factorized first), do something like:

- `\kern -\eolshift`
- `\grenobreak` (which is a penalty of 10001)
- `\hbox{}`
- `\grepenalty -\therelevantvalue`
- `\kern \eolshift`

this should be working...

rpspringuel:
Okay, here's my new implementation attempt.  The shifts are now moved so that they surround the end of syllable penalties.

Question: does `\gre@endafterbar` need the eolshifts?

@henryso, I haven't forgotten the bit about the tests being broken, I'm just trying to get the eol shift behavior right before worrying about the other tests.

rpspringuel:
Okay, with that sorted out, I've looked at the tests that are differing and it seems to me that most of the differences (if not all) are at the front end.  I.e. They appear at the beginning of lines.  Since I didn't change those, @eroux, would you take a look and see if you like the results.  Personally I think they look fine and would be willing to accept them.

eroux:
It would be interesting to know where these differences come from, especially at the beginning of lines... the only reason I see for such differences (but I might be wrong) is that TeX eats only one space at beginning of lines, and thus adding eolshift might introduce a bug... I'll try to find out.

rpspringuel:
The shifts mostly seem to be towards the left and are mostly on the first line of a score.  That indicates to me that it more of the rightward bolshifts are getting eaten than before.

eroux:
Oh, ok, I think it comes from [here](https://github.com/rpspringuel/gregorio/commit/924804fe9b39d97ca72ee27832c0a772b284b7ca) then, it shouldn't be a problem.

rpspringuel:
Interestingly, reversing that commit doesn't resolve the differences and indeed creates a whole bunch of others, so it seems that there may be some interaction effects between the bolshift changes and the eolshift changes.  I'm going to take a closer look at the changes before accepting them.

rpspringuel:
Okay, I take back what I said earlier.  If I check out 924804f and 6ee9c36 (the matching commits from gregorio and gregorio-test, respectively) I get 39 tests failing due to spacing issues and the trend seems to be that neumes on the first line have been pushed **right**, not left as I first thought, but only if there is no initial.  Looking closely at the code, I think what's happening is that where as previously the lack of an initial lead to the first rightward bolshift being ignored and so we started with the leftward shift after the changes this shift is no longer ignored.  This must mean that the empty initial is triggering the "start of line" signal for TeX.  Indeed, I'm willing to bet that any initial triggers that "start of line" signal leading to the rightward bolshift having an effect, it's just that this behavior when the initial is present is consistent on both sides of this commit.

Personally I don't think an initial (or lack there of) should count as the "start of line" for the purposes of the bolshift.  That is, there should be an exception in the bolshift code that prohibits the rightward shift for the first syllable of the score.  If people agree with me I'll look into fixing this.

eroux:
bolshifts are about lines left-alignment. Why should the first line not be aligned with others (in the case where there is no initial)? I can't see why you would want to to make an exception for the first line...

Anyways, I think you're right, what happens with the first line is that lines are drawn directly at beginning of lines (while on other lines they are in a localleftbox), so TeX cannot eat spaces. So I think the solution would be not to move the first syllable of the score right. What do you think?

rpspringuel:
> bolshifts are about lines left-alignment. Why should the first line not be aligned with others (in the case where there is no initial)?

That is my point. While the old code preseved the left alignment, the new code does not. The exception would nullify just the rightward shift, allowing the leftward one to align the first line with the others.

eroux:
Is this ready to be merged or do you still plan to work on something?

rpspringuel:
Still working on the bolshift problem on the first line.

rpspringuel:
Please look at [584-2temp](https://github.com/rpspringuel/gregorio/tree/584-2temp) and [the corresponding branch in gregorio-test](https://github.com/rpspringuel/gregorio-test/tree/584-2temp).  This implements a potential fix for the bolshift issue.  It necessitates revising those tests where there was an initial (the first syllable moves closer to the initial), which I expect and have already accepted on the branch.

There is still a difference in the tex-output/clef-change test that I'm not sure what to do with.  It seems that bare clefs (i.e. ones which appear mid-line but which are not part of a `(z0::c3)` structure, are being shifted to the right (with knock-on effects on subsequent glyphs).  I'd appreciate it if people would look at this shift and decide if it's acceptable, or if I should try to drag the clefs back to the left.  You don't need to get into the code details (though hints are appreciated) I'm just interested in the visual decision at this point.

Once that last test is ironed out, I'll merge 584-2temp into this branch and start checking for interaction effects.

eroux:
The code looks good, can you just paste a screenshot of the difference with the clef here?

rpspringuel:
Red marks the expected position (where things were before we started messing with this), blue the position where things are showing up after the changes.

diff-page

eroux:
I admit that's a bit strange... but the shift is quite big, I think it would be worth inspecting. Do you know the exact commit causing this shift?

rpspringuel:
I haven't figured that out yet, but it almost has to be 3d334bc (my changes which stop the bolshift from affecting the first syllable of the score) or 924804f (your changes which removed the dependency of the first bolshift on lastoflinecount).  The only other commit on the 584-2temp branch is the merge which brought in the most recent changes on develop, and those changes were passing all the tests.  I suppose some weird interaction effect could be causing it, but I'm more inclined to look at the actual changes to bolshifts first.

rpspringuel:
Okay, I can confirm that 924804f introduced the push of the clefs the clef-change test fails on this commit, but passes on the one immediately before it in the history.  Since the clefs are getting pushed right, that would imply that `\gre@lastoflinecount` was not 2 when `\GreChangeClef` was invoked, resulting in the rightward shift getting ignored because it was under that `\ifnum\gre@lastoflinecount=2` conditional.

Inspecting the gtex file produced for clef-change.gabc, I note that the clefs that are getting pushed are inside `\GreDiscretionary` and that is preceeded by `\GreSetLinesClef` while the others are not.  Looking at these two functions, the thing that jumps out at me is that `\GreDiscretionary` changes `\gre@lastoflinecount` to 1 inside of the first argument of `\discretionary` and to 2 inside of its second argument.  After adding some debug messages to monitor what is happening it appears that The first argument of `\discretionary` is the one getting printed.  This is weird because the first argument of `\discretionary` is supposed to be the pre-break text, not the no-break text.  However, looking in the log file I also see that the last discretionary pairs argument 2 and 3 (what should be the post-break and no-break text).  It thus appears that `\discretionary` is expecting the arguments as `{<no-break>}{<pre-break>}{<post-break>}`.  However, if a use `\discretionary` in a dummy document compiled by LuaLaTeX it works with the arguments in the usual order (`{<pre-break>}{<post-break>}{<no-break>}`).  Does anyone have any ideas as to why the argument reordering is happening and what might be done to fix it?

Even without fixing the above, however, I can now tell why the clefs were pushed left in the old code: the bolshift right wasn't being executed inside of the discretionary.  So the question is should it?  If so, it's fairly simple to add another exception that prevents it from happening inside a discretionary (I've already have a tested solution ready to go).

eroux:
`\discretionary` in TeX is a bit difficult, I'm not sure what code gets executed, maybe all (which would invalidate the setting of lastoflinecount)... it's just that it keeps the different results in boxes and during the paragraph break pass, it tries to find the best possibility. So maybe the lastoflinecount setting in GreDiscretionary is just wrong... Anyway, the bolshift should be applied to what's inside discretionaries, which certainly means removing the first shift right when inside a discretionary.

rpspringuel:
Okay, I've fixed the bolshifts and have merged that fix into this branch.  I'm now looking at the eolshift code as there are still a whole lot of tests failing.

rpspringuel:
Okay, I've got the end of line shifts working.  Since the default behavior is to apply them, this is breaking a whole bunch of tests either by stretching the lines to the right a bit or by changing the location of line breaks.  Before I go accepting all these changes, would someone else take a look at them and tell me what they think?  Some of the results seem to put the last neume awfully close to the custos.  As an example here's the gabc-output/montes.gabc test.
page
Lines two and three exemplify what I'm talking about.  Should I try to increase that distance some or is this an acceptable result.

henryso:
Is the minimum space before the custos adjustable?  If not, it should be, and it should default to something greater than what you're getting, because I don't find that result aesthetically pleasing.

rpspringuel:
The space before a custos is adjustable, but there is only one such space defined: `spacebeforecustos`.  Adjusting that would affect custos everywhere, not just here.

Looking at the space, it's a rubber length, but we're using its nominal value in determining the size of the eolshift, not its minimum.  Perhaps I should change that so that eolshift only takes into account the minimum.

henryso:
Are you saying that using the nominal value can allow a space between the last note and the custos which is smaller than the minimum distance, then that sounds wrong to me, and I think it should be fixed.  If, on the other hand, that small space *is* the minimum distance, then I suppose it is working correctly and people (like me) who feel the space is too small can easily fix it by adjusting `spacebeforecustos`.

eroux:
I admit it looks a bit odd... Maybe TeX eats too much space? Did you try to replace spacebeforecusto by something big like 5cm, to see if it was taken into account?

rpspringuel:
@henryso, what I'm saying is that `spacebeforecustos` can be shrunk by TeX (being a rubber length) and so might end up being smaller than it's nominal value on a given line (rubber lengths are composed of a nominal value, a stretch, and a shrink).  The eolshift, however, is a fixed length based on the nominal value of `spacebeforecustos`.  What I'm thinking is that it might need to be based on the minimum value of `spacebeforecustos` (its nominal value minus its shrink).

@eroux, I haven't done any testing yet.  I'll do that now and see what I can figure out.

rpspringuel:
Okay, I've done some testing and it seems that `spacebeforecustos` is sometimes respected and sometimes not.  Look at the following example:

page

In this example (a modified form of the tex-output/SalveReginaOP test), `spacebeforecustos` has been set at 1cm.  However, lines 4 and 5 don't respect that.  For the life of me, I can't tell why.  When I first noticed this, I thought it was that lines which end with bars respected and ones which ended with notes don't, but with this setting, line 3 ends with a note and does respect the space.

Does anyone have any ideas as to what is going on or where I should be looking in the code for the potential problem?

eroux:
I tend to suspect the hyphens...

henryso:
@eroux In the previous example (Montes...), the `alleluia,` at the end of the third line also appears to be a problem, and does not end with a hyphen.

eroux:
indeed, and it's not even an end of word... interesting... can you add a `\hbox{}` before or after the spaces so that we're sure that TeX doesn't eat them?

rpspringuel:
A variant on the SalveReginaOP which shows a hyphen respecting the space:
page
`spacebeforecustos` = 1.1cm in this one.

rpspringuel:
Adding `\hbox{}` either before or after (or both) the rightward eolshift doesn't have any effect.
Adding one before or after (or both) the leftward eolshift breaks the beginning of line positions (as expected).

I'm suspecting that the size limits on the eolshift need some adjustment.  Right now the only limit is that it isn't allowed to be bigger than the enddifference.  I think in the case of small enddifferences, though it needs to be more complicated than that.  I'm attempting to work through the logic of how much space we're allowed to reclaim and under what conditions.

rpspringuel:
Okay, after thinking things through, here's what I've concluded about amount reclaimable space (the size of the `eolshift`):

1. There is only reclaimable space if the lyrics go further than the notes (`enddifference` is negative).
2. The amount of reclaimable space is `-enddifference - spacebeforecustos`.  Note that this is negative for if `enddifference` is smaller than `spacebeforecustos`.
3. The amount of reclaimable space maxes out at the width of a custos plus the `spacebeforecustos`.  Anything bigger would cause the lyrics to stick out into the margin.

I haven't had time to turn this into code yet, but these rules are not what's in `\gre@calculate@eolshift` at the moment.  I'm going to try changing that (either tonight or tomorrow) and see if that doesn't result in something nicer.

Unless of course some one comes along and points out a flaw in the above description.

eroux:
Shouldn't custo width be taken into account too? Like replacing `spacebeforecustos` by `(spacebeforecustos+custowidth)` in your previous comment?

rpspringuel:
@eroux, I see the logic there, but having never seen a note overlap with the custos, it seems that TeX is already taking the width of the custos into account.  Indeed, after some additional testing, it appears that it is taking into account the `spacebeforecustos` skip in the `localrightbox`, despite the fact that we're seeing instances of it appearing to be ignored.  Really, this should have been obvious from the bars (which don't have `eolshift`s) and the behavior when `eolshift`s were disabled.  So, the amount of reclaimable space is just `-enddifference`.

This means that the `eolshift` calculation was nominally right, but obviously wasn't working as intended.  As a result, I've modified the `eolshift` calculation and have been able to make it work.  The new code isn't as visually efficient as the old (more lines), but it does appear to work as intended.

I'm going to work my way through all the failing tests to make sure and will commit the changes once that's done.

eroux:
Good job!

rpspringuel:
Changes posted.  There is still one test which fails (tex-output/bugs/fix-396), and it fails pretty badly, but it also looks like the original result wasn't pretty either.  Can someone else look at that result and tell me what they think.

Another thing I noticed in examining the results is that there's an alignment issue (i.e. an incorrect `bolshift`) when one switches from a do clef to a fa clef.  See below:

page

You'll note that the line which begins with the fa clef has the notes starting later then those which begin with a do clef.  Looking at the code, this is because the `bolshift` is dependent on the actual size of the current clef, and naturally the fa clef is wider.  I'm thinking that the `bolshift` calculation should be based on the width of just one of the clefs (probably the wider, fa clef) and not on whichever one happens to be the current clef.  Comments?

henryso:
The purpose of tex-output/bugs/fix-396 is to test that the euouae block doesn't cause TeX to start the line following it with the double-bar (see #396).  Now that the shifts cause the entire euouae block to be pulled into the previous line (which isn't, in and of itself, bad), the test needs to be tweaked (somehow) to exhibit the same condition, where TeX would likely put the double bar on the next line, but is prevented from doing so by the penalties set there.  This is why I originally added add the idiotic repercussion on "Patri" for the test.

henryso:
Regarding the fa clef: I know it's not simple, but I wouldn't compute the bolshift based on the fa clef unless there is actually a fa clef somewhere in the piece.  That said, I don't find the current result above particularly unpleasant.

rpspringuel:
> The purpose of tex-output/bugs/fix-396 is to test that the euouae block doesn't cause TeX to start the line following it with the double-bar (see #396). Now that the shifts cause the entire euouae block to be pulled into the previous line (which isn't, in and of itself, bad), the test needs to be tweaked (somehow) to exhibit the same condition, where TeX would likely put the double bar on the next line, but is prevented from doing so by the penalties set there. This is why I originally added add the idiotic repercussion on "Patri" for the test.

In that case the logical solution seems to be to disable end of line shifts for that test in order to restore the old behavior.

henryso:
Would you be able to make that change as part of this?  If not, I can do it, but will need a little guidance as how to "disable" the eolshifts (set them to 0?).

eroux:
I think we can keep the current behavior with the clef for now, as the result is not obviously wrong... So there should probably be an option for that... Can you report it in another issue? Doesn't have to be in 4.0 milestone. I think it would be quite easy to add though, something like `\LargestKey{c}` at score beginning of the score...

eroux:
The changes look ok; from my point of view this can be merged, thank you for taking care of this!

rpspringuel:
I've updated the tests so that gregorio-test#17 should pass everything.  That means that the original intentions of this pull request have been met.

I'll open the new issue for the key issue.

I still haven't tackled Sr. Ruth's point on the `bolshifts` needing to account for the presence of an alteration so that alignment is done on the notes, not an initial alteration.  I can always open a new pull request for that one later though if someone wants to merge this one as is.

eroux:
Let's make another pull request for #599, this one is already quite big, thanks a lot!

PR #596, opened 2015-09-05
Cleared the glyph id attribute at the end of the score to stop leakage.

### LINKED ISSUE #595: weird spacing issue in gregorio-v4.0.0beta2
I have the same chant included twice, but the text on the line that needs to be lower ("Et Jesum, benedictum fructum" only works the first time.  It seems to have something to do with the first one being in braces.  This was the smallest I could get the code down to still reproduce the issue.  If I remove the first set of braces, both chants work.  Also if I remove the \greannotation from the second one, it works fine.

gist: to clone <https://gist.github.com/f454ddd7ab9baa39954b.git>, to view in a browser <https://gist.github.com/f454ddd7ab9baa39954b>

henryso:
Fixes #595.

As this is a fix to a beta feature, I am not adding to the change log.

Ready for review and merge if satisfactory.

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -1088,6 +1088,7 @@
     \parfillskip=\gre@save@parfillskip\relax%
   \fi %
   \global\let\ifgre@vowelcentering\ifgre@save@englishcentering\relax % DEPRECATED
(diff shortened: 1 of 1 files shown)


eroux:
Thanks!

PR #600, opened 2015-09-11
Added "firstletter" lyric centering.

henryso:
As discussed on the gregorio-devel mailing list.  Gregorio will produce five-argument calls to the syllable-setting macros in order to provide the extra break-down of the lyric after the first character.


henryso:
The gabc->gtex tests fail as expected.  Everything else passes.  I will also add an additional test specifically for the new feature if/when this is merged.

rpspringuel:
This TeX code looks a lot nicer than the the pipe version.  Just one thing I'd change.  Can you add a short verbal description of the cases when using `\ifcase` in addition to identifying the number?  It'll make things easier to read if someone has to revisit this code well down the line.

henryso:
Done.

eroux:
ok for me!

Issue #603, opened 2015-09-13
prepare for local discpenalty


eroux:
A change should appear in the next few days (or hours) in LuaTeX for [local disc penalties](http://tracker.luatex.org/view.php?id=928). It will certainly come with LuaTeX of next TeXLive, so I think we should prepare now. The idea is the following:

- we want clef change to have a specific penalty
- clef change are discretionaries
- other things could be discretionaries in the future, with other penalties
- currently all the discretionaries of a paragraph share penalty (which is the value of `\discpenalty` at the end of the paragraph)
- with the future version of LuaTeX, discretionaries' penalty will be the current value of `\discpenalty` when the discretionary is created, so in the same paragraph, discretionaries can have different penalties

To have a discretionary with a speciic penalty, one can do something like

    \begingroup
    \discpenalty mypenalty
    \discretionary{what}{e}{ver}
    \endgroup

What I propose is the following: the `\GreDiscretionary` macro should take an additional argument, which would be a type (it would have only one possible value now, `0` which would mean `clef change`). Then the `\discpenalty` would be set according to the type, inside a group. To be backward compatible, we should keep the current behavior (which is to set `\discpenalty` to the clef change penalty for the whole paragraph), but it could be marked as deprecated, with a removal date of summer 2017.

What do you think?

henryso:
I don't feel qualified to speak to the code behind this, but the logical argument makes sense to me, and doing it in 4.0 sounds like a good idea.  One thing I would change, however, is to relabel this issue as an enhancement rather than a bug.

eroux:
Indeed, the labelling was a mistake (I didn't want to label it that way)

henryso:
Is everything for this issue done?

eroux:
current state is ok yes

PR #604, opened 2015-09-13
add type argument to GreDiscretionary (for #603)

eroux:
This adds the argument to `\GreDiscretionary` as explained in #603. It is not taken into account yet, as the code is not merged yet in LuaTeX and the LuaTeX developers just told me that they might use `\hyphenpenalty` instead of adding `\discpenalty`... But the handling of the argument can wait a future release...

### DIFF
--- a/doc/Command_Index_gregorio.tex
+++ b/doc/Command_Index_gregorio.tex
@@ -168,8 +168,8 @@ Typesets a custos.
 \macroname{\textbackslash GreDagger}{}{gregoriotex-symbols.tex}
 Macro to typeset a dagger (\GreDagger).
 
-\macroname{\textbackslash GreDiscretionary}{\#1\#2}{gregoriotex-signs.tex}
-A GregorioTeX specific discretionary used to avoid clef change at beginning or end of line, or even with more complex data (z0::c3 for instance).  We require a special function because in the normal discretionary function you cannot use \verb=\hskip= (but you can use \verb=\kern=) and you cannot use \verb=\penalty= (which is useless indeed).  This macro corrects for these two limitations.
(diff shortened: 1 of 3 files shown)


doc/Command_Index_gregorio.tex
```diff
@@ -168,8 +168,8 @@ \section{Gregorio Controls}
 \macroname{\textbackslash GreDagger}{}{gregoriotex-symbols.tex}
 Macro to typeset a dagger (\GreDagger).
 
-\macroname{\textbackslash GreDiscretionary}{\#1\#2}{gregoriotex-signs.tex}
-A GregorioTeX specific discretionary used to avoid clef change at beginning or end of line, or even with more complex data (z0::c3 for instance).  We require a special function because in the normal discretionary function you cannot use \verb=\hskip= (but you can use \verb=\kern=) and you cannot use \verb=\penalty= (which is useless indeed).  This macro corrects for these two limitations.
+\macroname{\textbackslash GreDiscretionary}{\#1\#2\#3}{gregoriotex-signs.tex}
+A GregorioTeX specific discretionary. Currently only used to avoid clef change at beginning or end of line, or even with more complex data (z0::c3 for instance).  We require a special function because in the normal discretionary function you cannot use \verb=\hskip= (but you can use \verb=\kern=) and you cannot use \verb=\penalty= (which is useless indeed).  This macro corrects for these two limitations. The first argument allows to select the penalty assigned to the discretionary by recent version of LuaTeX.
```
henryso: I think it would be a good idea to add the argument table here, which indicates that 0 = clef change.

eroux: added!

henryso:
Looks good to me.  One test fails because of the new argument, so that's acceptable.

PR #606, opened 2015-09-14
Unix version of system-setup diagnostic tool

rpspringuel:


henryso:
This should be added to the Makefile.am so that it's included in the distribution archive.  The Windows batch file should probably also be added to that file as well.

rpspringuel:
Done, though I'm not so sure that adding the Windows batch file is really necessary as Windows users aren't likely to be downloading the distribution tarball.

henryso:
This is true, but we might as well include it for completeness.

PR #608, opened 2015-09-14
Used VersionManager.py to control the squarize.py version string.

henryso:
This eliminates the need for python2 when building the fonts (the built-in fontforge python interpreter is still necessary).

All tests pass on the resulting fonts.


henryso:
Since this appears to work, I am fast-tracking the merge so I can use it in response to Felix Yeung's problem on the mailing list.

PR #609, opened 2015-09-15
Deprecation fixes

rpspringuel:
The flag for the deprecation option needs to be defined on the PlainTeX side.  See gregorio-project/gregorio-test#19

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -30,7 +30,7 @@ The optional arguments are:
     THEN Gregorio\TeX\ will compile a new gtex file.\\
   forcecompile & Gregorio\TeX\ will compile all scores from their gabc files.\\
   \hline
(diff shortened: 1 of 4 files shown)


henryso:
This is a minor nitpick, so feel free to ignore it, but I keep getting caught by the name of this flag.  In my head, the sense is -- at first glance -- reversed.  I see it as "show deprecation warnings" rather than "allow deprecated usage."  This would not happen (again, in my head) if the flag were renamed to, for example, allowdeprecated.

rpspringuel:
I've no problem changing the name.  @eschwab named it originally, so I'd like to hear his input at least before making the change.

eschwab:
`allowdeprecated` is fine by me.

rpspringuel:
Done, despite the autocorrect "fixing" my commit message.

PR #614, opened 2015-09-18
Feature/distance flags

rpspringuel:
This PR changes a few things related to the distances:

1. In keeping with the new naming conventions the `0` and `1` arguments of `\grecreatedim` and `\grechangedim` are changed to `fixed` and `scaling` respectively.
2. Previously `\grechangedim` had called `\grecreatedim` as its internal workhorse function.  This was messy when I went looking to introduce deprecation messages, so I refactored out the common internals into a separate function `\gre@dimension` which both functions can now call.
3. The scaling flag for each distance is switched from being a number stored in a macro to being a boolean.
4. `\grescaledim` wasn't working right for all of its keywords.  I've fixed this and added a new keyword (`scaling`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - All of the keywords for `\grescaledim` were now work as described in the documentation.
 
 ### Changed
-- `\grecreatedim` and `\grechangedim` now take keywords for their third argument (`scaling` and `fixed`( instead of integers (`1` and `0`) to make the more in keeping with the overall user command conventions.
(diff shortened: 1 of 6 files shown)


henryso:
I think "scaling" should be "scalable".  You'd say "this distance is fixed" or "this distance is scalable".  Saying "this distance is scaling" or "this distance is scaled" is not exactly correct since the distance is not in the process of scaling nor has scaling already occurred.  What do you think?

rpspringuel:
I was thinking in terms of "scaling distance" vs. "fixed distance" but now that you mention it, even "scalable distance" sounds more appropriate.  I'll make the change when I finish running some tests on another branch.

eroux:
I think this could be merged? That would be ok for me

Issue #599, opened 2015-09-09
bolshift for flats


eroux:
When a line starts with a flat, the beginning of line alignment makes it aligned with the notes, while it would be better, when possible, to shift it left a bit more (it can be quite close to the key). I believe that a little change in `\gre@compute@bolshift` might be enough for this... Maybe by changing this line:

    \advance\gre@dimen@clefwidth by -\gre@dimen@minimalspaceatlinebeginning %

and changing `\gre@dimen@minimalspaceatlinebeginning` to `clefflatspace` if the first glyph is a flat...

See [mail archives](https://mail.gna.org/public/gregorio-users/2015-09/msg00028.html) for original report

rpspringuel:
Seeing as I'm currently futzing around with the bolshift code, I'll see what I can do about this.

eroux:
Thanks a lot! if it's a lot more complex than what I describe, don't hesitate to ask for help, I'll try to be more helpful than on the eol thing!

rpspringuel:
I've started work on this and have run into an issue.  I obviously need to trigger the altered `bolshift` calculation based on the presence of an alteration at the beginning of the glyph.  Since these alterations are already taken into account when aligning the text and the notes, Gregorio clearly already has the necessary infrastructure in place to know this.  Examining the functions, it appears that this should be indicated by the notes alignment type argument (#4) for `\GreGlyph`.  However, when I try to trigger off this, nothing happens.  After adding some debug messages to figure out why, I've come to the realization that the command-line tool always sets this argument to 0 (and by always I mean that all of the tests in gregorio-test always set this argument to 0 when writing `\GreGlyph` into the gtex file).  This would imply that the code related to this argument is dead and should be eliminated.  Is this the case?  Are there conditions (not present in the test repository) where this argument takes on a non-zero value?

henryso:
In the output from tests/gabc-output/montes.gabc, there are several cases where argument #4 is not 0.

rpspringuel:
My mistake.  I had a `<` where I should have had a `>` in the code that was looking for those occurrences.  So, the code is not dead.  Now to run some more tests to figure out why my code isn't working the way I intended it to.

eroux:
@rpspringuel this can be closed, right?

rpspringuel:
Yes.

PR #615, opened 2015-09-19
Initial alterations are ignored when aligning the first note of each line (bolshift)

rpspringuel:
Addressing #599

I've added an additional component (`bolextra`) to the `bolshift` which allows an initial alteration to be ignored when aligning the first notes of each line.  This works with the current setting for `minimalspaceatlinebeginning` but there are a couple of problems:
  1. if `minimalspaceatlinebeginning` is increased the lines which start with an alteration do not move
  2. if the initial alteration is a sharp, it overlaps with the clef.  i.e. there is not enough space for the alteration

I have also noticed that `minimalspaceatlinebeginning` has an upper limit on what is respected.  This makes me think that `bolshift` isn't being calculated correctly (at least in this aspect) and may be the source of the above two problems.

Any suggestions on improving this are appreciated.

### DIFF
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1652,7 +1652,6 @@
       \gre@skip@temp@four = \gre@skip@alterationspace\relax%
       \global\advance\gre@dimen@bolextra by \dimexpr\gre@skip@temp@four\relax%
       \gre@debugmsg{bolshift}{bolextra: \the\ger@diment@bolextra}%
-%			\kern -\gre@dimen@bolextra\relax%
     \fi%
   \fi %
   \gre@calculate@glyphraisevalue{#1}{0}%
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1647,12 +1647,6 @@
   \setbox\gre@box@temp@width=\hbox{#2}%
   \ifnum#4=0\relax %
     \gre@newglyphcommon %
-    \ifgre@firstglyph%
-      \global\gre@dimen@bolextra = \wd\gre@box@temp@width%
-      \gre@skip@temp@four = \gre@skip@alterationspace\relax%
-      \global\advance\gre@dimen@bolextra by \dimexpr\gre@skip@temp@four\relax%
-      \gre@debugmsg{bolshift}{bolextra: \the\ger@diment@bolextra}%
-    \fi%
   \fi %
   \gre@calculate@glyphraisevalue{#1}{0}%
   \ifgre@hidealtlines %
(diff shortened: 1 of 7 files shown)


tex/gregoriotex-signs.tex
```diff
@@ -1644,19 +1644,21 @@
...
+  \setbox\gre@box@temp@width=\hbox{#2}%
   \ifnum#4=0\relax %
     \gre@newglyphcommon %
+    \ifgre@firstglyph%
+      \global\gre@dimen@bolextra = \wd\gre@box@temp@width%
+      \gre@skip@temp@four = \gre@skip@alterationspace\relax%
+      \global\advance\gre@dimen@bolextra by \dimexpr\gre@skip@temp@four\relax%
+      \gre@debugmsg{bolshift}{bolextra: \the\ger@diment@bolextra}%
+%			\kern -\gre@dimen@bolextra\relax%
+    \fi%
```
eroux: I think computing bolextra separately from bolshift is not such a great idea, I think it would be much better to compute a different bolshift if the first glyph is an alteration, this would make calculations easier and cleaner, and avoid the bugs you have... is it a lot more complex?

rpspringuel: `bolextra` depends on the size of the alteration and thus needs to be computed in a location where GregorioTeX knows what sort alteration is being printed.  This information is only available inside the macro which prints the alteration, and is not currently available to `\gre@calculate@bolshift`.  I thus have to pass that information out from `\gre@alteration`, which I do via `bolextra`, which is then used in the calculation of `bolshift`.

You might be confused by the `\kern` line in the portion of the code you're looking at.  That's a left over from my testing of where the best place to incorporate the extra shift was and is commented out.  I'll delete it in order to save the confusion.

eroux: Not really, in fact we need the alteration shift only if `bolshift > width(alteration)`, which will, I think, almost never be the case... In almost all cases, what we'll want is an alteration separated from the clef by `clefflatspace`. Do you see what I have in mind? If we do this, then there are two cases:

- the note is still to far right, well, at least we'll have done what we could, we cannot make it move left any further anyway, so it's ok
- the note is too far left, but I don't believe this case can happen...

do you see what I mean?

rpspringuel: I don't really understand what you mean, but I will point out that your condition (`bolshift > width(alteration)`) is still dependent on knowing what sort of alteration we're dealing with as the alterations all have different widths.

eroux: this conditional was never meant to go into TeX, it was just for the discussion. I'll make the pull request if you want, maybe it will be easier. I'll try that tomorrow.

rpspringuel:
I've reworked the `bolshift` calculation in a way that I think works better, based on my testing.  It incorporates `bolextra` right from the beginning and applies the appropriate limits in the various cases to make sure that stuff isn't pushed too far to the left.

This breaks some tests, but in an expected way.  See gregorio-project/gregorio-test#25 for the test changes (including a new test which expands the number of beginning of line conditions tested).

eroux:
Seems good to me, thanks a lot for this! Do you think it's ready to be merged?

rpspringuel:
Stupid autocorrect.

Unless I've missed something else in the documentation, I think this is ready to merge.

PR #616, opened 2015-09-20
unfinished proposal for #599

eroux:
@rpspringuel this is the kind of things I'm thinking about. It seems more simple than your proposal, and should work in 99.99% of cases. This is not finished, but it's just to give you the algorithm I'm thinking about.

BTW, there was a missing `%` that I fixed, maybe it's important that we parse the .tex files before next release, to see if it was the only one...

### DIFF
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -538,6 +538,7 @@
   \gre@debugmsg{general}{}%
   #1%
   \gre@firstglyphtrue%
+  \xdef\gre@firstisalteration{0}%
   \gre@calculate@textaligncenter{\gre@firstsyllablepart}{\gre@middlesyllablepart}{0}% we first get the width between the alignment point and the end of the syllable
   \gre@syllablenotes{#9}% we put the notes in a box, so that we have the width of it
   % now we calculate the begin difference, that is to say \gre@dimen@notesaligncenter - \gre@dimen@textaligncenter
@@ -739,6 +740,7 @@
   % the main goal is, when there is no text under the bar, to put the bar in the middle of the space between the last note of the previous syllable and the first note of the next syllable. But there are limits : a bar can't go very far above text. For example if there is "nuncncncncn" with a punctum on the u, the bar can't go above the fourth n, the most far position is the position where the end of the bar is above the end of the word. The same limitation applies for the syllable after the bar.
   % there are two different cases that have almost nothing in common : the case where there is something written under the bar, and the case where there is nothing.
   % first of all we need to calculate previousenddifference, begindifference, enddifference and nextbegindifference.
+  \xdef\gre@firstisalteration{0}%
   #1%
   \gre@calculate@textaligncenter{\gre@firstsyllablepart}{\gre@middlesyllablepart}{0}%
   \setbox\gre@box@syllabletext=\hbox{\gre@fixedtextformat{\gre@pointandclick{\gre@firstsyllablepart\gre@middlesyllablepart\gre@endsyllablepart}{#6}}}%
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -556,7 +556,7 @@
   % took me years to find it!
   \gre@debugmsg{bolshift}{Current glyph: \the\gre@attr@glyph@id}%
   \ifgre@bolshiftsenabled%
-    \gre@calculate@bolshift{\gre@dimen@begindifference}%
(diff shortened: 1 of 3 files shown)


rpspringuel:
I noticed that missing `%` too and fixed it in my branch as well.

tex/gregoriotex-syllable.tex
```diff
@@ -556,7 +556,7 @@
   % took me years to find it!
   \gre@debugmsg{bolshift}{Current glyph: \the\gre@attr@glyph@id}%
   \ifgre@bolshiftsenabled%
-    \gre@calculate@bolshift{\gre@dimen@begindifference}%
+    \gre@calculate@bolshift{\gre@dimen@begindifference}{\gre@firstisalteration}% TODO: compute firstisalteration
```
rpspringuel: You've still got to know that the first piece is an alteration, which involves doing something in the same basic place as where I calculated `bolextra`.  Frankly, I don't see the advantage of setting a flag vs. calculating a distance.  Both have to be passed to `\gre@calculate@bolshift` either implicitly (as I did) or explicitly (as you did).

rpspringuel:
Once made effective, this will be the equivalent of setting `bolextra` = `\gre@dimen@minimalspaceatlinebeginning - \gre@skip@clefflatspace` in my version of the code.  I did that to see what the results would look like and this is what I got (using gabc-output/bugs/fix-243):

page

The diff makes it a bit clearer what's happening:

diff-page

The lines which start with an alteration are all moved left, but they are all moved the exact same amount (the difference between `minimialspaceatlinebeginning` and `clefflatspace`).  The result is they are not any better aligned amongst themselves and still about a punctum's width out of alignment with the bare punctum in the last line.

Compare that to the result from my code:

page

Here all the punctums are aligned (which I thought was the desired goal), but the sharp gets pushed into the clef.

tex/gregoriotex-syllable.tex
eroux: Indeed (I admit I tend to prefer to make code as explicit as possible, having many globals make it more difficult to understand (hence the additional argument to calculate@bolshift)

eroux:
I disagree, this is not equivalent to setting bolextra as you said, as there are extra checks and transformations performed after the conditional, it's not just adding a distance...

eroux:
can you try with latest comit?

tex/gregoriotex-syllable.tex
rpspringuel: I don't necessarily disagree with you philosophically, it's just that TeX is designed in such a way that variables are global.  Hence when programming in TeX I tend towards the more implicit style.

rpspringuel:
I'll try it when I get home, but a little closer look at yours and my code reveals that you're right on them not being exactly equivalent.  I would also need to move where `bolextra` gets applied up a few lines in the `bolshift` calculation (to get it before the checks you referred to) to make them strictly equivalent.

rpspringuel:
Okay, I had to fix a few bugs to make things work (so please check to make sure I did those right), but once that's done, the results I'm getting seem to indicate that this is not the correct solution.

First, gabc-output/bugs/fix-243 passes with these changes.  That means they have no effect on a test which I think should fairly clearly end up being different.

These are the tests that do break under these changes:

 - gabc-ouput/bugs/fix-18
 - gabc-output/bugs/fix-284
 - gabc-output/bugs/fix-511
 - gabc-output/glyphs/styles
 - tex-output/bugs/fix-396
 - tex-output/symbols
 - tex-output/variable-height

Most of these tests don't even have an alteration in them, and the one which does (tex-output/bugs/fix-396) doesn't have it at the beginning of a line.

eroux:
aren't the changes in the output related to the fix of the missing `%`? (it should have an impact on almost everything as it changes beginning of lines) if not, I admit i don't know...

rpspringuel:
The changes are not due to adding the missing `%`.  I tried your fix with that removed and I still get the same test failures.

Still, it's not so much the new test failures that bother me (though, admittedly I spent a good deal of my last post enumerating them).  They are small changes that don't really affect the visual impression of the score.

What bothers me is that gabc-output/bugs/fix-243 passes.   I would expect this test to fail if the changes are doing what we want them to do.  After all, it has several lines which begin with alterations and those are the lines we're trying to move.

rpspringuel:
I've done some more work on my solution in #615, which I think solves the problem.

eroux:
ok, closing

henryso:
Can this branch be deleted?

PR #624, opened 2015-09-28
Added the missing liquescent salicus glyphs.

### LINKED ISSUE #622: Undefined control sequence \GreCPSalicusOneTwoDescendens
When typesetting from Nocturnale Romanum 2002, pg. 126, I run into an error with current (but even a month old) gregorio develop:
il(hiOk>)la(hihho~)
With
il(hiOk)la(hihho~)
it is typeset fine, though not the intended shape I need (salicus liquescens, so instead of the upper note note that looks like kv it should look like k>).

henryso:
Fixes #622.

All tests pass, and the salicus test has been updated to include the new glyphs (see gregorio-project/gregorio-test#27)

This is ready for review and merge if satisfactory.


eroux:
ok for me!

PR #625, opened 2015-09-29
Made it so that liquescence on a multi-stropha only affects the marked notes.

### LINKED ISSUE #623: Distropha and tristropha liquescens
If I want a liquescens tristropha, typing (gssgs<) results in all the strophas liquescens, rather than just the last one.  (gss/gs<) works fine, but as I really have no idea when somebody would actually want to write all the 3 (or 2) strophas liquescens rather than the last one, which is pretty common thing when trying to match manuscripts, I wonder if (gssgs<) or perhaps even (gsss<) shouldn't be parsed as liquescence only on the last stropha and if for whatever reason somebody wants all 3 liquescens, write (gs<gs<gs<) or something similar.

henryso:
Fixes #623.

The tests that use bistropha and tristropha obviously break.  See gregorio-project/gregorio-test#28.

This is ready for review and merge if appropriate and satisfactory.


eroux:
ok for me (I think the bug is quite "recent", the new behavior is the correct one indeed.

henryso:
I don't know how "recent" it is, but it's a bug in 3.0, so I'll leave the change log and upgrade notes.

PR #627, opened 2015-09-30
Ignored cscope index files.

henryso:


henryso:
This is just an additional ignore for files generated by `make cscope`.  This should be safe to merge, so I'm merging it.

PR #628, opened 2015-09-30
Fixed errors detected by valgrind: memory leaks and reads of freed memory.

henryso:
I ran valgrind against all the gabc files in gregorio-test and fixed all the memory leaks it found.  Fixing one of the memory leaks (in `close_glyph`) led to valgrind detecting a read of memory after it being freed, so I fixed that as well.

All the tests pass with no changes required.


eroux:
Looks good, thanks a lot!

PR #630, opened 2015-10-01
Added biginitialraise dimension to allow a separate shift for big initials.

henryso:
Since you can style initials separately from big initials, I think the same should be true for the shift.

All the tests pass.

Please review and merge if acceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 
 ### Added
 - New distance, `initialraise`, which will lift (or lower, if negative) the initial.
+- New distance, `biginitialraise`, which will lift (or lower, if negative) the big initial.
 - The first word of the score is now passed to a macro that allow it to be styled from TeX.  The first word is passed to `\GreFirstWord#1` and is styled by changing the `firstword` style.
 - A new type of lyric centering, enabled with `\gresetlyriccentering{firstletter}`, which aligns the neume with the first letter of each syllable.
 - `\greornamentation` allows access to the two ornamentation glyphs.  The ability to access these two glyphs via `{\gregoriosymbolfont \char 75}` was broken by the new interface to the glyphs in greextra.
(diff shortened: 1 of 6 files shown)


eroux:
ok for me

rpspringuel:
I don't like this because it really can lead to a huge proliferation of stuff.  If big initials get their own `initialraise`, shouldn't they also get their own `manualinitialwidth`, `beforeinitialshift`, and `afterinitialshift`?  Personally I'd be in favor of going in the other direction and collapsing the separate styles rather than proliferating the amount of stuff associated with the initials.

Of course, this is at least in part due to a request that Br. Innocent mentioned to me in person once, but which I've yet to take any action on (or even mention, come to think about it): the ability to have 3 and 4 line initials.  If we follow the pattern of having separate variables/styles for each size then implementing this new feature will mean 4 sets of them.  That strikes me as a bit excessive.

henryso:
Ok, let's talk about collapsing the styles.  How you would envision doing that?

rpspringuel:
I would have just one style: `initial`  This would be used for all initials, regardless of size.  By default, we'd probably size it for a 1-line initial, but the user would be free (and indeed expected) to redefine it when the score they're working with has a two-line (or more, if we ever get there) initial.

Implementation wise, in the immediate term I'd simply replace instances of `\gre@style@biginital` by `\gre@style@initial` in the code (likewise for `\endgre...`), I'd remove `biginitial` from the list of defined styles, and I'd add a conditional to `\grechangestyle` which checks to see if the user is attempting to modify `biginitial` and if so emit a deprecation warning while redirecting the modification to `initial`.

I might also include an extra warning in the code which prints a big initial which lets users who might have been satisfied with the defaults (and thus not making changes to `biginitial`) know about the change.

Along similar lines, I'd like the transition between the various sizes of initials to be in the TeX rather than the gabc, much like we migrated the lyric centering decision, and for GregorioTeX to be able to automatically break lines when the initial is 2 (or more) lines large.  However, I realize that this aspect would be a much bigger undertaking (lots more interacting parts to deal with) and would willingly see it set aside as beyond the scope of this particular change.

henryso:
Would we also eliminate gabc's `initial-style: 2`?

henryso:
Please continue this discussion in #632.

PR #633, opened 2015-10-03
Adapted GregorioTeX to the new version of LaTeX and corresponding luatexbase.

### LINKED ISSUE #629: adapt to new LaTeX version
There's a new move in LaTeX to adapt to new engines such as LuaTeX (better late than never!), so a few changes will come with the new version of LaTeX, see [announcement](http://article.gmane.org/gmane.comp.tex.latex.latex3/3683). Gregorio is affected by these changes, especially by the removal of the `luatex` prefix of primitives like `localleftbox` and `localrightbox`, and also by the change in the attribute allocators. As Gregorio is not yet integrated into TeXLive, I think it should adapt to both versions (old and new).

henryso:
Fixes #629.

Caveats:

- This requires the fixes made for josephwright/ltluatexsupp#2 and josephwright/ltluatexsupp#4.
- There is a problem running our tests that I believe is a bug in fontspec (more likely that they have not yet updated for the new LaTeX).

In order to run tests, you need to add the following lines after loading fontspec to work around the fontspec bug (see wspr/fontspec#215):
```
\expandafter\let\csname xetex_suppressfontnotfounderror:D\endcsname=\suppressfontnotfounderror
\expandafter\let\csname luatex_prehyphenchar:D\endcsname=\prehyphenchar
\expandafter\let\csname luatex_posthyphenchar:D\endcsname=\posthyphenchar
\expandafter\let\csname luatex_preexhyphenchar:D\endcsname=\preexhyphenchar
\expandafter\let\csname luatex_postexhyphenchar:D\endcsname=\postexhyphenchar
```

(Note that this workaround uses the new luatex-prefix-less-names in the new LaTeX).

All tests run fine under vanilla TeX Live 2015 (to which I finally upgraded)

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -214,7 +214,8 @@ local function init(arg, enable_height_computation)
   if enable_height_computation then
     new_line_heights = {}
 
-    local mcb_version = luatexbase.get_module_version('luatexbase-mcb')
+    local mcb_version = luatexbase.get_module_version and
+        luatexbase.get_module_version('luatexbase-mcb') or 9999
     if mcb_version and mcb_version > 0.6 then
       luatexbase.add_to_callback('finish_pdffile', write_greaux,
           'gregoriotex.write_greaux')
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -70,13 +70,19 @@
 
 \AtBeginDocument{\IfStrEq{\gre@debug}{}{}{\typeout{GregorioTeX is in debug mode}\typeout{\gre@debug\space messages will be printed to the log.}}}%
 
-% Since TeXLive 2009, the extra primitives of LuaTeX have a luatex prefix, and are not
-% available without prefix. We mostly rely on \directlua that is still available
-% without prefix, but we also rely on the Omega primitives, that have a luatex
-% prefix in TeXLive 2009.
+% Depending on version of TeX / LaTeX, some primitives may or may not have a
+% luatex prefix, so we need to check and handle either
 
-\let\gre@localleftbox\luatexlocalleftbox%
-\let\gre@localrightbox\luatexlocalrightbox%
+\ifdefined\localleftbox%
+  \let\gre@localleftbox\localleftbox%
+\else%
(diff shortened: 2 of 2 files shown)


tex/gregoriotex.lua
```diff
@@ -214,7 +214,8 @@ local function init(arg, enable_height_computation)
   if enable_height_computation then
     new_line_heights = {}
 
-    local mcb_version = luatexbase.get_module_version('luatexbase-mcb')
+    local mcb_version = luatexbase.get_module_version and
+        luatexbase.get_module_version('luatexbase-mcb') or 9999
```
rpspringuel: There's probably some implicit Lua grouping that I'm not aware of, but I think explicit logical grouping should be used to make clear whether this is `a and (b or c)` or `(a and b) or c`.

eroux:
Apart from @rpspringuel 's note, it looks fine, thanks a lot! After TeXLive 2016 is released with Gregorio and the new LaTeX, we may want to use the new LaTeX directly, but this new state is enough for 4.0 I believe.

tex/gregoriotex.lua
henryso: This is a Lua idiom for a ternary operation.  See http://lua-users.org/wiki/TernaryOperator.  I am using the "standard" solution.  Adding grouping would make it seem more like a boolean operation (which it isn't) so I'm inclined to leave it as-is.

eroux: ok for me, this use of `and` and `or` in Lua is really nice, too bad it's not that common in other languages (though it works in most).

henryso: Probably because most other languages actually have a ternary operator or a functional-if and don't need such an idiom.  I've seen shell (which also has neither) use this idiom as well.

rpspringuel: Well, if this is the "standard" solution, I guess it's okay.  Personally I find it to be sloppy to rely on operator precedence for the logical operators because their precedence order isn't as well standardized as the precedence for mathematical operators, but I can live with it.

henryso:
I suppose this is good to merge.

PR #636, opened 2015-10-07
Implemented salicus flexus glyphs.

### LINKED ISSUE #631: Neography issue: Salicus with liquescens note after it connected to it
In Nocturnale Romanum 2002, p. 17, I ran into a neumatic element that I've tried to type as:
non(efOhf>) (but that has the problem that the liquescens note is not connected with the previous note), or as non(efOh!f>) which looks almost right, except that the stem is too long and thus visible below the liquescens note.  I don't have corresponding St. Gall neumes for this antiphon (maybe
it is in the Bud. 118 codex, but I don't have access to that: http://cantusdatabase.org/id/003697 ), so not sure exactly what is it in the original.  The best shape would be like the non(efOh!f>) up to the top note, and then as if it is e.g. non(fhf>) - the top note and the last one.  Would it be possible to add this?
Shall it be typeset for (efOhf>) ?  If user doesn't want to connect them together, he can always type
(efOh/f>) ...

henryso:
Fixes #631.

Please review and merge (into develop) if satisfactory.


eroux:
ok for me!

PR #640, opened 2015-10-09
Added missing void prototype.

henryso:
This fix was proposed by Peter Breitenlohner from the TeX Live team.


henryso:
I have reviewed Peter's patch, and see no reason not to merge it.

PR #643, opened 2015-10-10
Fix up aliases for scandicus and scandicus with episema.

jakubjelinek:
Fix a spelling error in the documentation (resuspinus should have been resupinus).
Fix up tilde handling in the Nabc Reference documentation, so that it is not silently
eaten, or not shown up at all.


jakubjelinek:
I'd like to see this still in 4.0, should I create a pull request against release-4.0 branch instead, or both develop and that branch, something else?

henryso:
@jakubjelinek You should open close this and re-open the pull request against the release-4.0 branch.

PR #644, opened 2015-10-10
Fix up aliases for scandicus and scandicus with episema.

jakubjelinek:
Fix a spelling error in the documentation (resuspinus should have been resupinus).
Fix up tilde handling in the Nabc Reference documentation, so that it is not silently
eaten, or not shown up at all.


henryso:
Looks fine to me.

PR #646, opened 2015-10-12
Automate the insertion of the version number into CHANGELOG.md

### LINKED ISSUE #639: VersionManager and CHANGELOG
It would be really useful if `VersionManager.py` would update CHANGELOG by inserting the version number and date (and a couple of blank lines) below the `## [Unreleased][unreleased]` line.  I often forget to do this as part of prepping a release and end up having to make a separate commit (and move the version tag) as a result (see 8996797, for example).

I can do this myself, I just don't have the time right this minute to implement it.  I'm posting the issue as a reminder to get this done, or so someone else can beat me to it.

rpspringuel:
VersionManager will now automatically insert the appropriate version number and date into CHANGELOG, effectively starting a new Unreleased block
I also noticed that when inserting the date on a PARSE_VERSION_DATE line (i.e. in the documentation) the space between the comma and the date was being lost if a single digit date was replaced by a double digit one.  I just modified the substitution rule a bit to prevent this.

Resolves #639

### DIFF
--- a/VersionManager.py
+++ b/VersionManager.py
@@ -184,7 +184,7 @@ def replace_version(version_obj):
                     result.append(re.sub(r'(\d+\/\d+/\d+)', today.strftime("%Y/%m/%d"), newline, 1))
                 elif 'PARSE_VERSION_DATE' in line:
                     newline = re.sub(r'(\d+\.\d+\.\d+(?:[-+~]\w+)*)', newver, line, 1)
-                    result.append(re.sub(r'([ \d]\d [A-Z][a-z]+ \d{4})', today.strftime("%e %B %Y"), newline, 1))
+                    result.append(re.sub(r'(\d{1,2} [A-Z][a-z]+ \d{4})', today.strftime("%e %B %Y"), newline, 1))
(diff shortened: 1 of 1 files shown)


henryso:
Are you sure about PARSE_VERSION_DATE?  According to the documentation for (C) strftime, `%e` is supposed to be like `%d` but with the leading `0` replaced by a space.  So `%e` *should* produce <space><digit> for single digit dates and <digit><digit> for double digit dates.  If this is true, then your change will cause an accumulation of spaces if we release on single digit dates.

rpspringuel:
I'm sure that before my change we'd lose the space when going from single digit dates to double digit dates (resulting in something like `\large Version \textbf{4.0.0-rc1}, 8 October 2015` becoming `\large Version \textbf{4.0.1},12 October 2015`.  I have to admit I didn't check the other way around (seeing as we're currently a double digit date, that's a little more difficult).

Perhaps `%e` should be replaced by `%-d`?  That should produce a number without either leading 0 or leading space.

henryso:
Sounds good to to me.  Let's use `%-d`.

rpspringuel:
Done.

henryso:
Looks good to me.

PR #647, opened 2015-10-14
Made custos spacing consistent.

### LINKED ISSUE #642: Manual custos and automatic custos (on clef change) have inconsistent spacing
The manual custos of a figure like `(g+::c3)` has additional space behind it compared to `(z0::c3)`.  It is actually coded this way: `\GreManualCustos` adds the space before calling `\GreCustos`.  As such, it works as coded.

Neither I nor @rpspringuel know whether this is the desired behavior or is just some inconsistency that worked its way into the code (and thus would be considered a bug).

Does anyone know the history here?  Is this a bug?

henryso:
Fixes #642.

Tests pass except for the two that exhibit this bug and thus now appear correct.


eroux:
ok for me!

rpspringuel:
I haven't checked the tests yet, but the code changes look good on a simple read through.

PR #649, opened 2015-10-18
Elcapitan

rpspringuel:
This series of commits revamps the scripts associated with the Mac installer significantly to accomplish the following:

1. The installer is now SIP compliant, and will thus work on El Capitan.
2. The installer now recognizes installations of TeXLive which were not done using the MacTeX installer (namely using the MacPorts texlive port and the TeXLive command line tool that is platform independent).
3. The installer scripts log their operations, allowing for an easier diagnosis of what might have gone wrong with a failed installation.

I've also made some adjustments to what is installed to bring it up to 4.0 standards.


henryso:
I'm not really qualified to review the Mac OS stuff, but it seems fine to me.

rpspringuel:
I don't think any of the other regular developers could examine this for functionality (as was pointed out I'm the only one with a Mac).  We probably have to rely on my testing and fix bugs as they are pointed out by users.

On the testing front, I did try the installer out on Mavericks against a TeXLive 2014 install from the command-line tool, on Yosemite against both a MacTeX and MacPorts install of TeXLive 2015, and on El Capitan against both a MacTeX and MacPorts install of TeXLive 2015.  In all cases the installer saw a system which had never had Gregorio installed before and resulted in a working installation.  Do you think there are any other combinations that I should test?  The biggest reason for not testing other combinations is that I don't have ready made snapshots of my VMs in those other conditions.  As it was I had to create the El Capitan VM in its various states in order to test this and that sort of thing takes a while (it is, after all, doing a fresh install of the OS and then installing the rather large TeXLive package a few times in different ways).

The one thing I do think that other developers could comment on intelligently is the coding style of my shell scripts.  This is the only project for which I do any significant shell script coding and so I don't have a whole lot of experience to draw upon when it comes to assessing their readability, etc.

henryso:
OK, then from that point of view, I would protect all directory and file name variables in the shell scripts with double quotes in case there's a space in the name.

rpspringuel:
What about when the variable is part of a string (and thus, in a sense, already inside double quotes)?

henryso:
In a construct such as:
```
MESSAGE="The value of MYVAR = $MYVAR"
```
then the variable is already in double quotes.

Do you mean something else?

rpspringuel:
No, that's exactly what I meant.

rpspringuel:
If no one has any further comments on this I'll merge it.

henryso:
I have no further comment.

PR #651, opened 2015-10-19
Added a case to handle T_TORCULUS_LIQUESCENS (climacus praepunctis deminutus).

### LINKED ISSUE #650: problem with horizontal episema on first note in a torculus that ends with a liquescent
I can't get a neum to like this image

It comes out like this 
image

when I use `(ixh_ihg~)`

It looks like it has been broken since after version 2.3

henryso:
Fixes #650.


eroux:
ok for me!

PR #656, opened 2015-10-23
Implemented a toggle-able forced hyphen for empty initial syllables.

### LINKED ISSUE #653: Stylistic hyphen
When the first word has more than one syllable, `initial-style` is not `0`, and the first syllable consists only of the initial, some users may want a hyphen to appear regardless of whether spacing requires one.

This being a stylistic decision, it should be enabled/disabled from TeX and should default to enabled.

This is not a 4.0.0 deliverable.

henryso:
Fixes #653.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,12 @@
 
 This file contains instructions to upgrade to a new release of Gregorio.
 
+## Unreleased
+
+### Empty first syllable hyphen
+
+Prior to version 4.0, GregorioTeX would put a hyphen below an empty first syllable (which happens when the first syllable consists only of the initial).  The spacing algorithm was improved in 4.0, but that eliminates the hyphen if there is not enough space between the note of the first syllable and the next syllable.  The `\gresetemptyfirstsyllablehyphen` command controls this behavior.  It defaults to `force`, which restores the pre-4.0 behavior, but may be changed to `auto` (i.e., `\gresetemptyfirstsyllablehyphen{auto}`) if you prefer the space-based behavior.
+
 ## 4.0
 
 ### Font changes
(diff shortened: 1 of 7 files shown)


UPGRADE.md
```diff
@@ -2,6 +2,12 @@
 
 This file contains instructions to upgrade to a new release of Gregorio.
 
+## Unreleased
+
+### Empty first syllable hyphen
+
+Prior to version 4.0, GregorioTeX would put a hyphen below an empty first syllable (which happens when the first syllable consists only of the initial).  The spacing algorithm was improved in 4.0, but that eliminates the hyphen if there is not enough space between the note of the first syllable and the next syllable.  The `\gresetemptyfirstsyllablehyphen` command controls this behavior.  It defaults to `force`, which restores the pre-4.0 behavior, but may be changed to `auto` (i.e., `\gresetemptyfirstsyllablehyphen{auto}`) if you prefer the space-based behavior.
```
eroux: Maybe it would be best to document this for 4.0, as it it where the change happened? Thinking about that again, I'm wondering if it's a good idea not to do this hyphen thing for 4.0... what do you think (again)?

eroux:
Apart from the commentary, it looks good

UPGRADE.md
henryso: I'm a little confused from your statement.  Are you saying that we *should* do this change in 4.0 or leave it here in develop?  If you think the scope of these changes is not "too much" for this late in the 4.0 cycle, then I can reopen the pull request against release-4.0.

My only concern with doing it this late in 4.0 is code quality.  Of course, programmers never think they write buggy code, but experience often proves otherwise...

If this is not what you mean, can you clarify?

eroux: That's what I mean indeed, I'm just concerned by this change in 4.0 that might not be easily spotted by users but would "damage" their scores...

henryso:
I will re-open against release-4.0.

PR #657, opened 2015-10-24
Implemented a toggle-able forced hyphen for empty initial syllables.

### LINKED ISSUE #653: Stylistic hyphen

henryso:
Fixes #653.

I changed the wording a bit from #656 since this is now part of 4.0.

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -790,15 +790,22 @@ Macro to determine whether the last line of the score should be justified or not
 \subsubsection{Hyphenation}
 
 \macroname{\textbackslash gresethyphen}{\{\#1\}}{gregoriotex-main.tex}
-Tell Gregorio\TeX\ to when to place a hyphen between
-syllables in polysyllabic words in a score.  This is done by overriding \texttt{maximumspacewithoutdash} so subsequent changes to this dimension will override this command.
+Tells Gregorio\TeX\ how to place a hyphen between syllables in polysyllabic words in a score.  This is done by overriding \texttt{maximumspacewithoutdash} so subsequent changes to this dimension will override this command.
 
 \begin{argtable}
   \#1 & force & Hyphens will appear between all syllables in polysyllabic words.\\
-  & auto & Hyphens will appear based on the setting of \texttt{maximumspacewithoutdash}
+      & auto  & Hyphens will appear based on the setting of \texttt{maximumspacewithoutdash} (default)
 \end{argtable}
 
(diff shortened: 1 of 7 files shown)


doc/Command_Index_User.tex
```diff
@@ -790,16 +790,23 @@ \subsubsection{End of Line Behavior}
 \subsubsection{Hyphenation}
 
 \macroname{\textbackslash gresethyphen}{\{\#1\}}{gregoriotex-main.tex}
-Tell Gregorio\TeX\ to when to place a hyphen between
-syllables in polysyllabic words in a score.  This is done by overriding \texttt{maximumspacewithoutdash} so subsequent changes to this dimension will override this command.
+Tells Gregorio\TeX\ how to place a hyphen between syllables in polysyllabic words in a score.  This is done by overriding \texttt{maximumspacewithoutdash} so subsequent changes to this dimension will override this command.
```
rpspringuel: I realize that this is old code that you haven't actually changed, but it got me thinking.
Given that you've introduced a flag `\ifgre@forcehyphen` my naive expectation would be that this should be used here instead of altering `maximumspacewithoutdash`.

Perhaps your flag should be called `\ifgre@showhyphen` instead, in which case we could revamp the way this setting works so that it doesn't have to alter `maximumspacewithoutdash` (but that sort of optimization need not be part of this change).

henryso: Let's tackle that as a separate issue.

It would be a little more complex than what you mention.  I introduced that flag so I could have an "or" condition that causes the hyphen to appear: either it's the first initial and the hyphen is forced or the spacing requires a hyphen.  As such, this flag is reset by \GreSyllable.  We could use it for `\gresethyphen`, which should then need to set its own flag (rather than adjusting `maximumspacewithoutdash`) which would then be used as an additional clause of the above "or" condition.

rpspringuel: I agree that reworking `\gresethyphen`'s internals would be a separate issue.  I'm just saying that the name `\ifgre@forcehyphen` makes me think you're doing what that issue would address, rather than what you're actually addressing here.

Basically the only issue I have right now is in your choice of names, but if you want to stick with it as is, I can confirm that everything works as expected and is ready for merge.

henryso:
@rpspringuel Thanks for clarifying.  I renamed the flag.

rpspringuel:
You missed the occurance in the documentation.

henryso:
Fixed.  Thanks.

PR #659, opened 2015-10-25
Windowsfileversion

rpspringuel:
Windows Installers key off the `FILEVERSION` variable in the fixed-info section of the resource file to determine whether or not to replace a file when installing an upgrade.  This commit updates that variable (and `PRODUCTVERSION` in the same block, for good measure) to match the current version number.
It also adds rules to VersionManager.py for updating those variables correctly.  For these binary version numbers, all four components must be (16-bit) integers; `-beta` and `-rc` are not allowed.  To accommodate this, I've mapped the fourth component of the binary version number as follows:
  - `-beta1` to `-beta9` map to `11` to `19`
  - `-rc1` to `-rc9` map to `21` to `29`
  - all official releases have a fourth component of `30`

In this fashion, this component will increase as we progress towards a release.
Note: This mapping does limit us to 9 betas and 9 release candidates, but I think that's enough.
I also made it so that newly created betas are tagged as `-beta1` and added a new option to VersionManager.py for incrementing beta numbers.


eroux:
Ok for me, thanks a lot!

henryso:
Seems fine to me.

PR #665, opened 2015-10-31
Fixed lost "endgroup" for first letter centering of empty syllables.

henryso:
This is an edge case where the first syllable is empty (or has only the initial when initials are enabled).  The fifth argument to `\GreSetThisSyllable` was lost.  This corrects that bug.


rpspringuel:
Checks out for me.

PR #678, opened 2015-11-26
Corrected end-of-word detection when there is no whitespace at end-of-file.

### LINKED ISSUE #675: Problem with no newline after a single-syllable word on the last line
If a file has no newline on the last line and the last word is a one-syllable word, the syllable is incorrectly detected as `WORD_BEGINNING` rather than `WORD_ONE_SYLLABLE`.  If there *is* a newline after the last word, then the syllable is correctly detected as `WORD_ONE_SYLLABLE`.

I'm not sure this is a problem in any real-world situation, but it is something peculiar I noticed while debugging some tests.

henryso:
Fixes #675.

Tests pass once the affected test expectation is updated.
 
### DIFF
--- a/src/gabc/gabc-score-determination.l
+++ b/src/gabc/gabc-score-determination.l
@@ -25,12 +25,14 @@
 #include <string.h>
 #include "struct.h"
 #include "messages.h"
+#include "bool.h"
 
 #include "gabc.h"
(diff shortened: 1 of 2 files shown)


eroux:
ok for me!

PR #681, opened 2015-11-29
Simplify and fix gprocess to work with 4.0.0.

avbop:
The gprocess script wasn't working with 4.0.0-rc1, but this should fix it. I also pulled out the PDF viewer code, as it seemed unnecessary.


henryso:
Looks good to me.

Issue #663, opened 2015-10-27
Crunching space around alterations mid-line


rpspringuel:
Much like alterations which lead a line, it should be possible for alterations which lead a syllable mid-line to invade the space of the previous syllable so long as a few conditions are met:

1. The alteration extends out to the left of the lyrics of its own syllable (i.e. the syllable has a positive `begindifference`)
2. The previous syllable had lyrics which extended beyond the end of its notes (i.e. `previousendifference` is positive)
3. The space between the notes of the previous syllable and the alteration is at least `spacebeforealteration`.

PR #686, opened 2015-12-03
attempt to fix 663

eroux:
#663 is clearly a bug to me, this is an attempt at fixing it. Before going further, I'd like to know your opinion on two points:

- I think it should be fixed for 4.0, the core of the fix (line 295 of gregoriotex-syllable.tex) is quite small
- I changed alterationspace to be a dimen instead of a skip, as I think everything involved in nextbegin different should be a dimen: if a skip gets much tighter than expected, the result will be quite bad in terms of spacing of the text of two syllables of the same word: if they are close, they might even sometimes collapse a little, but this change may be more relevant in 4.1...

Also, the diff is not really readable due to my text editor automatically removing some spurious spaces, sorry!

### DIFF
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -37,7 +37,7 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %% macros for discretionaries
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-% 
+%
 % In order to avoid clef change at beginning or end of line, we use discretionaries
 % for clef change, or even with more complex data (z0::c3 for instance). The problem
 % with discretionaries is that:
@@ -154,7 +154,7 @@
 }%
 
 % macro that typesets the key
-% arguments are : 
+% arguments are :
 %% #1: the type of the key : c or f
 %% #2: the line of the key (1 is the lowest)
 %% #3: if we must use small key characters (inside a line) or not 0: if not inside, 1 if inside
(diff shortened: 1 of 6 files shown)


tex/gregoriotex-signs.tex
```diff
@@ -1664,14 +1664,14 @@
...
-    \gre@skip@temp@four = \gre@skip@alterationspace\relax%
+    \gre@skip@temp@four = \gre@dimen@alterationspace\relax%
     \ifgre@firstglyph%
       \gre@debugmsg{bolshift}{making adjustments for leading alteration}%
       \global\advance\gre@dimen@notesaligncenter by \wd\gre@box@temp@width %
       \global\advance\gre@dimen@notesaligncenter by \dimexpr\gre@skip@temp@four\relax %
       \kern\gre@skip@temp@four %
       \global\gre@dimen@bolextra = \wd\gre@box@temp@width%
-      \gre@skip@temp@four = \gre@skip@alterationspace\relax%
+      \gre@skip@temp@four = \gre@dimen@alterationspace\relax%
```
rpspringuel: Not your fault, but I don't think this line is needed at all.  It's a repeat of line 1667 and `\gre@skip@temp@four` hasn't been changed in the half-dozen lines in between.

rpspringuel:
If the fix really is as small as you indicate, then I'm not adverse to putting this into 4.0.

`alterationspace` isn't involved in the calculation of `nextbegindifference` until after your change.  I think the more pertinent question in deciding to make the change is whether the space between a flat (or sharp or natural) and a note should be rubber (as that's what `alterationspace` is used for).  Personally, I have no problem with that being a fixed distance as when TeX makes it's justification adjustments I'd prefer that alterations not be separated from their note groups.

Also, what about this fix still fails?  The test you added seems to indicate a pass.

henryso:
Question: why are there merges from release-4.0 into develop as part of this change request?  Is that some weird github thing?

rpspringuel:
They're there because @eroux accidentally branched from develop instead of release-4.0.  That obviously needs to be fixed.  Good catch as I hadn't noticed that until you mentioned it.

rpspringuel:
For anyone looking at this PR, I just wanted to note that it was closed, not merged.  I don't know why GitHub is showing otherwise, but examining the history clearly shows that this was not merged.

henryso:
Please see #688 for the rebasing of this pull request against the release-4.0 branch.

PR #696, opened 2015-12-08
Applied fixes suggested by Karl Berry for integration into TeX Live.

henryso:
Karl Berry's email:
```
    https://github.com/gregorio-project/gregorio/blob/release-4.0/src/gregorio-utils.c

Thanks for the url.  From looking at the first page of the file:

0) A comment on define_path saying what it is supposed to do would be
nice, though not a security issue :).  (And a one-liner at the top of
the file with the 10000m description of the whole program.)

1) you must always check the return status of malloc/realloc/whatever.
I imagine you don't want to depend on kpse/gnu xmalloc/xrealloc/whatever
being available, but you should make your own trivial wrappers.

2) I don't think exit(-1) is a good idea.  exit(1) is simpler and more
reliable to mean "just an error".  That's what I do in kpse and what
pretty much everything else does, as far as I've noticed.  (Unless
you're worried about VMS :), which is the only system where 1=success, afaik.)

3) Allocating something the size of PATH_MAX is also not a good idea.
That can be huge, or conversely, not enough.  (One of rms's first
explicit technical goals for GNU was to place no limit on file name
length.)  It is better to (x)malloc the actual sizes that are needed.

4) In particular, the very first strcpy is, as far as I can see, a
security risk:
  char temp_name[PATH_MAX];
  ...
  strcpy(temp_name, string);

I see nothing guaranteeing that strlen(string) < PATH_MAX at that
point.  Therefore you're potentially overwriting the stack.

Better to follow the pattern as in your get_output_filename:
1) find length that is needed;
2) xmalloc that length;
3) strcpy the bytes.

Enough for today ...

Happy hacking,
Karl
```

0) I added the comment for define_path and file-descriptive comments to all the source files.
1) I created gregorio_malloc, gregorio_calloc, gregorio_realloc, and gregorio_strdup as wrappers to the allocation functions; these log a fatal error and exit the program if any allocation attempt fails.
2) I changed all exit(-1) to exit(1)
3) I eliminated some dependencies on PATH_MAX, but there are still two instances where it's used.  In one case, we pass such a variable to fgets, which doesn't do allocation. In the second case, we pass it to getcwd, which if we are sticking to POSIX (as opposed to GNU), does not allocate a buffer.
4) I implemented as specified

I also ran the code through valgrind and found an unclosed file (error_file), so I fixed that as well.

Please review this and merge if satisfactory.  I am also open to suggestions for eliminating the final two uses for item 3.

### DIFF
--- a/src/bool.h
+++ b/src/bool.h
@@ -1,4 +1,7 @@
 /*
+ * Gregorio is a program that translates gabc files to GregorioTeX
+ * This header provides a minimum of C11-like bool functionality.
+ *
  * Copyright (C) 2015 The Gregorio Project (see CONTRIBUTORS.md)
  * 
  * This file is part of Gregorio.
--- a/src/characters.c
+++ b/src/characters.c
@@ -1,4 +1,7 @@
 /*
+ * Gregorio is a program that translates gabc files to GregorioTeX.
+ * This file contains functions that deal with lyrics and styles.
+ *
  * Copyright (C) 2008-2015 The Gregorio Project (see CONTRIBUTORS.md)
  *
  * This file is part of Gregorio.
@@ -42,6 +45,7 @@
 #include "unicode.h"
 #include "characters.h"
 #include "messages.h"
+#include "support.h"
 #include "utf8strings.h"
 #include "vowel/vowel.h"
 
@@ -84,11 +88,7 @@ static bool read_vowel_rules(char *const lang) {
     char buf[PATH_MAX];
     size_t capacity = 16, size = 0;
     
-    if (!(filenames = malloc(capacity * sizeof(char *)))) {
-        gregorio_messagef("read_patterns", VERBOSITY_FATAL, 0,
-                _("unable to allocate memory"));
-        return false;
-    }
+    filenames = gregorio_malloc(capacity * sizeof(char *));
 
     file = popen("kpsewhich " VOWEL_FILE, "r");
     if (!file) {
@@ -100,15 +100,11 @@ static bool read_vowel_rules(char *const lang) {
     while (!feof(file) && !ferror(file) && fgets(buf, PATH_MAX, file)) {
         rtrim(buf);
         if (strlen(buf) > 0) {
-            filenames[size++] = strdup(buf);
+            filenames[size++] = gregorio_strdup(buf);
             if (size >= capacity) {
                 capacity <<= 1;
-                if (!(filenames = realloc(filenames,
-                                capacity * sizeof(char *)))) {
-                    gregorio_messagef("read_patterns", VERBOSITY_FATAL, 0,
-                            _("unable to allocate memory"));
-                    return false;
-                }
+                filenames = gregorio_realloc(filenames,
+                                capacity * sizeof(char *));
             }
         } else {
             gregorio_messagef("read_patterns", VERBOSITY_WARNING, 0,
@@ -216,7 +212,7 @@ static void determine_center(gregorio_character *character, int *start,
     if (count == 0) {
         return;
     }
-    subject = (grewchar *)malloc((count + 1) * sizeof(grewchar));
+    subject = (grewchar *)gregorio_malloc((count + 1) * sizeof(grewchar));
     for (count = 0, ch = character; ch; ch = ch->next_character) {
         if (ch->is_character) {
             subject[count ++] = ch->cos.character;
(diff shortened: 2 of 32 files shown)


eroux:
Good job! About item 3, maybe you can answer to Karl directly with this question? I think this PR can be for 4.0.0, but we can wait for a 4.0.1 to upload to ctan.

jakubjelinek:
On Mon, Dec 07, 2015 at 06:03:44PM -0800, Henry So wrote:
> Thanks for the url.  From looking at the first page of the file:
> 
> 0) A comment on define_path saying what it is supposed to do would be
> nice, though not a security issue :).  (And a one-liner at the top of
> the file with the 10000m description of the whole program.)
> 
> 1) you must always check the return status of malloc/realloc/whatever.
> I imagine you don't want to depend on kpse/gnu xmalloc/xrealloc/whatever
> being available, but you should make your own trivial wrappers.
> 
> 2) I don't think exit(-1) is a good idea.  exit(1) is simpler and more
> reliable to mean "just an error".  That's what I do in kpse and what
> pretty much everything else does, as far as I've noticed.  (Unless
> you're worried about VMS :), which is the only system where 1=success, afaik.)
> 
> 3) Allocating something the size of PATH_MAX is also not a good idea.
> That can be huge, or conversely, not enough.  (One of rms's first
> explicit technical goals for GNU was to place no limit on file name
> length.)  It is better to (x)malloc the actual sizes that are needed.
> 
> 4) In particular, the very first strcpy is, as far as I can see, a
> security risk:
>   char temp_name[PATH_MAX];
>   ...
>   strcpy(temp_name, string);
> 
> I see nothing guaranteeing that strlen(string) < PATH_MAX at that
> point.  Therefore you're potentially overwriting the stack.

Have to agree on all of these.

> Better to follow the pattern as in your get_output_filename:
> 1) find length that is needed;
> 2) xmalloc that length;
> 3) strcpy the bytes.

I'd add that once you know the length, you can as well use memcpy instead.
While gcc is able to optimize this in certain cases itself, the gregorio_malloc
wrapper is likely going to hide it.
So instead of ... gregorio_malloc (... strlen (something) + ... );
strcpy (..., something);
you can do:
  size_t something_len = strlen (something);
  ... gregorio_malloc (... something_len + ... );
  memcpy (..., something, something_len + 1);
or similar.

	Jakub

eroux:
Thanks @jakubjelinek , also what do you think about the question of @henryso on the point `3)` of the description of this PR?

jakubjelinek:
3) I eliminated some dependencies on PATH_MAX, but there are still two instances where it's used.  In one case, we pass such a variable to fgets,  which doesn't do allocation. In the second case, we pass it to getcwd, which if we are sticking to POSIX (as opposed to GNU), does not allocate a buffer.                                                                                                                                           

For the fgets case, you are basically looking for getline function, which used to be a GNU extension and recently is in POSIX too.  See
http://pubs.opengroup.org/onlinepubs/9699919799/functions/getdelim.html
For portability, I'm afraid you can't count on it, so either you need to use it only conditionally (based on whether configure detects it) and have some fallback, or just write your own version of it (doesn't matter if it is less efficient than the C library one).  For the fallback or your own version, it is pretty much about just allocating some initial length of the buffer, fgets into it, and if the fgets buffer is not terminated with a new-line, realloc a larger buffer and keep calling fgets further until the newline is read into the end of the buffer or end of file or error is signalled on the stream.

As for getcwd, again, your choice to use conditionally the GNU extension, or just always use a fallback - start with some reasonably sized buffer and if getcwd returns NULL and errno == ERANGE, reallocate it to twice the size and so on.  The first getcwd can be of course to a static/automatic buffer of reasonable size, and only if that fails you allocate from heap.

eroux:
Thanks a lot! by the way, @jakubjelinek do you think you could help us audit the code in terms of security? I know C but have never used it except for Gregorio, so I really don't have the ability to do this... I'm sure you can spot many things easily (doesn't have to be a deep analysis right now). The goal is to have gregorio in the `shell_escape_commands` of TeXLive (which is a first step towards having Gregorio usable easily in Overleaf or ShareLaTeX), and to do this we need Gregorio not to trigger a red light in the head of an experienced C developer looking at the code... And this is something I cannot do by myself... Or maybe you can point to some good best practices guide?

eroux:
Please continue the discussion on #697

PR #700, opened 2015-12-09
Develop

rpspringuel:
There were two conflicts: CHANGELOG.md and src/gabc/gabc-score-determination.l

Please double check my reconciliations (especially on the latter).

gabc-gabc/glyphs/horizontal_episemus.gabc fails, but this appears to be because the gabc source was modified in b5a78b0 but the gabc-gabc expectation was not updated (though the gabc-dump and gabc-output expectations were).  All other tests pass.

### DIFF
--- a/.gregorio-version
+++ b/.gregorio-version
@@ -1,4 +1,4 @@
-4.0.0-rc2
+4.0.0
 
 *** Do not modify this file. ***
 Use VersionManager.py to change the version.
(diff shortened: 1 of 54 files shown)


henryso:
The gabc-gabc expectations didn't *exist* in 4.0, so the merge updated the source file but of course not the expectation.

henryso:
The conflicted files you mentioned look fine.

rpspringuel:
> The gabc-gabc expectations didn't exist in 4.0, so the merge updated the source file but of course not the expectation.

Then why didn't this problem occur for all the gabc-gabc tests?

henryso:
Because I have been meticulously merging 4.0 into develop.  However, there wasn't enough time for me to get to that for these latest changes.

Issue #697, opened 2015-12-08
Improve security (in C code)


eroux:
The goal here is to have the command `gregorio` accepted in the `shell_escape_commands` of TeXLive, which means there will be a security audit. Once gregorio is in the `shell_escape_commands`, it will be more simple and safe for users (no need of `--shell-escape` switch), and it will also make Gregorio work on online platforms such as [overleaf](https://www.overleaf.com/) or [ShareLaTeX](https://fr.sharelatex.com/), lowering the barriers for new users.

We can continue the discussion started on #696 here, and PR related to security can reference this issue.

eroux:
(note that it is not blocking for 4.0.0 release)

eroux:
I've tried a few things:

- [SEI CERT standards](https://www.securecoding.cert.org/confluence/display/c/SEI+CERT+C+Coding+Standard) look promising but it's quite a lot of work and might not be that relevant...
- [splint](http://www.splint.org/) is very easy but outputs so many errors it certainly has to be tamed a bit, but it's hard for me to know which warnings should be disabled
- [frama-c](http://frama-c.com/) is really not easy to work with...

Anyway, non of these said anything about `PATH_MAX`, so even if we follow them strictly, it wouldn't be enough...

henryso:
Are TeX Lives security audit rules documented anywhere?  Maybe we can start there?  To be honest, I need some guidance how to proceed with this.

eroux:
There's no documentation about this... and I'm not really able to give any guidance... @jakubjelinek would you have a few ideas?

henryso:
I think the merge of release-4.0 into master auto-closed this issue, so I'm re-opening it.

henryso:
Would it be possible to get gregorio into TeX Live without getting it into `shell_escape_commands`, in case we are unable to get this fixed before the cut-off date for TeX Live 2016?

eroux:
Yes of course, but I think it would help many people to have it in `shell_escape_commands`... But if we can't deal with it before, we'll do when TeXLive 2016 will start to build (around February or March I think).

kberry:
i've never heard of any of the things you found (good job looking).

there is no hard and fast rule, since creating a "secure" program remains an unsolved and unsolvable problem, even in theory.

basically, the program should obey the kpathsea settings for reading/writing (already done i believe), and shouldn't crash.
two obvious things are don't overflow arrays (under any circumstances) and
[...]

PR #701, opened 2015-12-09
Eliminated dependency on PATH_MAX.

henryso:
For #697.

This is implemented according to @jakubjelinek's suggestion in #696 (growing buffers).

There is still a dependency on _MAX_PATH, because something needs to be passed for the *maxLength* argument for the Win32 **_fullpath** call.

However, quoting https://msdn.microsoft.com/en-us/library/506720ff(v=vs.100).aspx:

If the *absPath* buffer is **NULL**, **_fullpath** calls **malloc** to allocate a buffer and ignores the *maxLength* argument.  (All our usages of this will pass **NULL** as *absPath*)

Note: I re-created the release-4.0 branch from the v4.0.0 tag to be used for 4.0.x releases.

All tests pass.  Valgrind detects no leaks against gabc-gtex and gabc-dump tests (gabc-gabc tests don't exist in 4.0), and everything allocated is freed at the end.  Please review and merge if satisfactory.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -49,12 +49,6 @@
 #include "utf8strings.h"
 #include "vowel/vowel.h"
 
-#ifdef _WIN32
-#ifndef PATH_MAX
-#define PATH_MAX _MAX_PATH
-#endif
-#endif
-
 #define VOWEL_FILE "gregorio-vowels.dat"
 
 #ifndef USE_KPSE
@@ -85,7 +79,8 @@ static bool read_vowel_rules(char *const lang) {
     }
 #else
     FILE *file;
-    char buf[PATH_MAX];
+    int bufsize = 128;
+    char *buf = gregorio_malloc(bufsize * sizeof(char));
     size_t capacity = 16, size = 0;
     
     filenames = gregorio_malloc(capacity * sizeof(char *));
@@ -97,7 +92,17 @@ static bool read_vowel_rules(char *const lang) {
                 strerror(errno));
         return false;
     }
-    while (!feof(file) && !ferror(file) && fgets(buf, PATH_MAX, file)) {
+    for (buf[bufsize - 2] = '\n';
+            !feof(file) && !ferror(file) && fgets(buf, bufsize, file);
+            buf[bufsize - 2] = '\n') {
+        while (!feof(file) && !ferror(file) && buf[bufsize - 2] != '\n'
+                && bufsize < MAX_BUF_GROWTH) {
+            int oldsize = bufsize;
+            bufsize <<= 1;
+            buf = (char *)gregorio_realloc(buf, bufsize * sizeof(char));
+            buf[bufsize - 2] = '\n';
+            fgets(buf + oldsize - 1, oldsize + 1, file);
+        }
         rtrim(buf);
         if (strlen(buf) > 0) {
             filenames[size++] = gregorio_strdup(buf);
@@ -111,6 +116,7 @@ static bool read_vowel_rules(char *const lang) {
                     _("kpsewhich returned bad value for %s"), VOWEL_FILE);
         }
     }
+    free(buf);
     filenames[size] = NULL;
     pclose(file);
 #endif
--- a/src/characters.c
+++ b/src/characters.c
@@ -79,8 +79,8 @@ static bool read_vowel_rules(char *const lang) {
     }
 #else
     FILE *file;
-    int bufsize = 128;
-    char *buf = gregorio_malloc(bufsize * sizeof(char));
+    size_t bufsize = 0;
+    char *buf = NULL;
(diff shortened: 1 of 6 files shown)


eroux:
This looks fine, but maybe you could add a comment in the code with your note about the _fullpath call in the PR description?

jakubjelinek:
+    int bufsize = 128;
Please use size_t instead of int for variables holding sizes.
+    char *buf = gregorio_malloc(bufsize * sizeof(char));
sizeof(char) is 1 by definition, so it is both unnecessary and misleading to add * sizeof(char), bufsize is counted in the smallest addressable units (chars).
+    for (buf[bufsize - 2] = '\n';
This is a neat trick, but it won't work if you have a line where buf[bufsize - 3] will be set to '\n' and buf[bufsize - 2] to '\0' - your logic will consider that because it changed your '\n' to something else you must read further.
As you call strlen on the whole string after the fgets loop, one option is also to compute that strlen earlier and use that to find if the buffer ends with '\n' before the '\0' (of course you'd want to start from the position where you called fgets last time, otherwise it would be O(n^2) complexity.
+            int oldsize = bufsize;
+            bufsize <<= 1;
This is risky, at least on 64-bit architectures and somebody trying to exploit the program you could end up with negative bufsize this way.  I'd suggest using size_t and leaving the loop if bufsize is larger than SIZE_MAX / 2 or so (with an error).
+            fgets(buf + oldsize - 1, oldsize + 1, file);
You aren't checking if this fgets has not returned NULL.
Also, I'd suggest you just introduce gregorio_getline function with the same API as POSIX getline, do the fgets in there and use it, then it would be more readable what you are actually trying to do.

henryso:
I'm not disputing any of these points, but I have a question about your second-to-last point, which I realize becomes moot once I switch to `size_t`.  I am leaving the loop once `bufsize` reaches `MAX_BUF_GROWTH` (currently `((INT_MAX >> 1) + 1)`).  How is it possible to get a negative bufsize in this case?

jakubjelinek:
Oops, sorry, you're right.  I saw that in the second changed file, but missed in the first one.  That said, the error handling once you pass the first condition (first fgets returns non-NULL) looks weird, even later ferror or fgets returning NULL or reaching the too large bufsize should result in the current buf not being considered as a filename (break out of the outer loop).

henryso:
SIZE_MAX is C99, so I can't use that.  I'm thinking of setting an arbitrary maximum buffer size like 32kb or something,  Is 32kb too large for size_t on some (relatively) modern architecture?

henryso:
I think I've addressed everything brought up so far.  I ended up setting MAX_BUF_GROWTH to 32kb.  Let me know if there's a better number I can use.

jakubjelinek:
32kb is too limiting.  size_t must be unsigned type, so SIZE_MAX must be equivalent to ~(size_t) 0,
so you can just use (~(size_t) 0) >> 1 as the limit.  Though, both fgets and getcwd take int as argument, so you also need to break out of the loop if (size_t) (int) size != size.  So just use INT_MAX as the limit?

henryso:
I did think of using `INT_MAX` (if you recall, I was using `INT_MAX`), but research into `size_t` led to the point that size_t may be smaller than int, the example being MS-DOS, where size_t was something like 64kb.

henryso:
Practically speaking, do we really need to deal with path names that are larger than 32kb?

henryso:
I think the value for MAX_BUF_GROWTH in 471be6a should be reasonably portable.  Any other suggestions?

henryso:
Any objections to a merge?

eroux:
Ok for me, but I'm not the expert here!

rpspringuel:
This fails to build on my system with the following error:

```
In file included from gregorio-utils.c:43:
./support.h:30:9: error: 'SIZE_T_MAX' macro redefined
      [-Werror,-Wmacro-redefined]
#define SIZE_T_MAX (~((size_t)0))
        ^
/usr/include/i386/limits.h:98:9: note: previous definition is here
#define SIZE_T_MAX      ULONG_MAX       /* max value for a size_t */
        ^
1 error generated.
```

henryso:
@rpspringuel I think this should fix your error.

rpspringuel:
Builds now, but I'm getting test failures.  Probably because of the bugfixes which have been implemented on release-4.0 since this branch was started.  I'm retesting at the moment after merging in all the release-4.0 changes.

rpspringuel:
Tests complete and all pass. If you're satisfied with the code (I'm no expert there), then go ahead and merge.

henryso:
@jakubjelinek Does this satisfy your checks?

henryso:
This has been sitting here long enough.  It passes the tests, and it's at least not worse than the old code, so I guess I'll merge it.

PR #702, opened 2015-12-10
Obsoletion

rpspringuel:
Techinically we don't need to do this until it's time to prepare a 4.1 release, but I figured it would be easier to find and identify the deprecated and obsolete code that needed to be progressed now rather than later.

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -1634,22 +1634,6 @@
   \relax%
 }%
 
-\def\includetexscore#1{%
-  \gre@obsolete{\protect\includetexscore}{\protect\gregorioscore[n]}%
-}%
-
-\def\greincludetexscore#1{%
-  \gre@obsolete{\protect\greincludetexscore}{\protect\gregorioscore[n]}%
-}%
-
-\def\includegabcscore#1{%
-  \gre@obsolete{\protect\includegabcscore}{\protect\gregorioscore[c]}%
-}%
-
-\def\greincludegabcscore#1{%
-  \gre@obsolete{\protect\greincludegabcscore}{\protect\gregorioscore[c]}%
-}%
-
 % The internal macro called when \gregorioscore is called with the optional argument.  Behavior is determined by the value of the argument:
 % n - do not attempt to compile the score.  Simply include it as is.
 % a - perform the checks to see if the score needs to be recompiled and do so only if necessary
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -1488,7 +1488,7 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 \def\grechangestyle#1#2{%
-  \IfStrEq{#1}{biginitial}{\gre@obsolete{biginitial style}{initial style}}{}%%% DEPRECATED for removal in 5.0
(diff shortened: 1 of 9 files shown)


rpspringuel:
Oh, all tests pass with no changes (as expected).

tex/gregoriotex-main.tex
```diff
@@ -1537,7 +1488,7 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 \def\grechangestyle#1#2{%
-  \IfStrEq{#1}{biginitial}{\gre@deprecated{biginitial style}{initial style}}{}%%% DEPRECATED for removal in 5.0
+  \IfStrEq{#1}{biginitial}{\gre@obsolete{biginitial style}{initial style}}{}%%% DEPRECATED for removal in 5.0
```
henryso: This is not obsolete yet.  The deprecation of biginitial style was introduced in develop.

tex/gregoriotex-main.tex
```diff
@@ -1703,7 +1633,7 @@
 % Macros for compiling short snippets of GABC directly expressed in TeX
 
 \long\def\gre@gabcsnippet@option[#1]#2{%%% DEPRECATED by 4.1
-  \gre@deprecated{\protect\gabcsnippet  with initial-style argument}{\protect\gresetinitiallines}%
+  \gre@obsolete{\protect\gabcsnippet  with initial-style argument}{\protect\gresetinitiallines}%
```
henryso: This is also not obsolete yet.  Deprecation of initial-style was introduced in develop.

tex/gregoriotex-main.tex
```diff
@@ -405,86 +402,81 @@
...
-        \endgre@style@initial% keep this line
-      \fi%%% DEPRECATED for removal in 5.0
-    \else%  DEPRECATED
-      \gre@deprecated{\protect\grebiginitialformat}{\protect\grechangestyle{biginitial}}%  DEPRECATED
-      \grebiginitialformat{\global\gre@dimen@temp@four = \gre@dimen@afterinitialshift}%  DEPRECATED
-    \fi%  DEPRECATED
+    \fi%%% DEPRECATED for removal in 5.0
+    \ifx\gre@empty@biginitialformat\grebiginitialformat% OBSOLETE
+    \else%  OBSOLETE
+      \gre@obsolete{\protect\grebiginitialformat}{\protect\grechangestyle{biginitial}}%  OBSOLETE
```
henryso: The way github shows this diff makes it a little difficult to see if it's really correct.  It *looks* correct, but you should double check that the deprecated usage of \gre@style@biginitial still works.

rpspringuel: I checked this using an old score with a big initial that I had and it worked.

rpspringuel: Along those lines, do you think it would be a good idea to add another class of tests which specifically checks for backwards compatibility with deprecated functions?  I know I changed all the current tests to prohibit the use of deprectated functions, but a separate class (where tests go to die) might help make sure we keep things backwards compatible.

henryso: That sounds like a good idea.  It needs a bit of thought, though, so I'll create an issue so it's not forgotten.

henryso:
This seems fine.

PR #711, opened 2015-12-12
Fix 683 bolnoclef

### LINKED ISSUE #683: BOL Shifts interact with invisible clefs
I noticed this as I was preparing the revised summary, but when the automatic clefs are set to be invisible, the bolshift causes the text to stick out into the margin.  This should not happen.  There are two ways to implement this:

1. The easy way: when `\gresetclef{invisible}` is called, it automatically calls `\gresetbolshifts{disabled}`.  The problem is that what to do when `\gresetclef{visible}` is called later.
2. The less easy way: Add a flag check to the bolshift calculation that checks to make sure clefs are visible before calculating the bolshift.  If clefs are invisible, then bolshift is automatically 0.

rpspringuel:
This fixes #683.  Now, when there are no beginning of line clefs, there is a gap of `noclefspace` at the beginning of the line and then the syllable will start immediately (either lyrics or notes, which ever sticks out to the left further).

This PR is against develop, because that's where I branched from in my repository when starting this work.  However, looking back, we had originally discussed this as a bug in 4.0.  Should I take the time to shift these changes over to build off that branch?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+### Fixed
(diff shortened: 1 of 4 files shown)


henryso:
If you can, yes.  This would eliminate the known issue.

rpspringuel:
See #712

PR #712, opened 2015-12-12
Fixing bol problems

### LINKED ISSUE #683: BOL Shifts interact with invisible clefs

rpspringuel:
`noclefspace` is now applied to the first line of the score, bringing it into line with the rest of the score
`bolshift` is disabled when there are no beginning of line clefs

Fixes #683 as a patch release for 4.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ## [Unreleased][unreleased]
 ### Fixed
 - `\greseteolcustos` now retains its setting across multiple score inclusions (see [#703](https://github.com/gregorio-project/gregorio/issues/703)).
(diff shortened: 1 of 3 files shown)


henryso:
Just a question about this.  What do we do about the Known Bugs section once this goes in?  Do we leave it there?

rpspringuel:
Hmm... Hadn't thought about that.

I don't think we should delete it entirely, for the same reason that yanked releases still have an entry in the CHANGELOG, but perhaps we should explicitly mark the bug as fixed (indicating which version fixed the bug).

henryso:
Sounds like a good idea.  I guess you'll make that change?

henryso:
Looks good.

PR #714, opened 2015-12-14
Reversed the naming of the patty cross ("Maltese") and the patonce cross

### LINKED ISSUE #713: \grecross and \grealtcross have been swapped
\grecross used to produce a maltese cross but now you need to type \grealtcross in order to produce the same result...

henryso:
Fixes #713.

All tests pass, but I modified one to include `\grecross` and `\grealtcross`.  Please review and merge if satisfactory.


eroux:
ok for me

Issue #717, opened 2015-12-17
Compatibility with luaotfload 2.6


eroux:
It seems Gregorio is not compatible with latest luaotfload (2.6), see [this thread](https://mail.gna.org/public/gregorio-users/2015-12/msg00036.html)

henryso:
@rpspringuel Would you be able to test this against Tex Live 2014?

henryso:
It's still not working against luaotfload 2.6 and the latest TeX Live 2015.  I'll see if I have time this weekend to try to dig into why it's broken, but I'll take any help anyone can offer.

eroux:
Thanks for your help, I'm still unable to do anything before Monday, but I think a good start would be to isolate the problem outside of Gregorio (working in raw TeXLive, requiring no gregorio font nor files, but including the font access code of Gregorio), and make a PR against [the luaotfload test repo](https://bitbucket.org/phg/lua-la-tex-tests). This example should raise an obvious error at compile time if the bug is detected. @phi-gamma would that be ok for you?

phi-gamma:
> It's still not working against luaotfload 2.6 and the latest TeX Live 2015. I'll see if I have time this weekend to try to dig into why it's broken, but I'll take any help anyone can offer.

I”ll look into it as soon as I can, probably on monday. The table
accesses were the first thing that came to my mind as a potential
source of errors. (Indexing ``font.fonts`` is destructive and
overwrites parts of fonts defined by Luaotfload.)

henryso:
The problem is this function in `gregoriotex.lua`:
```
local function map_font(name, prefix)
  log("Mapping font %s", name)
  local glyph, unicode
  for glyph, unicode in pairs(get_score_font_resources(name).unicodes) do
    if unicode >= 0 and not string.match(glyph, '%.') then
      log("Setting \\Gre%s%s to \\char%d", prefix, glyph, unicode)
      tex.sprint(catcode_at_letter, string.format(
          [[\xdef\Gre%s%s{\char%d}]], prefix, glyph, unicode))
    end
  end
end
```
The `for glyph, unicode in pairs(get_score_font_resources(name).unicodes)` loop only iterates four glyphs from the greciliae.ttf font:
* zwj = 8205
* hyphen = 45
* zwnj = 8204
* space = 32
We have thousands of glyphs in the private use area; none of them are iterated by this loop.

The relevant part of the log is:
```
[...]

PR #718, opened 2015-12-18
gregoriotex.lua: overhaul access to Luatex fonts tables

phi-gamma:
Related to

    https://github.com/gregorio-project/gregorio/issues/717
    https://github.com/lualatex/luaotfload/issues/308

The font table was indexed directly which is unsafe when using
Luaotfload. Instead, font.getfont() should be used which is overridden
by the fontloader. Due to caching, those accesses are faster too than
indexing the font table directly as well as using the builtin
font.getfont() API.

In order to facilitate future extensions to the code, a number of
frequent lookups have been implemented as accessor functions.

       ----------------------------------------------------
        expecting font id       expecting name
       ----------------------------------------------------
        get_font_by_name()      get_score_font_id()
        get_font_resources()    get_score_font_resources()
       ----------------------------------------------------

The *score_font* varieties look up the string in the ``score_font``
table first. Since in the majority of cases font accesses are followed
directly by a lookup of the ``.resources`` subtable, this is covered by
the *resources* variants.

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -573,12 +573,14 @@ local function direct_gabc(gabc, header)
 end
 
 local function check_font_version()
-  local gregoriofont = font.getfont(font.id('gre@font@music'))
-  local fontversion = gregoriofont.shared.rawdata.metadata.version
-  if fontversion ~= internalversion then
-    fontname = gregoriofont.shared.rawdata.metadata.fontname
-    fontfile = gregoriofont.shared.rawdata.metadata.origname
-    err("\nUncoherent file versions!\ngregoriotex.tex is version %s\nwhile %s.ttf is version %s\nplease update file\n%s", internalversion, fontname, fontversion, fontfile)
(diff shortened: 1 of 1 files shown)


phi-gamma:
Hey guys, following the discussion on the list and the referenced
tracker items, I decided to look into Gregorio. There are problems
with how fonts are accessed. Please review the code in this PR, I
can extend / adapt it according to your needs. NB this isn’t tested
at all: I couldn’t get Gregorio installed on my distro without pulling
yet another Texlive install as dependency. *Thus it’s very likely that
the code contains blatant errors.*

Please test whether this works with recent Luaotfload.

eroux:
Thanks a lot, it really helps! Will this work with every (recent) versions of luaotfload?

eroux:
For other developpers: I'm ok to merge this even if you can only test with luaotfload 2.5 (I'll test on monday otherwise)

henryso:
Supplanted by #719.

PR #719, opened 2015-12-18
gregoriotex.lua: overhaul access to Luatex fonts tables

henryso:
Note: this pull request pulls the fix into the release-4.0 branch (for 4.0.1) rather than master, thus supplanting #718, which I will close.

Related to

* #717
* lualatex/luaotfload#308

The font table was indexed directly which is unsafe when using
Luaotfload. Instead, font.getfont() should be used which is overridden
by the fontloader. Due to caching, those accesses are faster too than
indexing the font table directly as well as using the builtin
font.getfont() API.

In order to facilitate future extensions to the code, a number of
frequent lookups have been implemented as accessor functions.

       ----------------------------------------------------
        expecting font id       expecting name
       ----------------------------------------------------
        get_font_by_name()      get_score_font_id()
        get_font_resources()    get_score_font_resources()
       ----------------------------------------------------

The *score_font* varieties look up the string in the ``score_font``
table first. Since in the majority of cases font accesses are followed
directly by a lookup of the ``.resources`` subtable, this is covered by
the *resources* variants.

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -573,12 +573,14 @@ local function direct_gabc(gabc, header)
 end
 
 local function check_font_version()
-  local gregoriofont = font.getfont(font.id('gre@font@music'))
-  local fontversion = gregoriofont.shared.rawdata.metadata.version
-  if fontversion ~= internalversion then
-    fontname = gregoriofont.shared.rawdata.metadata.fontname
-    fontfile = gregoriofont.shared.rawdata.metadata.origname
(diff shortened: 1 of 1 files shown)


henryso:
I'll review this tonight.  I was able to get my VM up-to-date yesterday evening.

henryso:
This does not work with 2.5: `getfont` (an alias to `font.getfont`) is returning nil when given a font id that was returned by `font_id` (an alias to `font.id`).  I'll test with 2.6 later.

henryso:
I spoke too quickly.  My first attempt failed because of typos, so I am doing more testing with 2.5, then 2.6 later.

Issue #692, opened 2015-12-07
Neography issue: virga cum punctum


eroux:
A quite important Abbey (anonymous for now) wishing to test Gregorio encounters a few limitations, the first one is that they want to be able to typeset this:

![neume1](https://cloud.githubusercontent.com/assets/60868/11631732/bb727fa2-9d03-11e5-9a12-8b618989bdf3.png)

I'm not sure about the way to note this in gabc... what do you think?

jakubjelinek:
I think this is related to issue #687 and issue #679.  At least for this issue and #687, one possibility would be to allow to say in gabc you want a stemless neume (in both cases clivis, but in other cases could be ancus, etc.), then just combine with ! or /.  I've also seen in Nocturnale Romanum 2002 (hghf)
which has not been typeset as a porrectus flexus, but like (h!g!hf) except that the final clivis should not have a stem.  Either it could be done with \ proposed in #679 or something different (some flag on the first note of the neume?).

eroux:
Sounds good! I don't particularly like '\', but it's totally subjective and I have no rational argument against it, so I have no objection to this solution!

henryso:
I don't have my heart set on `\`, but I do think this is a good use for the feature.  Any better suggestions as for which character to use to "glue" two parts of a neume together?

eroux:
In the remaining usual ascii characters, I think `*` or `@` could be candidates too... I kind of like `@`, but I don't have a strong opinion about this, at all...

henryso:
By the way, in case anyone is waiting urgently for this change (i.e., all the issues @jakubjelinek mentioned plus this one), it's not a simple thing.  I have the basics mapped out and am currently working on the algorithm for detecting the shapes and the how the TeX will need to change.  One of my current issues is how to place the signs without defining a huge number of cases for extended shapes.  My current thinking is that I have to "invert" the GreGlyph macro to allow signs to be placed at a sub-glyph level.  It seems that the entire shape needs to be a single glyph for syllable alignment purposes.

eroux:
There's nothing urgent for this issue, but maybe it could be dealt with in an easier way, as I think the current structure could be kept for the neume of this issue?

henryso:
[...]

Issue #687, opened 2015-12-03
Neography issue: salicus with clivis


jakubjelinek:
In Nocturnale Romanum 2002, p. [171], I see another neume I haven't managed to typeset with gregorio:
![sa cl](https://cloud.githubusercontent.com/assets/9370665/11572930/e8f57acc-9a03-11e5-979a-77f96a374c22.png)

I've tried (deOfd), (d!eo<!fd), (d!eo<fd), without success, the second one is closest to it, except that it has the stem that should not be there.
In nabc the corresponding neume is sa!cl .
Perhaps another case for backslash - (d!eo<\fd) ?

henryso:
What is the meaning of the reverse oriscus in that instance?  What does it signify different from the "normal" oriscus?

jakubjelinek:
Hard to say, I don't know.  That said, I haven't found a way to typeset it even with the normal oriscus.
The original manuscript has:
http://www.e-codices.unifr.ch/en/csg/0391/113/0/Sequence-1325
(the last neume above dignissima in Felix namque responsory).

henryso:
Well, with the fix for #631, you *should* be able to implement it with a normal oriscus.

jakubjelinek:
I'll try that soon.  Perhaps we could use O< instead of O for reverse oriscus in the salicus?

henryso:
My understanding (and the reason for my previous question) was that the flow of the oriscus (in square notation) is based on the note that follows, so an oriscus that looks like up-down-up (a.k.a., "normal") leads to a higher note and down-up-down ("reverse") leads to a lower note.  However, this figure seems to contradict that, and drawing lines to connect the notes in this figure results in a non-flowing figure.  If that's just an "artist's interpretation" of an oriscus, then I think the salicus flexus of #631 would suffice, and that figure may even be considered a typographical error.  However, if there is a semiological reason for it, then we need to think about how to draw it.  If we use the `<` in the middle of the figure, this is the first time a "liquescent" is applied to something other than the whole figure, and would require some changes to the data structures or some other "clever" way of dealing with it.

jakubjelinek:
I'll try to talk to various people about this (hopefully during the weekend).

henryso:
Have you had a chance to speak with anyone about this?

henryso:
#### The Problem
After working through the cases, the oriscus design is a bit of a mess.
* `o` = oriscus
* `o<` or `o>` = reversed oriscus
[...]

Issue #679, opened 2015-11-26
Neography issue: quilisma porrectus ancus


jakubjelinek:
In Nocturnale Romanum 2002, p. 278*, I run into problem typesetting:
mun(h!iwjijih~)di(ih)
which is typeset IMHO wrongly as:
![mundi](https://cloud.githubusercontent.com/assets/9370665/11429440/063360d6-9478-11e5-9d6a-32b28143ab4b.png)
In NR02 it looks like:
![mundi2](https://cloud.githubusercontent.com/assets/9370665/11429505/17fcea48-9479-11e5-9e5c-7224e09b3841.png)
I'd expect something close to a combination of h!iwjiji overlapped with jih~ over the ji, without the stem, so:
![mundi3](https://cloud.githubusercontent.com/assets/9370665/11429530/aa9ec6c8-9479-11e5-8b65-478b857956f1.png)
Unfortunately in this case I can't find any workaround, putting ! anywhere in between creates something wrong.

henryso:
The salient neume in this case is a 6-note neume.  At the minimum this is going to be a fused neume.  Maybe we should come up with a way to create fused neumes generically.  This requires some thought and perhaps some engineering.

jakubjelinek:
Yeah.  Do we have a neume for the iwjiji or is that already composed too?  If yes, is it composed from the quilisma and jiji without stems?  If we have ancus without stem (jih~) it could be fused with the portion of porrectus.

henryso:
All 5-note shapes are fused.  We don't have a stemless ancus, and even if we did, it'll take some headwork to come up with a sensible way to fuse these extended shapes.

henryso:
I'm still thinking about this, but here are my thoughts so far, in case someone wants to stop me from going in the wrong direction:

I'm thinking of the addition of something to gabc that would indicate a desired fusing of notes.  For the sake of discussion, I'll use the backslash (`\`).  So to get the figure above, you would specify something like `mun(h\iwjij\ih~)di(ih)`.

This would tell Gregorio (and inevitably GregorioTeX) that there should be a line (if appropriate) to connect the `h` to the `iw` and the `j` after the "porrectus" stroke to the `i`.  This would also cause the system to place the `i` after the "porrectus" stroke to the right rather than to the left of the shape since it would be fused to the next shape.

I'm not sure how complicated this kind of thing would be to implement.

jakubjelinek:
[...]

PR #723, opened 2015-12-23
Neume fusion

henryso:
This is a pretty huge pull request that covers #679, #687, and #692.

In addition to neume fusion and specifically the items requested in those issues, I also cleaned up a few things along the way:
* I removed `L_AUCTA`, instead using `L_AUCTUS_ASCENDENS`
* I made the oriscus shapes consistent in greciliae
* I made the scandicus shapes with a second ambitus of one more consistent
* I cleaned up squarize.py:
  * I turned the widths dictionary back into a global variable for simplicity; this script will not be run from a bigger script anyway, and we use other globals anyway.
  * I took the copyright from the source `.sfd` file, which should make it easier for users (like me) to create custom fonts without changing the script.
* I modified the PDF documentation to take the dimensions directly from the macros.

Please review this, try it out, and let me know if there's anything I missed or that I should change.


eroux:
Well, that's quite impressive, thanks a lot, this really takes Gregorio one step further! Ok for me to be merged

henryso:
Since @eoux has done the preliminary review, I'll merge it, but due to its size, I'm sure it's not 100% bug-free, so anyone willing to beta test this feature is welcome and encouraged to do so!

src/gabc/gabc-glyphs-determination.c
```diff
@@ -774,31 +881,86 @@ gregorio_glyph *gabc_det_glyphs_from_notes(gregorio_note *current_note,
...
                         liquescentia, current_note->previous);
                 current_glyph_type = G_UNDETERMINED;
             }
-            liquescentia = L_INITIO_DEBILIS;
+            liquescentia = head_liquescentia;
+        }
+
+        if (autofuse) {
+            if (current_glyph_first_note == current_note) {
+                if (first_autofused_note) {
```
eroux: Testing the oriscus feature, I realize I have a warning here:

```
gabc/gabc-glyphs-determination.c: In function ‘gabc_det_glyphs_from_notes’:
gabc/gabc-glyphs-determination.c:895:20: warning: ‘first_autofused_note’ may be used uninitialized in this function [-Wmaybe-uninitialized]
                 if (first_autofused_note) {
```

(the line number doesn't correspond)

henryso:
That warning is actually a false positive because—in the flow of the loop—autofuse will never be made true without also setting first_autofused_note to true.  My compiler (clang) doesn't generate this warning, but I will put in an initial value to appease your compiler.

PR #728, opened 2015-12-24
Workaround for @eroux's compiler warning noted in #723.

henryso:
The warning is a false positive because (in the flow of the loop) autofuse
will never be made true without also setting first_autofused_note to true.


henryso:
This is a one-line fix, and has no effect on the tests, so I will merge it.

PR #730, opened 2015-12-25
Selected the correct glyph for a fused porrectus-like flexus.

### LINKED ISSUE #729: Problem selecting correct fusion on porrectus-like flexus
~~If there is fusion though a glyph texverb, gregorio will incorrectly select non-fusion glyphs on either side of the texverb.~~

Edit: I thought this was a bug with fusion through a texverb, but the problem is that a figure such as `(@[ghjgih])` selects the incorrect fusion glyph between the `h` and the `j`.

I discovered this bug when I was trying to use fusion to solve #707.

henryso:
Fixes #729.

This is ready for review and merge if satisfactory.  I also added a test for the bug.


eroux:
Ok for me!

PR #731, opened 2015-12-25
Added support for external headers.

henryso:
As discussed on gregorio-devel.  External headers start with `x-` and have names composed of letters and numbers, optionally separated by dashes (i.e., `x-my-header3` is a valid name).

This is ready for review and merge if satisfactory.  I have extended `test_all_headers` to cover external headers.


eroux:
Ok for me!

PR #732, opened 2015-12-27
Implemented heuristics for reducing note-episema space at ledger lines.

### LINKED ISSUE #716: horizontal episemus above k with no ledger line
In the very simple case

```
k(k_ c_0)
```

the horizontal episemus above `k` is placed too high and the episemus below `c` is too low, to cope with the case where there are ledger lines.

I admit I see no straightforward way to deal with both cases (ledger line, no ledger line). Here are a few options:

- euristics in gabc compilation: trying to see if there are possible ledger lines around the episemus
- an additional option for horizontal episemus in gabc (maybe for vertical episemus too)
- a combination of both: having simple euristics but the ability to correct them when they're wrong

I think the third option would be the best one, with euristics that can be very simple... Also, one should be able to disable this new feature on the TeX side (globally, not episemus by episemus).

What do you think?

henryso:
Fixes #716.

Note: I ended up realizing that the heuristic override is actually a note attribute, so rather than adding `6` to the episema specification, I implemented this as something like `[hl:0]` (meaning to act as if **no** high ledger line exists) and `[ll:1]` (meaning to act as if a high ledger line exists) attached to a note.


eroux:
This is really great! Everything is fine for me, both the implementation and the result I see in the PR for tests. Thanks a lot!

henryso:
I tried out a few extra cases this morning, and it doesn't handle the porrectus correctly.  I'll take care of that this afternoon.

henryso:
I updated the heuristic to handle upper and lower notes like at the end of a porrectus.  I think this is now good to go.

eroux:
Ok for me

PR #735, opened 2015-12-28
Applied the lazy-load workaround for compatibility with luaotfload 2.6.

### LINKED ISSUE #734: GregorioNabcRef won't compile
In preparing the 4.0.1 release I noticed that GregorioNabcRef won't compile.  I'm getting the following error:

```
./veni-4_0_0.gtex:196: Undefined control sequence.
l.1 ...pace\hskip \gre@skip@nabcinterelementspaceERR
                                                  
l.196 }
```
The "ERR" at the end indicates to me a variable that isn't getting resolved correctly when writing the gtex file.  Can some one with more knowledge of the nabc system look into this and figure out what the problem is?

henryso:
Fixes #734.

This is the same fix as #717 applied to gregoriotex-nabc.lua.

I tested it on TeX Live 2015 with both luaotfload 2.5 and 2.6, and it works fine.  All tests pass.

### DIFF
--- a/tex/gregoriotex-nabc.lua
+++ b/tex/gregoriotex-nabc.lua
@@ -71,6 +71,15 @@ local function gregallreadfont(font_id)
   local tab = {}
   local metrics = {}
   local fontdata = fonts.hashes.identifiers[font_id]
+  -- The unicodes table may be lazy-loaded, so iterating it may not
+  -- return everything.  Attempting to retrieve the code point of a
+  -- glyph that has not already been loaded will trigger the __index
+  -- method in the metatable (implemented by the fontloader) to load
+  -- the table until the glyph is found.  When iterating the
+  -- unicodes, we want the whole table to be filled, so we try to
+  -- access a non-existing glyph in order to force load the entire
+  -- table.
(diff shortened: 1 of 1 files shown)


tex/gregoriotex-nabc.lua
```diff
@@ -71,6 +71,15 @@ local function gregallreadfont(font_id)
   local tab = {}
   local metrics = {}
   local fontdata = fonts.hashes.identifiers[font_id]
```
eroux: I think it would be good to use `font.getfont` instead of accessing `fonts.hashes.identifiers`, as suggested in [this commit](https://github.com/phi-gamma/gregorio/commit/392e77a0c3c80e6da2d633ba3cc560b03644fa2a)

eroux:
apart from my comment, ok for me to merge

eroux:
see [this PR](https://github.com/henryso/gregorio/pull/7) for a fix

henryso:
This works under 2.5 as well.  I wasn't sure that `fonts.hashes.identifiers[id]` produced the same thing as `font.getfont(id)`.  Merging.

PR #738, opened 2015-12-30
Added half-space and ad-hoc space.

### LINKED ISSUE #736: Fine space tuning
For the new Antiphonale Monasticum project, some differences in the manuscripts are noted with differences in spacing, for instance `f/hoi` will be the transcription for a certain neume, while another will be noted by replacing the space with half a space.

The original request was to have the possibility of negative spaces, but I see several:

1. find a new symbol to denote half a space in gabc, maybe `/0`
2. ask people to tune the different spacings in a custom spacing file, using for instance `/` as half a space, `//` as a normal space, etc.
3. use `-/`, `-//` and `- ` for negative spaces, in this case `f/-/hoi` would be ok with `-/` tuned to be half a space
4. completely generic solution: `/{1/2}` is half a space, `/{-3/2}` is a negative space of width 3/2 of a normal space, `/{2}` is a space of twice the normal space, etc.

What do you think?

Edit: I've made a few edits since the initial post...

henryso:
Fixes #736.

This is ready for review and merge if satisfactory.


henryso:
OK'd by @eroux in #736.

PR #746, opened 2016-01-01
Used the correct glyph for punctum inclinatum deminutus positioning.

### LINKED ISSUE #742: alignment: vertical episemus on punctum inclinatum debilis
Doing `(gh!ijI'~H~)` in gabc results in poor alignment between vertical episemus and punctum inclinatum debilis:

episemus-alignment

the episemus is too much on the right. I discovered that changing the example.

henryso:
Fixes #742.

All current tests pass, and I added a test for this bug.


eroux:
Thanks again!

PR #749, opened 2016-01-02
Prevented fusion from cavum shapes.

henryso:
Looking at the fusion.gabc output more closely, I noticed that the system was attempting to fuse the cavum shapes to the following notes.  This pull request suppresses that.


eroux:
Thanks!

Issue #662, opened 2015-10-27
Integrate commentary into score


rpspringuel:
As it stands, commentary is not part of a score, but rather a convenience command (`\grecommentary`) for printing right-aligned text.  This request is that commentary be made part of the score, thereby allowing a couple of features:

1. Control of the distance between the staff-lines and the commentary text.  As is, any element which appears above the staff lines pushes the commentary and the staff-lines apart.  By bringing the commentary into the score, the position of the commentary can be made relative to the staff-lines, allowing greater control over their positioning.
2. Allow the `commentary` header field in gabc to transfer over much like the annotations fields do.

For implementation I see rewriting `\grecommentary` so that it fills a box, much like `\greannotation` does and then having that box be placed in the score by one of the macros used to construct the first line (again, much like the macro writing the initial is used to place the annotation box).

eroux:
Here's an attempt to control the distance between `\grecommentary` and the staff lines (inspired by @OldClaudio):

```tex
\makeatletter
\xdef\raisecommentary{1.4ex}

\def\grecommentary{\@ifnextchar[{\gre@commentary}{\gre@commentary[0pt]}}%

\def\gre@commentary[#1]#2{%
  \ifhmode\par\fi %
  \noindent\hbox to \hsize{\hfill #2}%
  \par %
  \kern\dimexpr#1+\raisecommentary-\baselineskip\relax %
  \relax %
}
\makeatother
```

This makes commentary closer to the lines by default, but the distance is tunable through an optional argument. I think it's a reasonable first step and it will help users using commentaries outside scores, but it doesn't really solve the issue...

What do you think?

Also, the example looks a bit better with the commentary in `\footnotesize` instead of `\small`, but that's not the core issue...

henryso:
I propose #750 as a solution for passing custom headers to TeX.  This proposed solution can be used to pass some header value to `\grecommentary`.  It does not enhance `\grecommenrary` to fill a box.

rpspringuel:
Because this information should be here for reference:

@eroux started to try a fix based on his suggestion above the fix-662 branch.  After looking it over, this was my response:

> I've taken a glance at fix-662 and have to say that the direction Élie is taking there is completely different from the solution I was thinking of.  As I haven't even tried to implement what I was thinking of yet I don't know which solution would be better or if they would both effectively get to the same end point.  Once I finish up with the annotation stuff, I'll play around with my ideas to see how it stacks up against what Élie is doing. 

[...]

PR #750, opened 2016-01-04
Allowed headers to be processed in TeX.

henryso:
This is my proposed solution to (the C-side of) #662.  The user can use a custom header (say `x-commentary`) and pass it to the `\grecommentary` macro by specifying `\gresetheadercapture{x-commentary}{grecommentary}{}`.  I thought that perhaps a general solution would be more useful to allow third-party use of headers both upstream and downstream from the gabc file.

Feel free to nix this if you don't feel it's a good idea.  Otherwise, please review and merge if satisfactory.


henryso:
I added the documentation you requested to GregorioRef and also put more comments in `gregoriotex-write.c` to explain the method of escaping and the reason behind it.

eroux:
Thanks a lot! I think it's ok like this

PR #751, opened 2016-01-06
Added two-, three-, and five-line staff support.

### LINKED ISSUE #429: 5-line staff
Some people requested to be able to work on five lines. It could be an option in TeX.

A difficulty is that in gabc, `o` already means oriscus, so it can't mean the letter above `n`... Maybe `o0` would be acceptable...

henryso:
Fixes #429.

I'm not sure about the DivisioMinor glyphs, so pay extra attention to them.

Of the current tests, only the gabc-gtex tests fail because of changes to the macros.  I added tests for the new feature.


eroux:
Apart from my two tiny comments, this looks perfect. This is quite impressive, and really good looking, thanks a lot!

henryso:
I made the change you suggested to the change log and elevated the warning in `gregorio_set_score_staff_lines` to an error.

eroux:
Ok, thanks a lot!

PR #757, opened 2016-01-07
Implemented the ability to add a secondary clef.

### LINKED ISSUE #755: feature: double keys
This is a feature that was asked by an abbey to ease transcription: it would be nice to be able to use two keys, stacked one on another. The syntax could be the same as "polyphony", for instance `c4{c1}`. Only the first specified key would be taken into account for the computations, the other one would just be typeset. It should also work for clef changes, and all new lines should contain both keys.

henryso:
Fixes #755.

While doing this, I found and fixed some bugs with flatted clefs on 5-line staves.

All the gabc-gtex tests fail because new arguments were added to old commands, so this was acceptable.  Two of the tests actually had incorrect expectations due to the 5-line staff bugs; these have also been accepted.  All other tests pass.  I added a couple of tests to exercise the new functionality from this pull request.


eroux:
Perfect, thanks a lot!

PR #759, opened 2016-01-07
Added a 'mode-modifier' header for typesetting text after 'mode'.

### LINKED ISSUE #756: gabc header: handle transported modes
gabc header `mode` can now contain only one arabic number, but some pieces require to indicate that the mode is "transported". Quoting:

> Each mode should finish with a fixed note: 1st and 2nd finish in D, 3rd and 4th in E, 5th and 6th in F and 7th and 8th in G. But in manuscripts sometimes this doesn't happen, and you can find, for example, a 6th mode that finishes in C: it's a 6th transported a 5th up.
> This should be indicated because cantores must know what note is the first to start psalmic tone.

So maybe the convention (and the code) for `mode` header should be one arabic number + an optional `t` (ie: `6t`)?

henryso:
Fixes #756.

All current tests pass except the gabc-gabc tests that emit `\GreMode`, which changed for this pull request.  I added some additional tests to check the new functionality.


eroux:
Thanks a lot!

PR #762, opened 2016-01-07
Fixes for make distcheck failures.

henryso:


henryso:
These are C90-related warnings found when running make distcheck.  This doesn't affect any tests.  Merging.

PR #763, opened 2016-01-07
Suppressed the custos immediately before EUOUAE blocks.

### LINKED ISSUE #761: optional removal of custo before <eu> blocks
An abbey asked for the possibility to prevent custo to appear if there is a line break before `<eu>` blocks

henryso:
Fixes #761.

Due to the nature of the `<eu>` block, I decided it made the most sense to suppress the custos immediately before it by default, but it may still be turned back on.  Because this is a change in behavior, I added a note to UPGRADE.md.

I made some small fixes to the documentation while I was documenting the new code.

There were two existing tests which failed, but gave an appropriate result given these changes, so I accepted them.  I also added a test for this new feature as well.


eroux:
Wonderful, thanks a lot!

PR #770, opened 2016-01-08
Allowed optional ragged automatic line break immediately before EUOUAE.

### LINKED ISSUE #764: left alignment before <eu> linebreak?
For the Antiphonale Monasticum project, it would be great if the lines were aligned left (and the custos removed as in #761) if a line break occurs before an `<eu>` block. This behavior should be optional too... I'm not entirely sure it's feasible easily (if at all)... The most important is that

```
(::) (Z) <eu>euouae(g) </eu>(::)
```

and

```
(::Z) <eu>euouae(g) </eu>(::)
```

remove the custo

henryso:
Also fixed bug preventing custos suppression between `(::) (Z)` and `<eu>`.
Fixes #764.

All current tests pass except for the gabc-gtex test that emits `\GreBeginEUOUAE` and `\GreNextSyllableBeginsEUOUAE` which now require an extra argument.


eroux:
Perfect, thanks a lot!

Issue #767, opened 2016-01-08
space around bars


eroux:
The space around bars in Gregorio are really too wide, but the algorithm should also be improved, and be like the one with normal syllables (only much more complex).

Also, I believe there is a bug somewhere because the space on the left of bars with text is really wider than the space on the right, so there might be two spaces, or a missing `%` somewhere...

eroux:
A few examples from Solesmes' books, some interesting edge cases to handle:

![exemple2](https://cloud.githubusercontent.com/assets/60868/12218250/6ab09898-b71a-11e5-8231-c1dec1eff087.png)

![exemple3](https://cloud.githubusercontent.com/assets/60868/12218251/6f44a516-b71a-11e5-8024-032906a8faf6.png)

![exemple4](https://cloud.githubusercontent.com/assets/60868/12218252/73fd8fa0-b71a-11e5-8b36-589738b079ba.png)

![exemple1](https://cloud.githubusercontent.com/assets/60868/12218255/7f63a0aa-b71a-11e5-91a1-8e8255785a4f.png)

rpspringuel:
The ones where the `*` appears under the notes strikes me as being more like `ra *(efef) (,)` rather than `ra(efef) *(,)` (to use the third as an example).  Making the latter print something like is shown would require identifying the text of the bar syllable as an asterisk.

The last example should be accomplishable in the following manner:

1. After printing a bar set `\gre@lastisbartrue`.
2. After a line break set `\gre@lastisbarfalse`.
3. At the beginning of a normal syllable, `\ifgre@lastisbar` and `begindifference` is negative (i.e. text starts before notes) then kern backwards the smaller of `spacearoundbar + barwidth` or `-begindifference`.
4. At end of normal syllable set `\gre@lastisbarfalse`

henryso:
To me, it looks as if the `*` is centered in the space between the syllables, which requires something a bit more clever than `ra *(efef) (,)`, no?

eroux:
Indeed, it's quite complex, the few notes I have (in my head) are:

- when no text is associated, spacing of text between previous and next syllable should have the same constraints as if there was no bar

- when there is text but no notes nor bars (it should be a BarSyllable in this case, I don't know how it is now), space between notes of the previous syllable and notes of the next syllable should have the same constraint as if there was not text, and text should be centered between syllables, with the same constraints as normal text (minimal space with previous and next syllable)

- where there is a bar and text:
[...]

PR #772, opened 2016-01-09
Added tunable spaces for bars with text underneath.

### LINKED ISSUE #766: dissociation for space around bars according to presence of text
An abbey told me that bars without text should be less stressed than bars with text, and thus need less space around them. Maybe there could be a way to note bars without text that should be stressed in gabc, but I'm not sure it makes much sense...

henryso:
Fixes #766.

There was already a test that demonstrates the change, so I only updated the expectations for the tests that failed in the expected manner.  I think the unit test changes will probably conflict with one or more of the other recent pull requests, but I'll fix fix them once everything acceptable is merged.

I left the current spaces as is and added slightly larger spaces for bars with text underneath.  Let me know if I should tune them in some other way.

To answer the question in the issue, you can put something like `{ }(:)` to use the larger spaces around a bar without text.


henryso:
@eroux If your ideas for #767 conflict with this, feel free to close this without merging.

eroux:
Thanks a lot, this looks good! The only thing I would change is reducing the spaces for bars without text and not enlarge the spaces for bars with text (all projects I follow tend to reduce the space around bars a lot)

henryso:
That will obviously break a lot of tests since most bars don't have text.  Can you quantify the size of the reduction/enlargement?  I don't have a particularly artistic eye, so I have little confidence in what I think looks good here.

eroux:
Me neither to be honest... but let's merge it like this, they'll change in #767

PR #775, opened 2016-01-10
Used different glyphs for liquescent stropha on a line.

### LINKED ISSUE #773: differenciate some glyphs on lines / between lines
The idea never crossed my mind, but it's quite a good one and it's needed for a custom font: the stropha aucta can be very tall and some fonts use a different drawing if the stropha aucta is one a line or not... Maybe there could be an optional `StrophaAucta.interline` in the font that would be automatically used for this purpose? I don't think there another candidate for such a distinction... maybe `Stropha`?

henryso:
Fixes #773.

The glyph used for the liquescent stropha on a line is called `StrophaAuctaLongtail`.

I also stretched out the liquescent stropha glyphs in greciliae, but I did it in a separate commit so I can back that out if we don't like they way they look.  Please take a look at the test expectations and let me know if I should change them or back that change out.

It passes all current tests if run against the non-stretched glyphs.  I added a test specifically to test the new glyph.  I also committed (in a separate commit) updated expectations to match the stretched glyphs.

Please review and commit if satisfactory.


eroux:
Ok, if I understand correctly, the long queue is for stropha on a line? If it's the case then ok for me to merge!

henryso:
Yes, contrary to my earlier comment in #773, The longer-tailed stropha is the one on the line.

I think the technical side of this pull request is confirmed.  I will split out the "artistic" font changes into a separate pull request and merge this.

PR #777, opened 2016-01-10
Stretched out the liquescent stropha glyphs.

henryso:
This was broken out from #775, implementing the `StrophaAuctaLongtail` of #773 in greciliae.

Technically, I like the idea of supporting the new feature with greciliae because that will keep the feature from stagnating.

Artistically, I created the `StrophaAuctaLongtail` by pulling the lower points down (at various distances) and adjusted the spline parameters to try to maintain a round-ish shape.  I also stretched the normal `StrophaAucta` a little to make them less dissonant.

I don't have confidence in my artistic skills, so any comments to this are welcome, including something like "I don't like it, let's not use it".

Please review.

### DIFF
--- a/fonts/greciliae-base.sfd
+++ b/fonts/greciliae-base.sfd
@@ -20,7 +20,7 @@ OS2Version: 0
 OS2_WeightWidthSlopeOnly: 0
 OS2_UseTypoMetrics: 1
 CreationTime: 1176402534
-ModificationTime: 1452395542
+ModificationTime: 1452449649
 OS2TypoAscent: 0
 OS2TypoAOffset: 1
 OS2TypoDescent: 0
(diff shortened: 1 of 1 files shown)


eroux:
I admit I'm not a big fan... I don't really like the fact that it crosses the line below, maybe instead of making it taller, it could go further on the left? (like in Solesmes font)

henryso:
Here's another try moving the tail further to the left and stretching it slightly.  The effect is *much* more subtle, and does not cause the tests to fail due to the pixelization at the dpi of the image comparison.  However, I have accepted the newer PDFs in the corresponding test pull request so that you can see what it looks like.

eroux:
This looks fine to me!

PR #778, opened 2016-01-11
Implemented oriscus orientation dependency on the note that follows.

### LINKED ISSUE #774: Oriscus direction at the end of a syllable
In looking through the 2005-era Antiphonale Monasticum volumes out of Solesmes, It is very clear to me that the direction of the oriscus at the end of a syllable depends on the note that follows the oriscus, even though it is not connected to it in the way a salicus or pressus would.  We currently don't do this, but it would be pretty easy to do.

From a clean slate, what I would envision is that the system automatically detects the direction of the oriscus, defaulting to the down-up-down oriscus if it appears at the end of the score (no notes follow).  Appending `>` to the figure would force a down-up-down oriscus, and `<` would force an up-down-up oriscus should the user desire something specific.

Obviously, doing this would break existing scores.

There are several options:
1. Leave things as they are
2. Implement this, breaking existing scores
3. Implement this, but for a different letter (i.e., `u` instead of `o`)
4. Implement this as the default behavior, but add a header that will disable it
5. Implement this as a feature that can be enabled with a header setting (i.e., it is off by default)

The pros and cons of the first two are obvious.  The others add the feature in various degrees in exchange for additional complexity, both in documentation, in implementation, and most likely in support.

Any thoughts?

henryso:
Fixes #774.

I implemented option 4.  The header is `oriscus-orientation`, which should be set to `legacy` to restore the old behavior.

In the tests we used '<' for the "reverse" oriscus, and I changed that to '>' to mean the "descending" oriscus.  I also added a new test to exercise the new header.


eroux:
Ok for me! Thanks a lot!

PR #779, opened 2016-01-12
Added support for typesetting simple slurs.

### LINKED ISSUE #776: inclinated brace
An abbey asked for the ability to do:

bracecourbe

it should be extremely rare so I'm not really sure it's worth having a gabc notation for this, it could be just something to put in a `<v>` snippet with hardcoded length values, but it should be possible...

If it's possible in 4.1 that would be great, otherwise there's no emergency.

henryso:
Fixes #776.

I created the slur with two simple bezier curves and a rotation to put the slur in place.  This probably needs some tuning, but I think it's a good foundation.

Please review and let me know what I should change, or merge it if it is satisfactory.

Note: I found some discrepancies in the documentation which I fixed incidentally as part of this pull request.

All current tests pass, but I added one to test the slurs.


eroux:
Wow, this is really nice! Very easy to typeset and perfect output! Ok for me to be merged!

PR #780, opened 2016-01-12
Modified to accept all headers and reduce the number of special ones.

henryso:
Preparation for the mode annotation challenges from #756.

In addition to accepting all headers, I also added some hook macros for before and after the headers are processed.  I cleaned up all the special capture of headers that did not cause any effect to the  generated gtex output.

This breaks a number of tests because of header order and the new macros, but the new results look good to me.

Please review and merge if satisfactory.  Once the way forward on mode is more clear, that will be a different pull request.


doc/Gabc.tex
```diff
@@ -118,12 +94,23 @@ \subsection{Header}
...
   \verb=\greannotation= then the \texttt{annotation} header field will not
   overwrite that definition.
-\item[x-\textit{external}] External headers start with \texttt{x-} and may
-  have any name composed of ASCII letters and numbers, optionally separated
-  by dashes.  These headers are blindly captured by gregorio without any
-  further error checking.  They may be useful for external applications to
-  use as metadata.
+\end{description}
+
+Although gregorio ascibes no special meaning to them, other suggested headers are:
```
eroux: I think it should be "ascribes"

src/dump/dump.c
```diff
@@ -109,50 +109,14 @@ void dump_write_score(FILE *f, gregorio_score *score)
...
-    if (score->si.book) {
-        fprintf(f, "   book                      %s\n", score->si.book);
-    }
-    if (score->si.transcriber) {
-        fprintf(f, "   transcriber               %s\n", score->si.transcriber);
-    }
-    if (score->si.transcription_date) {
-        fprintf(f, "   transcription_date        %s\n",
-                score->si.transcription_date);
+    if (score->author) {
```
eroux: I think this should be `score->manuscript_reference`

eroux:
Apart from the two comments, it looks fine to me!

henryso:
Thanks for catching those typos.

PR #783, opened 2016-01-14
Added a command-driven means of hooking into before/after header events.

### LINKED ISSUE #782: Add a hook-like interface to `\GreBeginHeaders` and `\GreEndHeaders`
Per @rpspringuel's comment in #756, rather than redefining `\GreBeginHeaders` and `\GreEndHeaders`, the user should call some (new) command  to specify things to do before and after the headers are processed.

Rather than LaTeX-like additive hooks, I intend for these to work like `\localleftbox` and `\localrightbox` in that a new invocation overwrites the behavior of the previous invocation.

henryso:
Fixes #782.

All current tests pass, and I added a test for the new functionality.


eroux:
Ok for me, sorry for the delay!

PR #785, opened 2016-01-15
Added mode-differentia header and improved Roman numeral mode handling.

henryso:
This should implement everything remaining in #756, with the exception (as stated in the issue) of the arbitrary ordering of mode headers in the annotation.

The expected tests fail (and have been accepted), and I've added the header into some of the existing tests to cover the changes.


src/dump/dump.c
```diff
@@ -109,9 +109,6 @@ void dump_write_score(FILE *f, gregorio_score *score)
     if (score->score_copyright) {
         fprintf(f, "   score_copyright           %s\n", score->score_copyright);
     }
-    if (score->language) {
-        fprintf(f, "   language                  %s\n", score->language);
-    }
```
rpspringuel: Why are we deleting this header field?  Is it present somewhere else in the file?

tex/gregoriotex-main.tex
```diff
@@ -646,37 +646,38 @@
...
+\def\gre@romannumeral@majuscule#1{\uppercase\expandafter{\romannumeral#1\relax}}%
+\def\gre@romannumeral@minuscule#1{\romannumeral#1\relax}%
+\def\gresetmodenumbersystem#1{%
+  \IfStrEq{#1}{roman-majuscule}%
+    {\let\GreModeNumber\gre@romannumeral@majuscule}%
+    {\IfStrEq{#1}{roman-minuscule}%
+      {\let\GreModeNumber\gre@romannumeral@minuscule}%
+      {\IfStrEq{#1}{arabic}%
+        {\let\GreModeNumber\relax}%
+        {\gre@error{Unrecognized option in \protect\gresethyphen}}%
```
rpspringuel: C/P error.  Should be `\gresetmodenumbersystem` not `\gresethyphen`

src/dump/dump.c
henryso: All headers (except the `def-m*` headers) are actually now stored in `gregorio_score.headers`.  The section you are seeing here dumps values that are additionally stored in the `gregorio_score` structure, and there's no need to store language in that structure, so I removed it from the structure and therefore from the dump.  If you take a look at https://github.com/henryso/gregorio-test/blob/fix-756/tests/gabc-dump/test_all_headers.dump you'll see that language is still dumped in the "Headers" section.

rpspringuel: Ah, so this is simply catching something missed up with another change.

henryso: Yes.  I didn't realize this field wasn't needed after parsing until I refactored the way headers are parsed in making the changes for this pull request.

doc/Gabc.tex
```diff
@@ -73,9 +73,13 @@ \subsection{Headers}
   \item There is a \verb=\greannotation= defined immediatly prior to \verb=\gregorioscore=.
   \item The \texttt{annotation} header field is defined.
   \end{itemize}
-\item[mode-modifier] The mode "modifier" of the piece. This may be any
+\item[mode-modifier] The mode ``modifier'' of the piece. This may be any
```
eroux: I globally agree with this PR, I just think it would be useful to have a few convoluted mode examples and tell the user which part goes in the modifier and which in the differentia field... so that some conventions are respected and parsing is easier (for 3rd party softwares)

henryso:
@eroux I added some examples.  Is that sufficient?

eroux:
Yes, perfect, thank you! I have no other objection for it to be merged

doc/Gabc.tex
```diff
@@ -109,10 +113,58 @@ \subsection{Headers}
...
+the \texttt{annotation} gabc header nor the \verb=\greannotation= \TeX{}
+command is used.
+
+The mode annotation will look like
+\writemode{mode}{\thinspace mode-modifier}{\thinspace mode-differentia}.
+
+The \texttt{mode} header is typically a number that will be typeset as a
+Roman numeral using the \texttt{modeline} style.  However, there are other
+modes.  If the first character of \texttt{mode} is a number from one
+(\texttt{1}) through eight (\texttt{8}), that number will be converted
```
rpspringuel: The sentence which starts with "However" seems out of place to me, but moving it has some knock on effects to the rest of the paragraph.  I might try it as follows:

> The `mode` header is typically a number that is typeset as a Roman numeral using the `modeline` style.  Therefore, if the first character of `mode` is a number from one (`1`) through eight (`8`), that number will be converted according to the `\gresetmodenumbersystem` setting.  However, there are other modes, so all other values for `mode` will be typeset directly.  If the `mode` header is omitted, none of the other headers will be typeset.

rpspringuel:
I'm getting a test failure in tex-output/bugs/fix-508.  Looking at the diff, it appears to be font related, something I ran into in my own commentary modifications.  I updated the test on that branch to use the packaged Alleygra font in order to avoid this coming up again, so I don't think we need to resolve that issue here too.

Beyond these comments, I see this as ready to merge.

henryso:
The fix-508 expectation was *not* changed as part of this pull request.  That said, based on past experience on this very same test, I'm pretty sure it doesn't have anything to do with the font.  My guess is that the value of `\parindent` is slightly different between your version of TeX and the one @eroux used when he accepted the differences to fix-508.  I haven't been able to test this yet, but I think setting `\parindent` to 0pt will solve this discrepancy.

rpspringuel:
Okay, I'm splitting off that test issue into gregorio-project/gregorio-test#100.

PR #786, opened 2016-01-15
Feature/commentary box

### LINKED ISSUE #662: Integrate commentary into score

rpspringuel:
We place the commentary into its own box which is then printed at the beginning of the score.  This allows finer control over the placement and spacing of the commentary and more freedom in where the `\grecommentary` commands occur.

We also implement styles for the annotation and commentary.

Fixes #662.


examples/main-lualatex.tex
```diff
@@ -30,10 +30,11 @@
 \gresetlinecolor{gregoriocolor}
 
 % We set VII above the initial.
-\greannotation{\small \textsc{\textbf{VII}}}
+\grechangestyle{annotation}{\small\scshape\bfseries}
```
eroux: maybe this part could just be removed, and the mode automatically printed?

rpspringuel: We could, but then the example doesn't demonstrate the use of `\greannotation` (something the Factus example below already demonstrates).  Perhaps what is needed is a slightly more complicated annotation which can't simply be imported from the mode field.  Maybe a two line annotation like:
```
\grechangestyle{annotation}{\small\bfseries}
\greannotation{Intr.}
\greannotation{\textsc{VII}}
```

eroux: ok yes, I have no further objection, it's ok for me

tex/gsp-default.tex
```diff
@@ -210,8 +210,12 @@
 \grecreatedim{initialraise}{0 cm}{scalable}%
 % Space between lines in the annotation
 \grecreatedim{annotationseparation}{0.05cm}{scalable}%
-% Amount to raise (positive) or lower (negative) the annotations from the default position (base line of top annotation aligned with top line of staff)
+% Amount to raise (positive) or lower (negative) the annotations from the default position
 \grecreatedim{annotationraise}{0cm}{scalable}%
+% Space between lines in the commentary
+\grecreatedim{commentaryseparation}{0.05cm}{scalable}%
+% Amount to raise (positive) or lower (negative) the commentary from the default position (base line of bottom commentary aligned with top line of staff)
+\grecreatedim{commentaryraise}{0.4cm}{scalable}%
```
eroux: Sorry, I have two final comments:

- maybe `commentaryraise` could be smaller, like 2 or 3mm. The Antiphonale Monasticum has commentaries close to the lines, and it really looks good! 
- it could be made easier to rise the commentary for one score only, with an optional argument to `\grecommentary` (as I did in fix-662)

rpspringuel: 1. I picked 0.4cm to be reasonably close to the existing behavior.  I can easily shrink it if something tighter is more desirable.
2. I actually don't like this idea because it provides extra pathways for the unexpected to happen.  I'd much prefer the user having a group bound `\grechangedim` command for changing `commentaryraise` for just one score.

eroux: 1. I think it is more desirable yes

2. I don't really understand your point... what unexpected? Also, easiness seems important for you, and I'm affraid changing a dimension in a group actually involves very difficult notions for a beginner... And when the commentary is close to the score, it's necessary to raise it by hand if there are some high notes.

3. Something else I forgot: are you sure your implementation never allows page breaks between commentary and score? (that was a problem with the old implementation)

rpspringuel:
Continuing discussion from outdated diff:

1. Done.

2. Let's say that you've started a project and have manually adjusted the commentary spacing for a few scores based on the presence of high notes.  As you continue working, you decide that globally you want more spacing between the commentaries and the score by default and thus add a `\grechangedim{commentaryraise}...` to your preamble.  What happens to those scores in which you've manually adjusted the commentary position?  If the changes were group localized, as I suggested, then they don't change.  However, if they were manually tweaked through an additional argument as you proposed, then they too will move as a result of the global change, forcing you to go back and retweak those scores.  That is one example of the sort of "unexpected" that I'm referring to.

3. I'm fairly certain that a page break is impossible after the commentary.  In my implementation the commentary box is the first element on the first line of the score (even before the initial) and for spacing purposes, TeX sees it as taking up 0pt.  It should thus should never want to break the line, let alone the page, after the commentary but before the initial.

eroux:
2. I don't really get your use case: if someone has a `commentaryraise` of `1mm` and uses `\begingroup\grechangedim{commentaryraise}{2mm}\includescore{}\endgroup` because of high notes, and then decides to make it 3mm, he still has to retweak everything... But maybe what you mean is that it would be better if the optional argument changes `commentaryraise` instead of adding something to it?

3. I looked at the implementation, and you're right, I should have checked earlier, sorry!

rpspringuel:
Hypothetical document using the optional argument:
```
\grecommentary{something}
\gregorioscore{something}

\grecommentary[1mm]{something with high notes}
\gregorioscore{something_with_high_notes}

\grecommentary{something else}
\gregorioscore{something_else}
```

Now, let's say I want to reduce the space between the commentary and the score in general from the default (now 2mm) to 1mm.  If I add `\grechangedim{commentaryraise}{1mm}{scalable}` to the beginning of the above document, what should happen to the score in the middle?  Does it's commentary simply get closer to the score as well?  Since I had to manually tweak it upwards to avoid contact with high notes, bringing it closer may cause it to conflict with those notes again, forcing me to go back and tweak it individually again (not bad if there's only one, but imagine this decision is made as part of the final editing process for a large book).

If, on the other hand, the middle score had been:
```
{
\grechangedim{commentaryraise}{3mm}{scalable}
\grecommentary{something with high notes}
\gregorioscore{something_with_high_notes}
}
```
In this case the global change would not affect this particular score because I'm already overriding `commentaryraise` for it.  My basic assumption is that once I've hand tweaked a parameter for a score, global changes to the default value for that parameter should not affect that score.

> But maybe what you mean is that it would be better if the optional argument changes commentaryraise instead of adding something to it?

This would be more along the lines of how I think the behavior should work.  Implementation is not trivial, but I could come up with something along these lines if that's what is wanted.

eroux:
Well, I understand your point, but I don't really think it's very important... in the case you say where the change is made as part of the final editing process for a large book, this simply means that there has been a very bad management on the book creation... And if this kind of change is made, this will completely change all vertical spaces everywhere, and they will all have to be tweaked again, not just the commentaryraise of scores with high notes, all spaces of scores between each other, etc. So retweaking the manually changed commentaryraise is really nothing compared to that...

But I'll implement a custom `\grecommentary` for the Antiphonale Monasticum project if you really think it's a bad idea, no problem.

henryso:
I'll admit that this discussion is a little bit over my head, but maybe a solution would be to have a *second* commentary raise which adds onto the existing one, but is reset to `0pt` at the end of the score.  Since it would be `0pt` unless explicitly set to something before a score, there ought to be no surprises, and you would be able to tweak the value for a given score.

Please feel free to ignore me since I don't think I fully appreciate the problem.

eroux:
That's basically the solution I'm thinking about yes

rpspringuel:
And that's exactly the solution to which I'm objecting.

> in the case you say where the change is made as part of the final editing process for a large book, this simply means that there has been a very bad management on the book creation...

I don't think that is necessarily the case.  Especially in instances where the commentary and the notes overlap, it becomes necessary to make this tweak early so that the drafts are readable.  Ideally layout and content decisions are made separately (content first, then layout), but the content has to have *some* layout applied to it if it is going to be checked for accuracy and it only makes sense to apply the layout that it will most likely have in its final form (even if that is still somewhat if flux).

However, if both of you feel that this sort of option is more useful than problematic, I'll implement it.

henryso:
It all depends on how you see the "document-wide" value.  If you see that as a "book style" decision, then you'd have to go back and readjust all the individual values anyway to match a change in the "book style."

I personally don't see a big problem with having the two (additive) values, as long as the difference between the two and how they are used is well-documented.  If you prefer to do things in the way @rpspringuel describes, then you simply don't need to use this additional adjustment, which would default to `0pt`.  Of course, it does give the user another way to get confused, but that should be alleviated though good documentation.

rpspringuel:
> Of course, it does give the user another way to get confused,

That is why I was objecting in the first place.

> but that should be alleviated though good documentation.

and that is why I'm willing to cede the ground if others think the feature is worth more than the potential problems.

eroux:
Well, the only book I'm more or less involved with is quite large (900 pages) and will use this (in fact it already has a custom implementation of grecommentary), so I think it's worth implementing... But @henryso is right, documenting the two options (optional argument and local change) is a good idea!

rpspringuel:
Feature added.  Please make sure my documentation is clear enough for you.

henryso:
I think the description is fine for the optional argument, though I might want to somehow emphasize (bold?) the fact that the additional dimension comes from the last commentary line and applies only to the next score, but perhaps that would stand out too much.

Should the use of group-bound settings be explored a bit more (in general) to explain the other use-case (and similar use-cases).  I'll admit that I've never fully understood how group-bound settings work in TeX.  For me it's always been a "try this and if it doesn't work, try something else" kind of thing.

eroux:
Looks ok to me, thanks a lot!

rpspringuel:
> though I might want to somehow emphasize (bold?) the fact that the additional dimension comes from the last commentary line and applies only to the next score, but perhaps that would stand out too much.

I did that in the **Nota Bene:** below the argtable.  Clearly that's not visible enough, so any suggestions on how to make it more visible?

henryso:
Perhaps move it above the table?  If it looks weird or breaks convention too much, I'm fine with it as is.  It was just a feeling, anyway.

rpspringuel:
Every other **Nota Bene:** appears below the argtable in the documentation, so I'm adverse to moving it above the table for just this one entry.  However, I'm not adverse to moving all of them if you think that would help.

henryso:
Let's leave it then; this looks good to me.

rpspringuel:
Question @henryso: Is it the intended behavior of `\gresetheadercapture` that it will effect all scores from the point of its invocation forward, or just the very next score?

henryso:
I suppose I had intended it be used from the preamble and would affect all score from point of invocation.  Is there some other way you believe it should be used?

rpspringuel:
No, looking it over that is about the way I would expect it to be used.  What's missing, and attracted my attention, is that there doesn't appear to be anyway to deactivate the header capture.  Once you start capturing a particular header you can reassign the capture to another macro (as is done in the tex-output/commentary test) but you can't tell GregorioTeX to stop capturing altogether.  This means the user who needs to stop the capture will either have to make use of group delimiters, or create a dummy macro which simply gobbles its arguments and redirect the capture to that macro.

Perhaps there should be a `\gresetheaderrelease` macro?

rpspringuel:
Actually, the group thing doesn't work.  I think at this point we've got so many `\globals` running around in our code (and they are by and large necessary to make things work the way we want them to) that we probably shouldn't expect group delimiting to work unless we're explicitly writing (and testing) for it.

eroux:
Indeed, the `\global`s should be tamed, but that's a huge work! I think this PR is ready for merge?

henryso:
Reviewing the comments, I think this is fine to merge.

PR #788, opened 2016-01-17
fix #765

### LINKED ISSUE #765: font scaling factor
I'm converting a gregorian font to Gregorio format for an Abbey, but I think it would be wrong to scale the glyphs in the font (as it makes a lot of approximations), so I think it would be best to make Gregorio open the font in a size multiplied by a `gre@fontscalefactor`. In my case this would be *0.173076293*. Gregorio should allow that easily, but I'm not sure about the best way, as `\setgregorianfont` already has an optional parameter...

eroux:
This fixes #765 and #789


eroux:
Thanks @henryso !

eroux:
See https://github.com/gregorio-project/gregorio-test/pull/101 for differences in tests, I think this is ready to be merged

eroux:
I'm not planning any more changes here, please review

eroux:
Sorry, I've just spotted a bug in the font creation (on `gof`), I'll push a commit in a few minutes

eroux:
Should be ok now, `fog` (and auctus variants) had a wrong width, and are lighter now, 3 tests have changed, they have been accepted in the corresponding PR

tex/gregoriotex-main.tex
```diff
@@ -1643,14 +1643,23 @@
...
+}%
+
+\def\gre@setgregoriofontscaled[#1]#2#3{%
+  \if\relax\detokenize{#1}\relax %
     \xdef\gre@gregoriofontname{#2}%
+    \expandafter\xdef\csname gre@fontfactor@#2\endcsname{#3}%
     \gre@loadgregoriofont%
     \gdef\GreCPVirgaReversaAscendensOnDLine##1{##1}%
   \else %
+    \directlua{texio.write_nl('test1')texio.write_nl('test1')}%
```
henryso: What is this for?

eroux: sorry, a debug symbol, I've pushed a fix

eroux:
If you think some code or coding style can be improved, don't hesitate to push and merge it without me reviewing (you should be able to test the custom font I did this code for)

doc/Command_Index_User.tex
```diff
@@ -394,6 +396,11 @@ \subsubsection{Score Font}
   & \texttt{parmesan} & Use the Parmesan font.\\
 \end{argtable}
 
+\macroname{\textbackslash gresetgregoriofontscaled}{[\optional{\#1}]\{\#2\}\{\#3\}}{gregoriotex-main.tex}
+
+This function is the same as above, with a third argument to scale the font. The fonts shipped with Gregorio do not need to use this function, but some custom fonts do. Note that you cannot use this to scale glyphs up or down, as they would not be placed correctly on the staff.
+
+The two first arguments are the same as \texttt{\textbackslash gresetgregoriofont}. The third argument is an integer representing the scaling factor, where the one used by \texttt{\textbackslash gresetgregoriofont} is 100000.
```
henryso: @rpspringuel Is it fine to describe the arguments this way for a user command, or should the argtable be repeated?

rpspringuel: I'd be inclined to repeat the argtable in general, but so long as the reference is to the command described immediately before (as it is in this case), I think this is acceptable.

henryso:
At a mechanical level, I am fine with this being merged, which I will do (unless it was already done) if there are no objections within the next couple of hours.

PR #794, opened 2016-01-18
TeXshop and TeXworks configuration scripts

rpspringuel:
I originally wrote these scripts to speed up installation testing on my VMs.  This PR robustifies them to make the more portable to other people's systems (including allowing the TeXworks one to work on Linux).

I've also endeavored to make them as user friendly as possible: the `.command` extension makes them double-click executable on Mac so that the user need not be familiar with Terminal.  This sort of behavior is preference controlled on Ubuntu, so with the right preference settings, the TeXworks configuration script also can be double-click run there.

TeXworks for Windows is not configured by the TeXworks configuration script because Bash doesn't run on Windows.  Suggestions on how to deal with this are welcome.


eroux:
Ok for me! This might be superseded by the integration of gregorio in `shell_escape_commands` in TeXLive 2016, but nothing is done yet

rpspringuel:
Parts of each script certainly will (in particular those parts which create/enable the LuaLaTeX+se typesetting tool).  The other parts, however, should remain valid and useful even beyond that point.

henryso:
I'm not the right person to review this, but @eroux has already given approval, so I'll merge.

PR #796, opened 2016-01-18
Corrected spacing and ledger lines for alterations outside the staff.

### LINKED ISSUE #790: alteration below staff lines
It seems alterations below staff lines could be improved, see for instance this result:

[bubug.pdf](https://github.com/gregorio-project/gregorio/files/93597/bubug.pdf)

from

```
name: test;
%%
aaa(axZ)aaa(gxg)
```

- the flat should be taken into account by the text-staff lines separation space algorithm (so that it doesn't eat the text
- there should be a ledger line around it

henryso:
Fixes #790.

With this change, I have switched from alteration as element/glyph with a shape-specific `gregorio_type` to a glyph with a `gregorio_glyph_type` of `G_ALTERATION` and whose note has the appropriate `gregorio_shape`.

In so doing, I eliminated some dead code (used for GregorioXML) and fixed a problem with point-and-click and newer versions of lualatex.

Tests have changed appropriately to this fix, and I added a new test to exercise the the ledger line/spacing fix.  I switched the `gabc-output` tests to use the point-and-click feature, which helps both with debugging as well as with exercising that feature itself (which has been broken since some version of TeX Live 2015).


eroux:
Looks perfect, thanks a lot!

Issue #787, opened 2016-01-16
Change the names of gsp-default space macros.


henryso:
During some of my recent updates, I've run into a problem differentiating between dimension registers and space macros.  For example, `\gre@dimen@glyphraisevalue` is a dimension register which can be scaled (`15.5\gre@dimen@glyphraisevalue` or `-\gre@dimen@glyphraisevalue`) and whose textual value requires the use of `\the`, but `\gre@dimen@annotationraise` is a textual space macro, which must be scaled using a `\dimexpr` or somesuch, and for which `\the` is illegal.  For a space macro, a leading `-` will work (because it just happens to be in the right place), but a leading number won't.

This is something that was obviously resolvable during testing, but made it a little bit more difficult during development.  Is it worth changing the prefix of spaces to make it more obvious?  For example, using `\grespace@dimen@annotationraise` (or somesuch) instead of `\gre@dimen@annotationraise`?

eroux:
Looks like a good idea!

henryso:
I'll work on this when development has calmed a bit, as this change is likely to cause merge issues.

rpspringuel:
I think it's a reasonable idea but would use something like `\gre@space@dimen@annoationraise` (if you want to emphasize how it is encoded) or `\gre@dimen@space@annotationraise` (if you want to emphasize what is encoded) instead.  The extra `@` makes it more difficult to type, but I find it makes it easier to read and I generally prefer code which is easy to read over that which is easy to type.

henryso:
I have no attachment to `grespace`.  Which would you prefer: `\gre@dimen@space` or `\gre@space@dimen`?

rpspringuel:
As I said, it's a question of emphasis.  To me at least `\gre@space@dimen` emphasizes that a spacing string is being encoded by placing `space` first.  `\gre@dimen@space` emphasizes that the macro represents a dimension (thought in text form).  Given that the problem you're having is based on the fact that spacing strings need to be treated differently in certain contexts, it would make sense to emphasize that aspect (i.e. use `\gre@space@dimen`).

PR #798, opened 2016-01-20
Switched to etex computation directives and added @space prefix to textual spaces.

### LINKED ISSUE #739: use `dimexpr` instead of `\advance`, `\multiply`, etc.
This is a long term issue with no deadline, but I was advised by @OldClaudio (Claudio Beccari) to use `\dimexpr` instead of the individual operations we currently use. As stated in [eTeX's manual](http://www.let.rug.nl/alfa/tex/tetex30/etex/base/etex-man.pdf#8), these operations are a bit safer and are sometimes a bit faster and more precise. We should certainly make new code use them, and convert some code from time to time...

henryso:
Fixes #739 and #787.

For `\advance`, `\multiply`, and `\divide`, I left macros which do not do their computation in one place.

Some of the tests were actually using the `@dimen` macros for demonstration purposes, so I updated those.  There were also some very slight output differences because (I suppose) the e-TeX directives round rather than truncate; these I have accepted.

### DIFF
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -1013,6 +1013,8 @@
 % This function converts a distance to the units indicated in #1 and returns it as a string.
 \def\gre@convertto#1#2{%
   \gre@debugmsg{general}{convertto (#1) (#2)}%
+  % \number\dimexpr returns the dimension in sp, so we must multiply #2 by
+  % 65536 (= 1pt in sp) before divding it by \number\dimexpr 1#1
   \edef\gre@converted{\gre@strip@pt\dimexpr#2*65536/\number\dimexpr 1#1\relax\relax #1}%
   \gre@debugmsg{general}{converted value is \gre@converted}%
 }%
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -1013,9 +1013,8 @@
 % This function converts a distance to the units indicated in #1 and returns it as a string.
 \def\gre@convertto#1#2{%
   \gre@debugmsg{general}{convertto (#1) (#2)}%
-  % \number\dimexpr returns the dimension in sp, so we must multiply #2 by
-  % 65536 (= 1pt in sp) before divding it by \number\dimexpr 1#1
-  \edef\gre@converted{\gre@strip@pt\dimexpr#2*65536/\number\dimexpr 1#1\relax\relax #1}%
+  % \p@ is equal to 1pt and is used here for a more accurate computation
+  \edef\gre@converted{\gre@strip@pt\dimexpr#2*\p@/\dimexpr 1#1\relax\relax #1}%
   \gre@debugmsg{general}{converted value is \gre@converted}%
 }%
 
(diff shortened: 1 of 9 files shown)


tex/gregoriotex-spaces.tex
```diff
@@ -792,18 +749,19 @@
...
 \def\gre@calculate@annotationtrueraise{%
   \gre@debugmsg{annotation}{Calculating the raise.}%
-  \global\advance\gre@dimen@annotationtrueraise by \gre@dimen@staffheight\relax%
   \gre@debugmsg{annnotation}{Added staff height.}%
-  \global\advance\gre@dimen@annotationtrueraise by \gre@dimen@spacebeneathtext\relax%
   \gre@debugmsg{annnotation}{Added space beneath text.}%
-  \global\advance\gre@dimen@annotationtrueraise by \gre@dimen@currenttranslationheight\relax%
   \gre@debugmsg{annnotation}{Added translation height.}%
-  \global\advance\gre@dimen@annotationtrueraise by \gre@dimen@spacelinestext\relax%
   \gre@debugmsg{annnotation}{Added spacelinestext.}%
```
rpspringuel: I'd get rid of/consolidate these debug messages now.  I needed them when writing this macro to debug a problem that was occurring in the math, but they're no longer necessary.  Having separate messages also tends to obfuscate the fact that all these calculations are now done in one step.

tex/gregoriotex-spaces.tex
```diff
@@ -1059,45 +1020,14 @@
...
 % This function converts a distance to the units indicated in #1 and returns it as a string.
-\newdimen\gre@unit%
-\newdimen\gre@base%
-\newdimen\gre@maxlen%
-\newcount\gre@unitfactor%
-\newcount\gre@basefactor%
 \def\gre@convertto#1#2{%
   \gre@debugmsg{general}{convertto (#1) (#2)}%
   \gre@debugmsg{ifdim}{ #2 = 0pt}%
   \ifdim#2=0pt\relax%
```
rpspringuel: See my comment in #739.  I don't think this `\ifdim` is needed anymore.

tex/gregoriotex-spaces.tex
```diff
@@ -1059,45 +1020,14 @@
 {\catcode`p=12 \catcode`t=12 \gdef\gre@makenum#1pt{#1}}%
```
rpspringuel: `\gre@makein` isn't needed anymore.

Might it make more sense to define `\strip@pt` in gregoriotex.tex (the way we do for a couple of other LaTeX primitives) and then use that instead of creating `\gre@makenum` which has an admittedly more complected usage syntax (`\expandafter\gre@makenum\the` vs. `\strip@pt`)?

rpspringuel: The appropriate definition from latex.ltx is:
```
\begingroup
  \catcode`P=12
  \catcode`T=12
  \lowercase{
    \def\x{\def\rem@pt##1.##2PT{##1\ifnum##2>\z@.##2\fi}}}
  \expandafter\endgroup\x
\def\strip@pt{\expandafter\rem@pt\the}
```

henryso:
Updated.  Thanks for the suggestions.

eroux:

henryso:
I added the requested comment.

henryso:
@rpspringuel after switching to `\@p`, I removed the comment about 65536 since it doesn't apply any more.

rpspringuel:
Somewhere in either Command_Index_User or _internal there's a note about the macros that correspond to each distance defined in the Distances section of User.  That should be updated with the new naming convention.

henryso:
OK, I made the two updates.

eroux:
Thanks, merging!

PR #799, opened 2016-01-20
Encoded the new shape for the curly brace.

### LINKED ISSUE #797: new drawing for curly brace
An anonymous contributor allowed Gregorio to use the following drawing for curly braces:

[accolade.eps.zip](https://github.com/gregorio-project/gregorio/files/95553/accolade.eps.zip)

Maybe the metapost-drawn curly braces could look more like it?

henryso:
Fixes #797.

Obviously, all tests will curly braces break, so I accepted the new result.


eroux:
Thanks! Do you think there is a way to keep the middle and the tails sharp so that they stay more or less the same when they're stretched?

henryso:
The middle, yes, and the code is doing that, but since I flattened the shape, it needs to get pretty wide before you can tell that it's doing something.  I'll tighten that up.  The tails are a bit harder, because then the shape looks less like a curly brace, but I'll see what I can do.

henryso:
This is how it looks with the middle, but not the tails kept sharp:
image

henryso:
This is how it looks with both the middle and the tails kept sharp:
image

henryso:
@eroux Which of the above do you prefer?  Or do you have any other suggestions?

eroux:
The second set looks perfect!

henryso:
I made the requested change.

rpspringuel:
One thing that might help retain the "curved" aspect of the brace as it stretches is to have it taper quicker as it heads towards the tails.  I don't know how easy that would be to pull off though, so feel free to ignore the idea if it's too difficult.

henryso:
That math is unfortunately beyond my meager abilities.  If anyone can help me formulate it better, I'm open.

rpspringuel:
Unfortunately I don't understand the MetaPost instructions well enough to provide any help on the math.

henryso:
It's pretty simple.  In terms of drawing a shape, it boils down to lines (denoted by `(x1,y1) -- (x2,y2)`) and cubic Bezier curves (denoted by `(x1,y1) .. controls (x2,y2) and (x3, y3) .. (x4, y4)`).  So the key is how to bias the curve towards the tail rather than just stretching the points out in the way that I did it.

rpspringuel:
@henryso, am I right in interpreting things that `p` is the parameter which positions the control/end points on the bottom of the curve and that it is proportional to the distance of the point to center of the brace?

Also, can you create a diagram of a generic brace which shows where each segment of the path begins and ends?

henryso:
`p` is one half the "stretchy distance" on the bottom of the shape and `q` is one half the "stretchy distance" on the top of the shape.  Take a look at the spreadsheet on #797 for the numbers (extrapolated from converting the .eps file to .svg and looking at the curves).

Here's a diagram that hopefully shows you what you asked for:
[Scratchpad - 20160120_224837.pdf](https://github.com/gregorio-project/gregorio/files/98638/Scratchpad.-.20160120_224837.pdf)

rpspringuel:
That image is what I was looking for.  Thanks.

As a first trial, I would raise the height of the points subject to `p` according to the value of `p` with larger values of `p` pushing points further away from the center up further.

Obviously the coefficient for this factor should be small and the overall factor should be 0 at the center (which is `p+0.08667`, if I understand what's going on).

Given that the curves already look pretty good up to 15mm ,I'd also consider raising `p` to a power higher than 1 so that the effect is stronger the larger the brace gets.  It may even make sense to use an exponential function to get a stronger "kneeing" action as the braces get bigger.  The problem, of course, will picking the parameters such that the effect kicks in like we want it to, but isn't so larger that the lower curve starts traveling above the upper curve for really large values of `p`.

To that end I'd try adding `0.005(x-(p+0.08667)^2)` (where `x` is the horizontal position of the point) to the vertical position of the appropriate points as a starting point and then tweak from there until something starts to look nice.  If that starting point is really ugly and it's not clear how to fix it quickly, then I'd simply abandon this as hopeless for now.

henryso:
Applying some of the techniques inspired by @rpspringuel, this is the best I can come up with.  Is this better than the previous result?

image

rpspringuel:
I certainly like this better.  The tails are still pretty sharp, but there's also more a a curved feeling to the ends that allows it to feel more like a curly brace at longer distances.

I'm still not 100% satisfied with the shapes at 75mm and above, but how often will people really need braces that long anyway?

henryso:
I tried raising and lowering the heights of the various points (which was one of your suggestions), but the control points need to be adjusted more "smoothly" than I was able to crudely guess; certainly doing it linearly produced a distorted figure.  I'll commit these latest changes.

rpspringuel:
I think on the really long braces it may be necessary to raise the top of the curve as well to get something really aesthetically pleasing.  The really long curves are just so gentle that lines look almost straight.

However, as I pointed out, braces that long are really going to be rare, and as you pointed out this sort of tweaking takes a lot of time and patience to do.  Given both of these I think we've reached (and perhaps passed) the point of diminishing returns, so if others are happy, I'm happy.

henryso:
I **completely** agree with you.  I certainly couldn't figure out a way to do it parametrically and achieve a good result.  Possibly higher order math, but that's not really my talent.

The biggest issue was that the "thickness" of the curving portion became distorted, with an inappropriate "bump" at the various nodes because control points toward the (non-scaling) center portion were not adjusting properly for the changes in the curve towards the tails.

I think this is certainly better than the first image posted above, and I also like it better than the second, but we'll see what everyone else thinks.

eschwab:
The latest image is the best yet. I do not see the need to tweak the long
braces.

eroux:
Thanks, merging!

PR #802, opened 2016-01-21
Added support for ad hoc enabling/disabling custos at forced line breaks.

### LINKED ISSUE #800: z0 and Z0 for line break without custos
Original thread [here](https://mail.gna.org/public/gregorio-users/2016-01/msg00007.html). It would be useful to allow line break with no custo with `z0` and `Z0`

henryso:
Fixes #800.

I also found/fixed a bug generating `Z` when writing gabc.

gabc-dump expectations are updated because of the new feature.  gabc-gabc expectations are updated because of the bug fix.  Everything else passes, and I added some new tests to exercise the new feature.


eroux:
Wonderful! Looks perfect to me

eroux:
Something puzzles me in your tests, not related to this feature (which can be merged anyway), on https://github.com/henryso/gregorio-test/blob/83d4cb4bd3eb49cde0b05b5cf4221fa4e70c2a66/tests/gabc-output/eol.pdf the final `-` doesn't appear in the text, where it should... Does gregorio remove it automatically? i think it's a mistake if it does... But it should be dealt with in another issue, beta1 doesn't need it fixed

henryso:
I didn't notice that, but yes, that would appear to be a bug (as you say, not related specifically to this pull request), possibly related to recent hyphenation changes.

PR #808, opened 2016-01-22
Fixes for Solaris build

### LINKED ISSUE #806: build fails on Solaris
See [this thread](http://tug.org/pipermail/tlbuild/2016q1/003339.html). This seems caused by the call to `realpath` (see [gnulib's description](https://www.gnu.org/software/gnulib/manual/html_node/realpath.html))

henryso:
Possibly fixes #806.

No tests change.

Please review, merge if OK, and submit to TeX Live for checking.


eroux:
Looks good, thanks a lot!

PR #810, opened 2016-01-22
4.1 documentation check

rpspringuel:
To prepare for 4.1 I looked over the documentation and found (and fixed) a few problems.


henryso:
Shouldn't this be into develop rather than master?

rpspringuel:
It should.  I'll fix that, assuming I can do so before the weather knocks out my internet connection on a more permanent basis (I'm already having it drop due to the snow).

PR #811, opened 2016-01-22
4.1 documentation check

rpspringuel:
To prepare for 4.1 I looked over the documentation and found (and fixed) a few problems.

### DIFF
--- a/doc/Command_Index_gregorio.tex
+++ b/doc/Command_Index_gregorio.tex
@@ -751,9 +751,6 @@ Opens the score.
   \#4 & \TeX\ code & Macros rendering the first syllable; should emit the initial and populate \verb=\gre@opening@syllabletext=.\\
 \end{argtable}
 
-\macroname{\textbackslash GreScoreReference}{\#1}{gregoriotex-main.tex}
-Currently does nothing.
-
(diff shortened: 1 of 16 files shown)


doc/Command_Index_gregorio.tex
```diff
@@ -751,9 +751,6 @@ \section{Gregorio Controls}
   \#4 & \TeX\ code & Macros rendering the first syllable; should emit the initial and populate \verb=\gre@opening@syllabletext=.\\
 \end{argtable}
 
-\macroname{\textbackslash GreScoreReference}{\#1}{gregoriotex-main.tex}
```
henryso: This may do nothing, but its use has not been removed.

henryso:
Other than that, it looks OK to me.

doc/Command_Index_gregorio.tex
rpspringuel: The function is marked as obsolete and will raise an error if used.  No other obsolete or deprecated functions are documented.

If it's supposed to be there (even doing nothing), then it shouldn't be raising an obsolete error.

henryso: Ok, then the mismatch is on the C side.  The C code still generates this.  I'll create an issue to take care of that.

PR #816, opened 2016-01-22
Modified to attempt to use gmake if make is not GNU-make.

henryso:
As requested by @mojca.


henryso:
This is copied from LuaTeX's build.sh.  I've reviewed it and it seems fine, so I'm merging it.

PR #819, opened 2016-01-23
Code cleanup

henryso:
For #697 

No tests change.

I expect to be creating more pull requests for this issue, but I'd rather merge often during the beta period than keep a single pull request for a long time.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -280,6 +280,7 @@ static void style_pop(det_style **first_style, det_style *element)
         return;
     }
     if (element->previous_style) {
+        assert(*first_style != element);
         element->previous_style->next_style = element->next_style;
         if (element->next_style) {
             element->next_style->previous_style = element->previous_style;
@@ -348,7 +349,7 @@ static __inline void verb_or_sp(const gregorio_character **ptr_character,
         }
     }
     if (i == 0) {
-        ptr_character = &current_character;
+        *ptr_character = current_character;
         return;
     }
     text = (grewchar *) gregorio_malloc((i + 1) * sizeof(grewchar));
@@ -433,7 +434,9 @@ void gregorio_write_text(const gregorio_write_text_phase phase,
             }
         }
 
-        current_character = current_character->next_character;
+        if (current_character) {
+            current_character = current_character->next_character;
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


eroux:
Thanks a lot for the discovery of `scan-build`, a very precious tool indeed! With the 3.8 version I still have these warnings after your changes:

```
/usr/share/clang/scan-build-3.8/bin/../libexec/ccc-analyzer -DHAVE_CONFIG_H -DBRANCH_VERSION='"henryso-fix-697-6c5d07c-3005"' -I.  -I../src -I../src/gabc  -I../src/dump  -I../src/gregoriotex -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2   -g -O2 -std=gnu89 -pedantic -fstack-protector-strong -fPIE -Wformat -Werror=format-security -Wstrict-prototypes -Wall -Wextra -MT gregorio-characters.o -MD -MP -MF .deps/gregorio-characters.Tpo -c -o gregorio-characters.o `test -f 'characters.c' || echo './'`characters.c
characters.c:720:63: warning: Use of memory after it is freed
        current_character->previous_character->next_character = element;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
characters.c:744:66: warning: Use of memory after it is freed
        (*current_character)->next_character->previous_character = element;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
characters.c:831:10: warning: Use of memory after it is freed
    if (!current_character->previous_character->is_character
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
characters.c:951:37: warning: Use of memory after it is freed
            suppress_this_character((*ptr_character)->next_character);
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
4 warnings generated.
```

is it normal?

henryso:
Those four boil down to `suppress_this_character`, which frees the character without clearing the pointer passed into it.  However, in all instances where this is called, it's called with `character->next_character` or `character->previous_character` or some lone character.  In the first two cases, the function should work properly, merging the links; in the third case, the pointer is set to `NULL` on return.

However, this function has a potential to cause a problem if the list is not linked correctly (i.e., if `character->previous_character->next_character` is not `character`), and I think that's what's being flagged here.

I'm not quite sure how to "fix" this problem (and I tried a few things last night), and I'm not sure it's worth it because it seems to be doing the right thing as long as our list is proper, but I'll think about it some more.  It's possible that we need three functions (suppress this, suppress previous, suppress next) which all alter the initial pointer; is that worth doing?

eroux:
I think it's ok then, thanks a lot for the explanation!

PR #820, opened 2016-01-23
Fix a memory leak in nabc handling.

jakubjelinek:


eroux:
Merging, thanks a lot!

PR #822, opened 2016-01-23
fix #821

### LINKED ISSUE #821: -g -O2 in CFLAGS?
See [this comment](https://github.com/gregorio-project/gregorio/issues/697#issuecomment-174176208): `./build.sh` does not add `-g -O2` to `CFLAGS`. 

A first guess was that we should call `AC_PROG_CC` instead of `AC_PROG_CC_C89`, the latter being called by the former according to [the manual](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/C-Compiler.html). But using `AC_PROG_CC` doesn't seem to fix the issue...

eroux:
- add relevant warnings from build.sh into configure.ac
- remove --warn option of build.sh
- remove hardcoded warning from build.sh (fixes #821)
- AC_PROC_CC_C89 -> AC_PROG_CC

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -47,12 +47,9 @@ OTHERARGS=
 FORCE_AUTORECONF=
 FORCE_FONTS=
 
-CFLAGS="$CFLAGS -Wdeclaration-after-statement"
-
 until [ -z "$1" ]; do
   case "$1" in
     --mingw     ) MINGWCROSS=TRUE ;;
-    --warn      ) WARN=TRUE ;;
     --host=*    ) CONFHOST="$1" ;;
     --build=*   ) CONFBUILD="$1" ;;
     --arch=*    ) MACCROSS=TRUE; ARCH=`echo $1 | sed 's/--arch=\(.*\)/\1/' ` ;;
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -28,7 +28,7 @@ AC_CONFIG_MACRO_DIR([m4])
 
 AM_INIT_AUTOMAKE([-Wall -Wno-portability subdir-objects foreign dist-bzip2 no-dist-gzip])
 
-AC_PROG_CC_C89
+AC_PROG_CC
```
henryso: We're still targetting C89, though...  Is this correct?

eroux: Yes, but as described in #821, `AC_PROG_CC` calls `AC_PROG_CC_C89`, the inclusion in the other direction is not documented

build.sh
```diff
@@ -93,7 +82,7 @@ then
   fi
   OLDPATH=$PATH
   PATH=/usr/$MINGWSTR/bin:$PATH
-  CFLAGS="-mtune=pentiumpro -msse2 -O2 $CFLAGS"
+  CFLAGS="-mtune=pentiumpro -msse2 -g -O2 $CFLAGS"
```
henryso: Should `-g` be forced?  Isn't that only to include the symbol table for debugging?

eroux: I added that mainly for coherence, as 99% of others systems will use these flags, so why not mingw. It can be removed of course, I don't have a very strong opinion

henryso: Ok, no one has complained thus far, so let's leave it for now.

henryso:
If you are good to go, I'm fine with this.

PR #824, opened 2016-01-23
Code cleanup

henryso:
For #697.

In this edition, I have fixed the `suppress_this_character` static check warnings by adding strategic assertions and code reordering.  Using `-fsanitize=undefined,address` found a potential issue in sha1.c, which I've fixed by forcing alignment with a memcpy to a stack variable.

No tests change.

Same caveats as before; I think it's better to merge these kinds of fixes sooner than to have a big, long-lived pull request.


eroux:

PR #825, opened 2016-01-23
4.1 documentation check

rpspringuel:
Since the `initial-style` header is deprecated, it doesn't need to be in the docs anymore.


henryso:
Thanks.

PR #839, opened 2016-01-27
Implemented elision support.

### LINKED ISSUE #838: elisions
Solesmes asked for a way to input elisions, syllables that are not sung in order to keep the same syllable number across sentences. These elisions are noted in italic.

A proposal is to use a new text markup `<e>`, for instance `clar<e>a</e>` would ignore the second a in text centering (and `<e>cla</e>ra` would ignore the first), and the second `a` would be in `elisionstye`, which would be italic by default.

What do you think?

henryso:
Fixes #838.

All current tests pass.  I added a new test for this feature.


eroux:
Looks perfect, thanks a lot!

PR #849, opened 2016-01-29
Implemented customizable special characters.

### LINKED ISSUE #846: more generic system for special characters
Solesmes is planning to input special characters not currently handled by Gregorio, like a H with bar, or a custom TQ and TP they'll have in their font. I think a good solution would be to have a macro like `\grespecialcharacter{H/}{\mySpecialHBar}`, and use if to catch all "special characters", and if a special character is not present, it would simply be printed as is (e.g. `H/` in previous case). What do you think?

henryso:
Fixes #846.

This should be fully backward-compatible.  The only tests which fail are the gabc-gtex tests that emit special characters, which is expected.  I also added a test specifically for this enhancement.


eroux:
Thanks a lot! I'll try that soon

eroux:
Works perfectly with active catcode for `'`, merging

PR #852, opened 2016-01-29
Corrected the name of the "op" fonts to be different than the normal fonts.

### LINKED ISSUE #851: "op" fonts should have different names than "non-op" fonts
This is a bug that was introduced when the names started being pulled from the base fonts.

henryso:
Fixes #851.

This bug can cause the luatex font cache to point greciliae at greciliae-op (and same with the other fonts), so this fix actually fixes tests that might break if the luatex cache has been messed up.  No tests change and all tests pass.


eroux:
Thanks!

PR #854, opened 2016-01-30
Added support for coverage testing with gcov/lcov.

henryso:
For #697.

In preliminary testing, I also found/fixed a bug in the vowel-subsystem buffer allocation which is not currently covered by the tests.

Note: I haven't yet added the new tests that will result from further coverage testing as part of gregorio-project/gregorio-test#112 and supported by this code.  Those will come as additional pull requests, along with fixing of any bugs found in testing.

All current tests pass given the caveat above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,11 @@
 *.lo
 *.la
 
+# Coverage
+*.gcno
+*.gcda
+/coverage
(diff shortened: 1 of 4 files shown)


eroux:
Looks good, thanks a lot!

PR #855, opened 2016-01-31
Fixed problems found while testing for code coverage.

henryso:
- Enhanced the coverage script for usability.
- Corrected to find all vowel files in TEXMF.
- Added lcov tokens to mark code that should not be reachable unless there is a
  programming error, a system fault, or gregorio runs out of memory.
- Added an assertion message level for programming assertions.
- Added some assertions based on coverage testing.
- Fixed missing nlba and/or euouae in dump and gabc output.
- Fixed bug with fusing down to a podatus.
- Corrected spelling of fusible.
- Changed gregorio_fix_style states to be an enum.
- Removed dead code.
For #697.

With this and the gregorio-project/gregorio-test#119, we are at 95.9% function coverage and 87.4% line coverage, not counting the flex- and bison-generated code (which I will analyze later).

There is still much work to do, since this preliminary result indicates there are a number of cases not covered by the tests and that gregorio-test will need to be enhanced for "should fail" tests and tests that involve running the gregorio executable to do things like get the usage or pass bad filenames, etc.

Please review this one carefully, as I have done quite a bit of refactoring, especially with respect to assertions.  The tests do pass in an acceptable manner (at least as far as I can tell).

If this is fit to merge, please merge it (and the corresponding tests).  As a modus operandi, I continue to believe smaller merges are better than a huge one for this particular issue.

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -46,12 +46,18 @@ EOT
     rm -rv coverage.info coverage
     echo "generating report"
     lcov $extra_args --directory . --capture --output-file coverage.info &&
+        lcov --remove coverage.info '*-y.c' '*-l.c' --output-file coverage.info &&
         genhtml coverage.info -o coverage
 
     if test -f __gcovtool.sh
     then
         rm -v __gcovtool.sh
     fi
+    ;;
+reset)
+    echo "resetting coverage files"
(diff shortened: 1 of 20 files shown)


eroux:
That's a great improvement, thanks a lot!

Issue #803, opened 2016-01-21
clivis stem length


eroux:
Clivis stem length may vary according to books, `squarize.py` needs to be able to output different types of stem length.

eroux:
(well, there are some incoherences I'm a bit puzzled about in this image, so it's going back to the drawing board, sorry...)

henryso:
Could it be that the "longqueue"-ness of the line depends on the position of the bottom of the line rather than the note at the top of the line?

eroux:
You're 100% right, I didn't realize that! Anyways, the global idea is:

- for deminutus, have same queue length as if it was not a deminutus figure
- for non-deminutus, have the queue slightly below the bottom of the second note

eroux:
Things are not 100% decided yet, but here are some features that could be implemented already:

- longueue-ness always depends on the second note
- apply the same exception for porrectus with first ambitus of 1 as for clivis: when the second note is between lines (`fgf`), the queue must go beneath the line (just like clivis)
- when the second note is on a line, the queue must go slightly beneath the second note (so there is a difference with the case where the second note is between lines, which is not fully decided yet)
- deminutus (`gf~`) follow globally the same rule as clivis (but details are not all decided yet)
- ancus follow the same rule as clivis (third note not taken into account except in particular exceptions, details to come)
- salicus, pes quadratum, etc. follow the same rule as clivis for stem length (details to come)
- queues should be shorter for lower notes (details to come)

Options for cases where the second note is between lines (`hg`):
1. keep the same length as we have now (if possible a bit lower, but that's not very important)
2. cross the bottom line
3. stop the stem in the middle of the second note

One of these options will be chosen, but it's good to keep that in mind, and, if it's simple (which I doubt, but who knows), users could choose between these options.

I think several points can be discussed before having all the details:

1. first, does everyone agree with those changes? It seems the two big projects using Gregorio agree, but Gregorio should remain independent!
2. what the best way to discuss that? Here in this topic? A very complete example should be built with all these stems so that we can "draw" the correct stem length on it, but I don't know what would be the best way to do that
3. all these new glyphs for porrectus (even just with ambitus one), ancus, salicus, clivis, pes quadratum, etc. will be many, will it fit in PUA? If not, can we take other private use parts (not supported in old LuaTeX), or should we use the fuse system?

henryso:
To answer your points directly:
[...]

PR #857, opened 2016-02-01
Fixed problems found in coverage testing.

henryso:
- Made coverage script quieter when cleaning a clean directory.
- Found and marked more assertions.
- Corrected determination of gvgvv, gvvgv, gsgss, and gssgs.
- Improved error checking of score lexer (fixed unexpected ECHO).
- Corrected gabc output of quadratum shapes.
- Removed dead gabc cases.
- Corrected morphing of oriscus added to cavum.
- Fixed double clef rendering on the same line.
- Fixed missing translation center gabc output.
- Removed determination of G_TRIGONUS since it is ambiguous/inaccurate.
- Added a check for unclosed braces, ledger lines, and slurs.
For #697.

An additional note about G_TRIGONUS.  We were detecting things like `(GHG)` and `(HGH)` as trigonus, but this is not accurate (i.e., trigon is also typically something like `(HHG)`) and was playing strangely with proper detection in sequences like `(GHGF)`, so I removed it, which only affects the glyph type and doesn't affect the final (TeX) output.

With gregorio-project/gregorio-test#121, this brings coverage to 97.4% functions and 92.2% lines.  I've discovered that the lines generated from flex and bison rules are not tested for coverage, so I no longer intend to include the `*-y.c` and `*-l.c` files in coverage testing.

There is still a good deal of work to do:
- Possibly remove the "multi-staff" polyphony vestiges in the C code (see my comments in #697).
- Create a test harness for running non-gabc tests against the executable
- More test coverage; much of what remains has to do with obscure sequences of notes
- Improved coverage in the code related to long/short queues.  I'd like to roll in the changes for #803 before concentrating too much on this.

Please double-check my assertions, review these changes, and merge them (along with the corresponding gregorio-test pull request) if satisfactory.

### DIFF
--- a/src/gabc/gabc-write.c
+++ b/src/gabc/gabc-write.c
@@ -463,22 +463,22 @@ static void write_note_heuristics(FILE *f, gregorio_note *note) {
  */
 
 static void gabc_write_gregorio_note(FILE *f, gregorio_note *note,
-        char glyph_type)
+        const bool is_quadratum)
 {
     char shape;
     gregorio_assert(note, gabc_write_gregorio_note, "call with NULL argument",
             return);
     gregorio_assert(note->type == GRE_NOTE, gabc_write_gregorio_note,
             "call with argument which type is not GRE_NOTE", return);
-    if (glyph_type == G_PES_QUADRATUM) {
-        shape = S_QUADRATUM;
-    } else {
-        shape = note->u.note.shape;
-    }
+    shape = note->u.note.shape;
     switch (shape) {
         /* first we write the letters that determine the shapes */
     case S_PUNCTUM:
-        fprintf(f, "%c", pitch_letter(note->u.note.pitch));
+        if (is_quadratum) {
+            fprintf(f, "%cq", pitch_letter(note->u.note.pitch));
+        } else {
+            fprintf(f, "%c", pitch_letter(note->u.note.pitch));
+        }
         break;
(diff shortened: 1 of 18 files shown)


src/gabc/gabc-write.c
```diff
@@ -491,11 +491,7 @@ static void gabc_write_gregorio_note(FILE *f, gregorio_note *note,
         }
         break;
     case S_PUNCTUM_INCLINATUM_AUCTUS:
-        if (note->next) {
-            fprintf(f, "%c<", toupper((unsigned char)pitch_letter(note->u.note.pitch)));
-        } else {
-            fprintf(f, "%c", toupper((unsigned char)pitch_letter(note->u.note.pitch)));
-        }
+        fprintf(f, "%c", toupper((unsigned char)pitch_letter(note->u.note.pitch)));
```
eroux: shouldn't it be "%c<"? Or is the liquescentia written later?

eroux:
Apart from my comment, it looks good, thanks a lot!

src/gabc/gabc-write.c
henryso: Liquescentia is written later.

eroux: Ok, merging then. This must have required a lot of patience, but it was definitely worth it, that's a very important improvement!

PR #859, opened 2016-02-02
Fixed problems found during coverage testing:

henryso:
- Removed the non-working multi-voice code.
- Corrected the determination of the ancus.
- Added more assertions.

For #697.

There are still a few more things to cover.  I need to add a few more command line option tests, and there are a number of edge cases that I have yet to find the exact formulas to tease out.  Of course, I will have to re-run the coverage tests after #803 is merged, and then I need to do some valgrind testing against the new test cases.

Along with gregorio-project/gregorio-test#122, this gets us to 94.7% line coverage and 99.5% function coverage.


eroux:
Thanks a lot!

PR #865, opened 2016-02-03
Fixed problems found during coverage testing:

henryso:
- Corrected problems in command line argument handling.
- Corrected problems rendering empty scores.
- Restored rendering of manual custos immediately after a bar.
- Fixed problems rendering styles around a forced center.
- Removed dead code.
For #697.

The tests I constructed in this phase highlighted two issues related to center determination (the `rebuild_characters` function).  The first is a problem with styles that surround a forced center.  This is fixed in the pull request.  The second is a problem when the initial is centered.  I did not have time to fix that for this pull request, but I will work on that next.

I rewrote the `rebuild_characters` function to hopefully de-spaghettify it a bit and help with these two fixes, so that needs a bit of careful review.

With the corresponding tests in gregorio-project/gregorio-test#124. this brings us to 96.8% line coverage and 100% function coverage.  However, #803 will bring in some new code and I still need to do valgrind and sanitize testing as well.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -391,11 +391,9 @@ static __inline void verb_or_sp(const gregorio_character **ptr_character,
     while (j < i) {
         if (current_character->is_character) {
             text[j] = current_character->cos.character;
-            current_character = current_character->next_character;
             j++;
-        } else {
-            current_character = current_character->next_character;
         }
+        current_character = current_character->next_character;
     }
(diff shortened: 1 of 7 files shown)


eroux:
Thanks a lot, I think argument parsing was something that really needed attention for security! 100% function coverage is quite impressive!

Issue #868, opened 2016-02-04
Bug in element determination


henryso:
I found a major and long-standing bug in the code.  In `gabc-elements-determination.c`, there is this code:

```
        if (is_puncta_ascendens(current_glyph->type)) {
            current_glyph_type = G_PUNCTA_ASCENDENS;
        } else {
            if (is_puncta_descendens(current_glyph->type)) {
                current_glyph_type = G_PUNCTA_DESCENDENS;
            } else {
                current_glyph_type = current_glyph->type;
            }
        }
```

The problem is, `current_glyph_type` is being tested against `gregorio_glyph_type` values but `current_glyph->type` is of type `gregorio_type`, and in this case is *always* `GRE_GLYPH` (which has the same numeric value as `G_2_PUNCTA_INCLINATA_DESCENDENS`).

So this code is *unconditionally* setting `current_glyph_type` to `G_PUNCTA_DESCENDENS` before making any further element determination.

If I fix this by changing the code to
```
        if (is_puncta_ascendens(current_glyph->u.notes.glyph_type)) {
            current_glyph_type = G_PUNCTA_ASCENDENS;
        } else {
            if (is_puncta_descendens(current_glyph->u.notes.glyph_type)) {
                current_glyph_type = G_PUNCTA_DESCENDENS;
            } else {
                current_glyph_type = current_glyph->u.notes.glyph_type;
            }
        }
```
... it changes **A LOT** of tests.  Do I fix it?

This code has literally been doing this since *inception* (see https://github.com/gregorio-project/gregorio/blob/588be00ed31072c8d0c2ef14d382718bdd130ecf/plugins/gabc/gabc-elements-determination.c).

henryso:
Note that doing this also completely breaks fusion because fusion was written with the assumptions of how the code was working, so if we decide to fix this, then we would have to make fusion work again.

Alternately, I can distill out the current behavior (as opposed to the original code's intent) and make the intent of the new code the same as the effect of the old code.

eroux:
Interesting finding! Apart from fusion, does fixing change a lot of pdf output? My first thought would be to fix it and the fusion, but I'll take a closer look at the results.

eroux:
I think your change is not enough to make everything work, I have a lot of

```
in function `close_element': assertion:previous element was not closed
```

when I try it

henryso:
[...]

PR #869, opened 2016-02-04
More fixes for problems found in code coverage testing.

henryso:
- Fixed rendering of a first syllable with a forced center at the beginning.
- Corrected closing and reopening of styles when a mid-stack style is closed.
- Removed GRE_SYLLABLE enum and type from gregorio_syllable since there is no
  variation in syllable type.
- Removed more vestigial multi-voice support code.
- Simplified gregoriotex_fix_style.
- Removed some dead code and corrected some assertions.
For #697.

A note about gregoriotex_fix_style.  The old code was being careful to look for style-center-style, but it wasn't handling close versus open style and this extra effort to look for "center" is actually unnecessary since center doesn't affect the overall style of the syllable when determining the fixed style.

Along with the corresponding tests (gregorio-project/gregorio-test#130), this gets us to 98.0% line coverage (and 100% function coverage, as before).  A good part of the missing coverage is the due to the bug in #868 never exercising all the code paths.  Other than that, exercising the paths is leading to more and more obscure edge cases.  Strangely enough, I don't feel that we are quite at the point of diminishing returns, as evidenced by the bugs found and fixed in this pull request.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -541,9 +541,13 @@ void gregorio_write_first_letter_alignment_text(
             switch (current_character->cos.s.style) {
             case ST_CENTER:
             case ST_FORCED_CENTER:
-            case ST_INITIAL:
             case ST_FIRST_SYLLABLE_INITIAL:
                 /* ignore */
+                break;
+            case ST_INITIAL:
+                /* in the case where the initial is closed before the first
+                 * letter */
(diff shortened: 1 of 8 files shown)


eroux:
Excellent job, thanks a lot!

PR #875, opened 2016-02-05
Fixed long-standing element determination bug.

### LINKED ISSUE #868: Bug in element determination

henryso:
Fixes #868.

This pull request does not implement the interglyph space change.  That will be a different pull request.


eroux:
That looks perfect, I'm really glad this bug is fixed, even though it has no big consequence on the output! Just a question: I think it's good if puncta inclinata always stay together in the same element (even weird figures like `GFGFG`), it seems to be the case with the code, I just wanted to ask if it was really the case before merging?

henryso:
Yes, puncta inclinata are kept together.  This was actually working before because the code *always* thought *everything* was a punctum inclinatum.

eroux:
Ok great!

PR #876, opened 2016-02-05
Added support for inter-glyph space.

### LINKED ISSUE #868: Bug in element determination

henryso:
Fixes #868.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
 - It is now possible to generate a font with a different stem length schema (apart from the default one, Solesmes' is available), see [fonts/README.md](fonts/README.md) for details.
 - Some forms (`ed` for example) will have a different stem length if a ledger line is present. Heuristics to determine the presence of a ledger line or not are quite simple and may not always be correct. Gabc now allows you to force a note to act as if a ledger line was present with `[ll:1]`, or as if there was no ledger line, with `[ll:0]`. See Gabc section of GregorioRef for details.
+- The ability to typeset an inter-glyph space using `\!` in gabc (see [#868](https://github.com/gregorio-project/gregorio/issues/868))
```
eroux: I think you mean `/!` instead of `\!`?

henryso:
Thanks for noticing.

eroux:
I'm just wondering if it would be better to have `SP_INTERGLYPH_SPACE` explicitely in the structure when relevant, like in `(gg)`?... This would make gabc output use it every time... But I'm not really sure, what do you think?

henryso:
I don't really see the benefit of adding it into the structure.

eroux:
Ok, let's merge it then, thanks a lot!

Issue #873, opened 2016-02-05
doc doesn't build in develop


eroux:
When trying to build branch, I get a few gregorio warnings:

```
976: in function `gregorio_rebuild_characters': assertion:encountered unexpected begin ST_CENTER
```

then it fails on

```
../src/gregorio -o veni.gtex ./veni.gabc
657: in function `gabc_y_add_notes': assertion:attempted to append notes
```

henryso:
I think this is fixed.  At least `make distcheck` works for me.

eroux:
Thanks! I don't think it's worth opening an issue for that, but I have a warning during build (coming from before this merge):

```
gregoriotex/gregoriotex-write.c: In function ‘fusible_queued_shape’:
gregoriotex/gregoriotex-write.c:521:12: warning: ‘name’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     return compute_glyph_name(glyph, name, LG_NONE, true);
            ^
gregoriotex/gregoriotex-write.c: In function ‘gregoriotex_determine_note_glyph_name’:
gregoriotex/gregoriotex-write.c:585:17: warning: ‘name’ may be used uninitialized in this function [-Wmaybe-uninitialized]
                 gregorio_snprintf(buf, sizeof buf,
                 ^
```

eroux:
Also, I still have

```
 [189]976: in function `gregorio_rebuild_characters': assertion:encountered unexpected begin ST_CENTER
1017: in function `gregorio_rebuild_characters': warning:encountered unexpected end ST_CENTER
976: in function `gregorio_rebuild_characters': warning:encountered unexpected begin ST_CENTER
1017: in function `gregorio_rebuild_characters': warning:encountered unexpected end ST_CENTER
976: in function `gregorio_rebuild_characters': assertion:encountered unexpected begin ST_CENTER
1017: in function `gregorio_rebuild_characters': warning:encountered unexpected end ST_CENTER
976: in function `gregorio_rebuild_characters': warning:encountered unexpected begin ST_CENTER
1017: in function `gregorio_rebuild_characters': warning:encountered unexpected end ST_CENTER
```

during compilation

henryso:
Working on it.

henryso:
It's very strange.  If I run that snippet as a gabc, it works fine.

henryso:
[...]

PR #879, opened 2016-02-05
Fixed problem dealing with text before the initial clef.

henryso:
Completes the fix for #873.

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -587,6 +587,7 @@ end
 local function direct_gabc(gabc, header)
   tmpname = os.tmpname()
   local f = io.open(tmpname, 'w')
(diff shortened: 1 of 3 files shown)


tex/gregoriotex.lua
```diff
@@ -587,6 +587,7 @@ end
 local function direct_gabc(gabc, header)
   tmpname = os.tmpname()
   local f = io.open(tmpname, 'w')
+  gabc = gabc:match('^()%s*$') and '' or gabc:match('^%s*(.*%S)')
```
eroux: I'm having difficult time undestanding what this does... could you just add a short comment above?

eroux:
Apart from the small comment, this can be merged, thanks a lot!

henryso:
Ok, added.

eroux:
Thanks, merging!

PR #881, opened 2016-02-06
Fixes #871

### LINKED ISSUE #871: not taking punctum mora into account in other spacings
Following #795, it seems punctum mora should neither be taken into account when determining the space between notes of different syllables, for example:

punctummora

has too much spacing between *su-* and *a*. This new behavior should be optional, maybe integrated in the option set up for #795 (renaming it).

eroux:
- don't take entirely into account punctum mora in hozirontal spacing between syllables
- fix a rare bug with previous way of ignoring punctum mora
- document GrePunctumMora

I'm not entirely sure about the default values, but they're easily tunable. I think it's good if it does it in beta2, even if we change the default values later


eroux:
See https://github.com/gregorio-project/gregorio-test/pull/138 for the changes in the tests

henryso:
Since this is a behavior change (i.e., it breaks PDF tests), perhaps a section is needed in UPGRADE.md for people who prefer the older behavior?

eroux:
Right, I'll add it

eroux:
done

rpspringuel:
Just adding a cross link to #871.

PR #885, opened 2016-02-07
Code coverage update

henryso:
- Removed warnings that are no longer relevant.
- Unified MAX_AMBITUS and MAX_INTERVAL values.
- Added debug command line option.
- Added display of kpathsea version when compiled with kpathsea.
- Excluded gregorio_readline (unused) when compiled with kpathsea.
- Added and corrected assertions.
- Removed dead code.
- Fixed buffer-use-after-free (bug in repercussion pointers) and buffer-overrun (with bad unicode).
- Switched from the kpse compatibility API to the newer kpathsea API.
- Added a kpathsea check for writing the error file.
- Updated the copyright (and moved the string to the top of the file).
For #697.

A few notes:

I have run the system through sanitize and valgrind as well, which revealed a couple of bugs and a problem with kpathsea.

With the accompanying tests (gregorio-project/gregorio-test#139), this gets us to 100% coverage*, both with kpathsea and without kpathsea.  The * indicates that this is with the exclusion of
1. certain lines (marked with LCOV tokens), because they are not reasonably testable or are assertions that should not be hit
2. bison- and flex- generated files because the key parts for testing (the code in the actions) are skipped by the coverage tools.

I switched to the newer kpathsea API because valgrind detected a memory leak coming from a call to `kpathsea_name_ok`, so I did some research and found the `kpathsea_finish` function which purportedly cleaned up, but would require a switch to the new API.  A few grueling hours later, I completed the switch, and valgrind found that the leak had worsened.  I tracked this down to the fact that the kpathsea library (6.2.1-20150521) included in Gentoo has `KPATHSEA_CAN_FREE` set to `0`, which is likely the cause of the original leak with the old API.  This means that running sanitize or valgrind against the executable with kpathsea linked in on my computer shows every test having memory leaks.  Perhaps if one you has a kpathsea than can free, you can try the valgrind/sanitize tests against a `--with-kpathsea` build.

I decided to leave the new code in since it uses the newer API and because, in so doing, I updated the copyright and added the kpathsea check for the error file which was missing earlier.  The kpathsea switch is in a different commit, so I can exclude that easily if you feel that I should.

I have not done a new static analysis (with scan-build), because it's getting pretty late.  I'll see if I can do that tomorrow (well, today actually), but it's Chinese New Year (and the last weekend before Lent), so I'll be at a party for much of the day.

Please review and merge if satisfactory, or let me know if there's anything I should change.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -71,11 +71,15 @@ static bool read_vowel_rules(char *const lang) {
 #ifdef USE_KPSE
     filenames = kpse_find_file_generic(VOWEL_FILE, kpse_tex_format, true, true);
     if (!filenames) {
+        /* It's not reasonable to break kpse in such a way that this would
+         * fail. */
+        /* LCOV_EXCL_START */
         if (strcmp(language, "Latin") != 0) {
             gregorio_messagef("read_patterns", VERBOSITY_WARNING, 0,
                     _("kpse_find_file_generic cannot find %s"), VOWEL_FILE);
         }
         return false;
+        /* LCOV_EXCL_STOP */
     }
 #else
     FILE *file;
@@ -331,16 +335,13 @@ static void style_pop(det_style **first_style, det_style *element)
 }
 
 /*
- * free_styles just free the stack. You may notice that it will never be used
- * in a normal functionment. But we never know...
+ * free_styles just frees the stack.
  */
 
 static void free_styles(det_style **first_style)
 {
     det_style *current_style;
-    if (!first_style) {
-        return;
-    }
+    gregorio_not_null(first_style, free_styles, return);
     current_style = (*first_style);
     while (current_style) {
         current_style = current_style->next_style;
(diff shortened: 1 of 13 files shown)


eroux:
Good job! I can really imagine how frustrating this must have been... Under Debian, `KPATHSEA_CAN_FREE` is also 0... @kberry, @norbusan, what's the best way to deal with that? Should we consider the memory leaks in *kpathsea* normal?

kberry:
yes, as i thought i already wrote to you (somewhere), i did not worry about losing a few bytes of memory from kpathsea, preferring code readability, and so such leaks are widespread in the source.  even in 1988, no one complained about an actual memory problem. just get the checkers to ignore kpathsea. i hope you can.

KPATHSEA_CAN_FREE -- i think this is a red herring, and is always 0.  it is something peter did, and apparently never finished. he never wrote me about it (that i recall/can find) and in fact i don't recall ever hearing about it before.  perhaps i will rip it back out.

as for the api's: there's no problem using either kpathsea_* or kpse_*. neither api will ever go away, as far as i'm concerned.

happy hacking.

henryso:
I ran the static analysis using clang 3.7.1 and eliminated the problems it found by inlining the problematic functionality to make it the code intent clear to the analyzer.  There are three remaining problems in the flex-generated code, which I am happily ignoring.

In the accompanying tests, I added suppression for the kpathsea leaks.  The only one that concerns me is the array of strings returned by `kpathsea_find_file_generic`, which our code is responsible for freeing.  I believe I am freeing it correctly, but it's hard to find the "wheat" among the "weeds" when suppressing the calls from kpathsea.

Code coverage is still at 100%.  In non-kpathsea mode, there are no memory leaks detected by Valgrind or AddressSanitizer and no undefined access detected by AddressSanitizer.  In kpathsea mode, with the suppressions mentioned earlier, the result is the same.

Please review the changes and merge if satisfactory.

eroux:
Looks good, merging. Thanks @kberry for your answer!

PR #886, opened 2016-02-07
Fix 858

eroux:
- make porrectus first note higher
- make porrectus second note higher if the second ambitus os more than one
- fix small bugs in the shape and height of porrectus with long strokes
- adjust porrectus stem length
- very small adjustment of pes, and third note of porrectus but should be almost invisible


henryso:
Looks good.

PR #887, opened 2016-02-07
Fix #882

### LINKED ISSUE #882: ledger line piercing flats
In this image, the ledger line in "piercing" the flat:

bxb

I admit I hoped nobody would ever used this... but the Antiphonale Monasticum project does!

I don't think it's easily fixable, but there should be a way to fix it by hand, for instance by specifiying 

- first a ledger line embracing all the glyphs
- then all the glyphs, but without the ledger lines

Once there's a way to manually do that (in gabc), then we can think about making the C code output spot this case and output the correct TeX, but I think we can keep that part for later...

eroux:
I've also tried some performance improvements, I didn't really take time to see the performance gain though.


henryso:
I'm no expert in TeX, but this looks OK to me.

eroux:
Ok thanks, merging

PR #888, opened 2016-02-07
Feature/miktex

rpspringuel:
This PR updates the Windows installer in several ways:

- The `--admin` option is removed for MiKTeX installations, allowing the installer to work better on Windows 10 (where admin rights were apparently an issue).
- MiKTeX installations no longer copy files into TEXMFLOCAL but instead registers the texmf folder which the installer creates with MiKTeX, allowing it to use those files in place.
- I've added an uninstall script which will run as part of the uninstall action and will remove the Gregorio executable and TeX files from their copied locations (at least using the `unins000.exe` file which is created in the program directory when the program is installed, I haven't tested using the Add/Remove Programs feature built into Windows).

There are two other things which I'd still like to do to the installer, but which may wait until after a beta2 relaease (and thus get their own PR):

- Consolidate the install-tl and install-mk scripts into a single script.  If you look at the uninstall script, you can see I figured out a way to distinguish between these two installations, allowing for just a single script.
- Modify the PATH variable instead of copying the executable.  I've found some instructions online for how to do this in InnoSetup, but haven't had time to implement and test them yet.

Testing is, at this point, a bit spotty.  I know everything works on Windows 10 with MiKTeX installed, but haven't had the chance to double check anything else yet.  Unfortunately I don't have time to do any more testing today, nor will I until Wednesday at the earliest.  That means merging this as is if we want this to be part of beta2.


henryso:
Without actually being able to test this, I cannot speak to it working, but reviewing it by eye, it looks OK.

eroux:
Same here, thanks a lot!

PR #892, opened 2016-02-09
first attempt to fix #891

### LINKED ISSUE #891: --version, suggest --help instead of dumping entire usage
it would be nice if gregorio supported the widespread  --version option along with --help.  Here is some info for it:
https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html

Also, it would be nice if gregorio, when invoked with no arguments, just said something like:
  gregorio: missing file operand
  Try 'gregorio --help' for more information.
instead of dumping the whole usage message.

Similarly, when there is an error in arg parsing, an invocation like "gregorio -o" with no argument, it's nicer to just report the error and give the suggestion instead of dumping the whole usage message.  That is usually too much info.

Also, the spaces before colons ("Usage :", "is :"), while I know are usual in French, are just weirdly wrong in English.  There's also no need for the newline before "Usage", or the newline after the "Usage:", or an extra newline at the end.  Fewer lines is better.

Also, it is highly useful to write a sentence or two about what the program does in the --help message.  Something like "Engrave Gregorian chant scores: convert a gabc file to GregorioTeX." 

Also, it's typical to capitalize metavariables like "FORMAT" and "FILE"  in the help message.

Also, the help message could/should end with a url to the overall project home page.

Also, it is not usual to capitalize "Chant" in "Gregorian chant", as far as I've seen.

See pretty much any GNU --help (and --version) message for examples of all this in action.  "mv --help" is reasonably short and instructive.

Happy helping.

eroux:
All tests pass. Please review my broken English

### DIFF
--- a/src/gregorio-utils.c
+++ b/src/gregorio-utils.c
@@ -184,7 +184,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.\n\
 static void print_usage(char *name)
 {
     printf(_("Usage: %s [OPTION]... [-s | INPUT_FILE]\n\
-\nEngrave Gregorian chant scores, convert a gabc file in GregorioTeX.\n\n\
+\nEngrave Gregorian chant scores, convert a gabc file to GregorioTeX.\n\n\
 Options:\n\
   -o, --output-file FILE    write output to FILE,\n\
                             default is basename(INPUT_FILE).FORMAT\n\
--- a/src/gregorio-utils.c
+++ b/src/gregorio-utils.c
@@ -5,7 +5,7 @@
  * Copyright (C) 2006-2015 The Gregorio Project (see CONTRIBUTORS.md)
  *
  * This file is part of Gregorio.
- * 
+ *
  * Gregorio is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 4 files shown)


eroux:
I've pushed a few changes

henryso:
I'm done with my review.

src/gregorio-utils.c
```diff
@@ -161,54 +161,62 @@ static char *get_output_filename(char *fbasename, const char *extension)
...
-\t-f format  specifies input file format (default: gabc)\n\
-\t-s         reads input from stdin\n\
-\t-p         generate point-and-click information\n\
-\t-h         displays this message\n\
-\t-V         displays %s version\n"), name, name);
-    printf(_("\t-L         displays licence\n\
-\t-v         verbose mode\n\
-\t-W         displays all warnings\n\
+    printf(_("Usage: %s [OPTION]... [-s | INPUT_FILE]\n\
+\nEngrave Gregorian chant scores, convert a gabc file in GregorioTeX.\n\n\
```
henryso: Ok, I just noticed this...  "in" should be "to".

src/gregorio-utils.c
```diff
@@ -161,52 +161,62 @@ static char *get_output_filename(char *fbasename, const char *extension)
...
+  dump      plain text dump (for debugging purpose)\n\
 \n\
-available formats are:\n\
-\t gabc      gabc\n\
-\t gtex      GregorioTeX\n\
-\t dump      simple text dump (for debugging purpose)\n\
-\n"));
+See <%s> for general documentation,\n\
+GregorioRef-%s.pdf and GregorioNabcRef-%s.pdf for full documentation.\
+\n"), PACKAGE_URL, VERSION, VERSION);
```
henryso: Since these are `#define`, they can be inserted directly: `"...GregorioRef-" VERSION ".pdf..."`

eroux:
more comments addressed, please review

henryso:
Looks good to me.

rpspringuel:
Do we have tests for this output?  If not, should we?

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

eroux:
All tests pass, I don't think it's necessary to have many tests for that... and even if some could be done, I think we should focus more on the features and bug requests for 4.1 (I have a big bug under Windows, but I didn't understand it enough yet).

rpspringuel:
Shouldn't the short usage info message print the command syntax?  I.e. What is the first line of the help message. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

eroux:
good idea, done

henryso:
We have tests for all the options, but they are pass/fail (return code) only and don't check the message content.

eroux:
I've transfered a few messages to stderr, and added a message for non-blocking case, please review, I think I'm done with changes (I thought I was for a few hours actually)...

henryso:
I think this is OK to merge.

PR #896, opened 2016-02-10
Feature/miktex

rpspringuel:
In addition to the immediate problem of the missing file in the Windows installation, I've also fixed a whole bunch of little things which I came across in the process.  As a result I'd like some one to double check what I've done before merging it into develop.

I will, however, tag and release beta3 from this branch so that those users who are testing on Windows can get something working sooner.


VersionManager.py
```diff
@@ -51,10 +51,14 @@
...
                   "tex/gregoriotex-chars.tex",
                   "tex/gregoriotex-signs.tex",
+                  "tex/gregoriotex-signs.lua",
                   "tex/gregoriotex-spaces.tex",
                   "tex/gregoriotex-symbols.tex",
+                  "tex/gregoriotex-symbols.lua",
                   "tex/gregoriotex-syllable.tex",
                   "tex/gregoriotex-main.tex",
+                  "tex/gregoriotex-nabc.tex",
+                  "tex/gregoriotex-nabc.lua",
```
henryso: Of all the things for me to use as a search term to find where `gregoriotex-symbols.lua` was missing, I chose `gregoriotex-nabc.lua`, which of course missed this file.  Thanks for finding it.

windows/uninstall.lua
```diff
@@ -55,13 +55,10 @@ local texmflocal = fixpath(kpse.expand_var("$TEXMFLOCAL"))..pathsep
 local texmfdist = fixpath(kpse.expand_var("$TEXMFDIST"))..pathsep
 
 function remove_executable()
-  if not lfs.isdir(texmflocal) then
-    return false
-  end
   print("Removinging gregorio.exe...")
```
henryso: Minor, but this is a typo.

henryso:
It looks OK to me (other than that minor typo), but I didn't run through the logic of the install script; I'm assuming your testing will cover that.

windows/gregorio.iss
```diff
@@ -54,59 +55,36 @@ Source: "../README.md"; DestDir: "{app}";
...
-Source: "../fonts/gresgmodern.ttf"; DestDir: "{app}\texmf\fonts\truetype\public\gregoriotex";
+Source: "../fonts/*.ttf"; DestDir: "{app}\texmf\fonts\truetype\public\gregoriotex";
 Source: "../fonts/convertsfdtottf.py"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
-Source: "../fonts/greciliae-base.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
-Source: "../fonts/greextra.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
-Source: "../fonts/gregall.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
-Source: "../fonts/gresgmodern.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
-Source: "../fonts/gregorio-base.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
+Source: "../fonts/*.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
 Source: "../fonts/Makefile"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
```
eroux: I think this file used to be relevant when it was not generated by autotools, now I don't think it's really usable outside of the build system, so I think it should be removed

windows/gregorio.iss
```diff
@@ -54,59 +55,36 @@ Source: "../README.md"; DestDir: "{app}";
...
-Source: "../fonts/parmesan-base.sfd"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
 Source: "../fonts/squarize.py"; DestDir: "{app}\texmf\fonts\source\gregoriotex";
```
eroux: I think the new `stemsschemas.py` should appear here too (and in the tarball too!), sorry

windows/gregorio.iss
```diff
@@ -54,59 +55,36 @@ Source: "../README.md"; DestDir: "{app}";
...
 Source: "../README.md"; DestDir: "{app}\texmf\doc\luatex\gregoriotex";
 
+[InstallDelete]
+Type: files; Name: "{app}\gregorio.exe"
```
eroux: I think you forgot `\bin` here?

windows/uninstall.lua
```diff
@@ -150,7 +147,11 @@ function remove_texmf_install()
     end
   end
   if install_was_present then
-    os.spawn("updmap")
+    --[[Since we removed some fonts, we need to rebuild the font databases.
+    Since this function is only used when extracting files from a TeXLive
+    texmf tree, we don't need to go through the distribution check.
+    ]]--
+    os.spawn("updmap-sys")
```
eroux: `updmap` does nothing with gregorio fonts, it was just here to handle very old versions of Gregorio using `ovf`, `ofm`, etc. fonts, which were handled by `updmap`. For the new database, `luaotfload-tool` is what you're looking for

windows/modpath.iss
```diff
@@ -0,0 +1,219 @@
...
+				for i := 0 to GetArrayLength(aSelectedTasks)-1 do begin
+					if comparetext(aSelectedTasks[i], taskname) = 0 then
+						ModPath();
+				end;
+			end;
+		end;
+	end;
+end;
+
+function NeedRestart(): Boolean;
```
eroux: Wow, this means that you need to restart Windows when changing PATH?? Are we still in the 70s?

eroux:
Thanks a lot for your work on this, I think it goes in the right direction! You also did the right thing releasing beta3, that will be extremely helpful for many users! Apart from the small comments I made, I'm wondering how things work when changing the PATH in this new installer? Is the gregorio path removed when uninstalled?

windows/modpath.iss
henryso: It looks like that's true for people in the 90s, as in Windows 95, or really anyone who's not `UsingWinNT()`.

eroux: Ok thanks :)

windows/gregorio.iss
rpspringuel: No.  I didn't want to add all of `{app}` to `PATH` so I moved the location of the executable to `{app}\bin`.  This way the folder which I add to `PATH` only contains the exectuable, and not all the other stuff that we install.  The point of this line is to remove old versions of the executable which were previously placed in `{app}`.  New versions, placed in `{app}\bin` will overwrite each other on upgrade.

eroux: Ok, understood, thanks!

CHANGELOG.md
```diff
@@ -17,7 +28,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
 - The `--admin` option is removed from the `initexmf` call for MiKTeX installations, allowing the installer to work better on Windows 10.
 - MiKTeX installations no longer copy files into TEXMFLOCAL but instead registers the texmf folder which the installer creates with MiKTeX, allowing it to use those files in place.  See [#884](https://github.com/gregorio-project/gregorio/issues/884)
-- Added an uninstall script to the windows installer which will run as part of the uninstall action and will remove the Gregorio executable and TeX files from their copied locations (at least using the `unins000.exe` file which is created in the program directory when the program is installed, I haven't tested using the Add/Remove Programs feature built into Windows).
+- Windows installer no contains an uninstall script which will run as part of the uninstall action and thus should remove the Gregorio executable and TeX files from their copied locations (leaving a clean post-uninstall system).
```
eroux: I think it should be `now` instead of `no`

rpspringuel:
It is my understanding of the modpath script that I used that it will undo the change to `PATH` when uninstalling Gregorio.  However, I have to admit that when I did my testing last night, I didn't actually check on that.  I'll pencil in some more time to do additional testing in the near future.

rpspringuel:
I've made the changes (and now uploaded them to the right repository for them to appear in this PR).  I won't be able to get to the testing until Friday at the earliest.  Feel free to decide whether you want to wait for that to merge or not.

eroux:
Apart from the small comment it's ok for me, maybe the best is that you merge after your tests?

rpspringuel:
I've fully tested this on Windows 10 and the fact that beta3 is on this branch is causing merge conflicts with ongoing bug fixes, so I'm going to merge this now.

Issue #897, opened 2016-02-10
Liquescent scandicus determination is broken


henryso:
Attempting to create a liquescent scandicus in gregorio 4.1-beta2 will fail with an assertion error.  This is due to an error in the assertion.

henryso:
In adding the assertions, I asserted that liquescent scandicus would be determined properly in glyph determination; it is not and (in the prior code) resulted in a fall back to three individual notes glued together.  There are three ways I can see to fix this:
- remove the assertion and restore this figure falling back to being three notes glued together
- determine it as something else (punctum + podatus, podatus + punctum?)
- draw the missing glyphs and determine it as liquescent scandicus.

Any thoughts or suggestions?

eroux:
What comes to my mind first is the first solution... let's wait for a user request before implementing the third one. About two, I'm pretty sure I've seen figures like `(ghi<)` somewhere, so let's not do it...

PR #901, opened 2016-02-10
Horizontal episema tuning

### LINKED ISSUE #872: cheating horizontal episemus height
In the following example:

episemi

it would be important to be able to make the second set of horizontal episemus (on *ter*) be on the exact same height as the one before (*gis*). I think detecting this automatically is just too complex, so I propose a notation to be able to "cheat" horizontal episemus placing. For example here `ter(ixg[hh:f]hi)` would make the horizontal episemus placing algorithm think the first note is an `f` and thus place the whole episemus on the correct position... What do you think?

henryso:
This also restores scandicus determination.
Fixes #872 and #897.

A number of tests change microscopically, and I added tests for these two issues.


eroux:
That really looks perfect thank you so much! That's quite an impressive feature, and though it won't be used in each and every score, it will really make a difference for many users and make Gregorio able to handle more use cases! Ok for me to be merged

Issue #903, opened 2016-02-11
gregoriotex-symbols.tex uses \gre@debugmsg when it might not be defined


henryso:
See https://github.com/gregorio-project/gregorio/pull/896#commitcomment-16022566

If a LaTeX file uses the `gregoriosyms` package alone, `\gre@debugmsg` will not be defined when `gregoriotex-symbols.tex` is loaded.

Issue #880, opened 2016-02-06
small error on horizontal episemus on first note of -g_ig'ig~


eroux:
Reviewing some tests, there seems to be an error on the width of the horizontal episemus on the first note of `(-g_ig'ig~)`:

![smallbug](https://cloud.githubusercontent.com/assets/60868/12866570/8a6ea3f0-ccd2-11e5-80d2-4970dad32a4a.png)

This is all but an emergency, but I thought it would be wort reporting it

Issue #863, opened 2016-02-02
Heuristics should be glyph-wise


eroux:
See [this comment](https://github.com/gregorio-project/gregorio/issues/803#issuecomment-178758131) and following, if `[ll:0]` or `[ll:1]` are found on any note of the glyph, they should count for the whole glyph, as I cannot imagine a documentation describing on which note to add the `[ll:x]` notation according to the glyph...

henryso:
Does this mean all notes in the glyph should be marked as having a ledger line globally (i.e., it affects sign positioning in TeX), or should this change only affect stem determination?

eroux:
That's a good question... An alternative would be that stem length algorithm looks at `[ll:x]` on the note carrying the stem (this ways the documentation can remain simple)... does it sound feasible?

henryso:
All the listed solutions are feasible without too much effort.  It all rides on which solution is most sensible.  To that end, I think your latest suggestion sounds like the best of these.

eroux:
ok then, let's do it that way

PR #904, opened 2016-02-11
Fixes for 862, 863, 880, and 903.

### LINKED ISSUE #862: Heuristics should include notes that are "glued" to the glyph
From https://github.com/gregorio-project/gregorio/issues/803#issuecomment-178697013:

The heuristics to know if a ledger line is present or not under the glyph should also check the notes glued to the current glyph, for instance `b!dv` should have a long form by default. I think notes separated by a `/0` space should be counted too

henryso:
Fixes #862, #863, #880, and #903.

This also fixes a bug found in the #901.

I updated tests as necessary.

Please review and merge if satisfactory.  If you'd rather I split this into multiple pull requests, I can do that as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - New bar spacing algorithm now accounts for translations (and other below lyrics stuff).
 - Horizontal episema on initio debilis is now sized correctly (see [#880](https://github.com/gregorio-project/gregorio/issues/880)).
 
+### Changed
+- Adjustments to the heuristic for ledger lines to include adjacent notes (see [#862](https://github.com/gregorio-project/gregorio/issues/862)).
+- The stem length determination will use the ledger line below the note with the stem (see [#863](https://github.com/gregorio-project/gregorio/issues/863)).
+
 ### Added
 - Controls for tuning horizontal episema vertical position.  See GregorioRef for details (for the change request, see [#872](https://github.com/gregorio-project/gregorio/issues/872)).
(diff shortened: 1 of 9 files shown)


rpspringuel:
It may be that #862 and #863 are refinements of new features in 4.1 and thus their CHANGELOG entries will eventually be consolidated away, but since these affect the visual output, I'm inclined to mention them in the CHANGELOG, at least during the release process.

rpspringuel:
I'm getting a test failure for this.

gabc-gabc/glyphs/horizontal_episema_tuning.gabc has two extra `[lh:m{]` in the output.  It appears that they are being repeated at the start of the subsequent syllable when the defined range spans multiple syllables.

henryso:
I thought that we only add CHANGELOG entries for that kind of thing after rc1 and not through the betas, but I'll add it since you feel that strongly about it.  I'll also the problem you noticed with the gabc-gabc test.

rpspringuel:
Another test failure.

It seems the new results from the bar-spacing-old test (i.e. the ones with the translation examples) didn't get merged in properly.  Can you simply accept the results for this test to fix that here?

henryso:
I'm not getting these test failures.  Maybe there's something wrong with my test directory.

henryso:
Regarding gabc-gabc/glyphs/horizontal_episema_tuning.gabc, are you saying it doesn't match the expectation or that the expectation is, in your opinion, wrong?

I noticed that gabc output was using the old letters (`lh` instead of `uh`), so I've fixed that, but otherwise, the expectation looks correct to me.

henryso:
I think I found the problem that you are seeing in the gabc-gabc test.  Valgrind found uninitialized variable usage, so that might be causing the problem.  I will have a fix up shortly.

henryso:
I believe I've addressed all the issues here, and I've rerun the coverage test and the valgrind/sanitize tests and corrected the issues they found.

Please review the result.

eroux:
Wonderful! Thank you very much for this, that will make the issue tracker lighter! The only thing left I see is documenting it in Gabc.tex, but let's keep it for another PR

PR #906, opened 2016-02-11
Fix #890, document #863 and #716

### LINKED ISSUE #890: Generated files left behind after `make maintainer-clean` in doc
This is far from being critical, but in building 4.1.0-beta2 I noticed that the secondary aux files (those for the included files), generated gtex files, and the gaux files (all created by latexmk when the docs are build) are left behind after `make maintainer-clean` is run on the repository.

eroux:
Fix #890, document #863 and #716

### DIFF
--- a/doc/Gabc.tex
+++ b/doc/Gabc.tex
@@ -220,9 +220,8 @@ line (\textit{d}). Most of the time they will take a short form:
 But when a ledger line is drawn below these notes, they should take a long
 form. The problem is that many cases are ambiguous: for instance if a note
 is close to a ledger line, one may want to make it long, others may not.
-To solve this problem, you can add \texttt{[ll:0]} to any note of a glyph to
-force the stem of the glyph to get its short form, or \texttt{[ll:1]} to force
-its long form.
+To solve this problem, you can add \texttt{[ll:0]} to the note carrying the stem
+to get its short form, or \texttt{[ll:1]} to force its long form.
 
 % This snippet makes LuaTeX segfault!
 %\gabcsnippet{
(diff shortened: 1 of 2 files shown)


doc/Gabc.tex
```diff
@@ -283,6 +282,15 @@ \subsection{Simple Slurs}
 Use \texttt{uslur} instead of \texttt{oslur} (with either form) to create an
 under-the-staff slur.
 
+\subsection{Horizontal episema placement for very high and low notes}
+
+Gregorio places horizontal episemus under c and above k closer to the notes when no ledger line is present.
```
henryso: As discussed before, episema is Latin 3rd declension, with citation form episema, episematis, neuter gender.  So the singular nominative or accusative form would be episema and the plural form is episemata.  So here, "episemus" should be replaced with episemata.

Also, `k` is not always the pitch below the ledger line; it depends on the number of staff lines.

doc/Gabc.tex
```diff
@@ -283,6 +282,15 @@ \subsection{Simple Slurs}
...
+The heuristics used by Gregorio are not perfect so it may be necessary to make
+the presence or absence of ledger line explicit for horizontal episemus placement.
```
henryso: Here, "episemus" should probably be "episema".

doc/Gabc.tex
```diff
@@ -283,6 +282,15 @@ \subsection{Simple Slurs}
...
+This is done in the exact same way as for stem length: place \texttt{[ll:0]} or
+\texttt{[ll:1]} on the note carrying the episemus, to force gregorio to consider the
```
henryso: "episemus" should be "episema".

doc/Gabc.tex
```diff
@@ -283,6 +282,15 @@ \subsection{Simple Slurs}
...
+absence or presence of a ledger line in episemus placement.
```
henryso: "episemus" should be "episema"

doc/Gabc.tex
eroux: Right, I'm having hard time getting used to English taking the native plural... French foreign words (Latin and Greek are considered foreign) take French plural with s (so episemus, puctums, virgas, etc.), fixing

eroux: fixed

henryso:
Looks good to me.

henryso:
By the way, I would consider Latin foreign as well.  Pluralization in English is typically by adding `s` as well, though there are probably more irregular cases than other languages because of a large influx of borrowed words.  So every time I see something like "episema", I have to look it up (and episema, being 3rd declension is particularly weird).

eroux:
Oh, I thought it was a general rule (like *phenomenon* -> *phenomena*)... In France some people find it smart-looking to use Latin plural (*un forum Internet* -> *des fora Internet*) though it's not the proper rule, I tend to find it a bit weird... i think if declination should be taken, then it should be taken properly: *the height of the episemorum* or *the height episemorum*! But each language has its own particularities, let's use *episema* in English (and *episemus* in French!)

henryso:
As far as I can tell, (episema, episematis, neuter) would pluralize episema as episemata, which is what I've been using for the plural form.

PR #909, opened 2016-02-11
Fixed centering when the determined center ends at an elision.

### LINKED ISSUE #907: elision centering is wrong
gabc `cla<e>ra</e>(gh)` gives

incorrect

while it should give

correct

the centering algorithm should be the normal one, except that letters inside `<e>` should never be considered as vowels.

henryso:
Fixes #907.

All current tests pass, but that means that this bug was not included in the tests, so I added the appropriate test.


eroux:
Looks perfect, thanks a lot!

PR #911, opened 2016-02-12
Updated vim syntax file for new gabc features.

henryso:
Since I use vim to edit pretty much everything, I decided to update the syntax file for most (maybe all) the new gabc features.

Please review and feel free to either merge it if you like it or discard it if you don't.

### DIFF
--- a/contrib/gabc.vim
+++ b/contrib/gabc.vim
@@ -1,6 +1,5 @@
 " Vim syntax file
 " Language:	Gabc gregorian chant notation
-" Maintainer:	Elie Roux <elie.roux@telecom-bretagne.eu>
 " Last Change:	2016 Feb 11
 
 " Quit when a (custom) syntax file was already loaded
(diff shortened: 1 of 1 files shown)


contrib/gabc.vim
```diff
@@ -1,7 +1,7 @@
 " Vim syntax file
 " Language:	Gabc gregorian chant notation
 " Maintainer:	Elie Roux <elie.roux@telecom-bretagne.eu>
-" Last Change:	2008 Nov 29
+" Last Change:	2016 Feb 11
```
eroux: Looks good, I trust you've tested it better than I would, the only remark is the "Maintainer" Line that should be changed or removed

henryso: OK, I removed the "Maintainer" line.  I have tested this against most of the gabc files in the test repository.  I'm sure there are missing edge cases because of the looseness of the "syntax" of gabc, but I think it covers 90+% of cases.

eroux:
Thanks!

PR #912, opened 2016-02-12
simplify fonts

### LINKED ISSUE #874: simplifying font by hand?
The fontforge script doesn't simplify glyphs even when `.simplify()` is called on each glyph or on a selection including all glyphs. Simplifying by hand gives these results:

- no simplifcation:
  - `greciliae.ttf`: 1047712 bytes
  - a small pdf including two scores : 21848 bytes
- simplified:
  - `greciliae.ttf`: 920228 bytes
  - a small pdf including two scores : 21512 bytes

which is 12% for the font and 1.5% for the pdf. Although the economy is quite small for the pdf, maybe it's still worth doing a simplification by hand for the different fonts we generate? It's fairly simple:

- open into fontforge
- select all glyphs (ctrl+a)
- Element->Simplify->Simplify
- File->Generate Font

What do you think about doing this before releasing the tarballs? Maybe not for betas and rcs (though it would be good to do it once before the final release to see if there is any problem)...

eroux:
fixes #874

### DIFF
--- a/fonts/simplify.py
+++ b/fonts/simplify.py
@@ -66,7 +66,7 @@ def main():
     fontfile = args[0]
     font = fontforge.open(fontfile)
     font.selection.all()
-    font.simplify()
+    font.simplify(0.1, ('mergelines','ignoreslopes','setstarttoextremum'))
     font.generate(fontfile)
     font.close()
 
--- a/fonts/simplify.py
+++ b/fonts/simplify.py
@@ -4,10 +4,11 @@
 
 
 """
-    Python fontforge script to convert from fontforge's native sfd
-    to a TrueType font (ttf).
+    Python fontforge script to simplify a font. This only exists
+    because simplification doesn't work in squarize.py due to bugs
+    in fontforge.
 
-    Copyright (C) 2015 The Gregorio Project (see CONTRIBUTORS.md)
+    Copyright (C) 2016 The Gregorio Project (see CONTRIBUTORS.md)
 
     This file is part of Gregorio.
 
@@ -24,10 +25,10 @@
     You should have received a copy of the GNU General Public License
     along with Gregorio.  If not, see <http://www.gnu.org/licenses/>.
 
-    This script takes a .sfd file and converts it to a .ttf file.
+    This script takes a font file, simplifies it and overwrites it.
 
     Basic use :
-        fontforge -script convertsfdtottf.py fontfile
+         fontforge -script simplify.py fontfile
 """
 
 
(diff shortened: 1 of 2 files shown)


eroux:
With these simplifications, the output almost doesn't change: I think the changes are mostly due to anti-aliasing, but please double-check, but it doesn't reduce size as much as I hoped (`104k` -> `97k` instead of `92k`). I tried more aggressive simplification but the output was changed, sometime visibly (for pes deminutus mainly). Tell me what you think about the right balance, maybe more aggressive simplification is ok.

henryso:
Are you sure about the numbers?  Greciliae is currently 1.1M.  Assuming you're off by an order of magnitude, a 40k reduction might be worth it.  Can you post some side-by-sides of the figures that change significantly with more aggressive simplification?

eroux:
Sorry, I got it wrong by a factor 10: `1048040` -> `976544`. The most changing figure is pes deminutus:

diff-simpl

henryso:
Can you post a side-by-side of the glyph?  I can see the difference in the contour you posted, but at a "normal" resolution, it might not matter all that much.

eroux:
not simplified:

old

simplified:

new

eroux:
Well... taking another look at it, I think both are fine (well... to say it another way: I cannot tell the difference) and it's probably a good idea to completely simplify the font?

henryso:
It is very subtle, but I honestly like the non-simplified slope of the line better.  However, if I didn't have a side-by-side, I probably wouldn't notice.

fonts/simplify.py
```diff
@@ -0,0 +1,74 @@
...
+            sys.exit()
+    if len(args) == 0:
+        usage()
+        sys.exit(2)
+    fontfile = args[0]
+    font = fontforge.open(fontfile)
+    font.selection.all()
+    font.simplify(0.1, ('mergelines','ignoreslopes','setstarttoextremum'))
+    font.generate(fontfile)
+    font.close()
```
henryso: This technique doesn't work in `squarize.py`?

eroux: No, it produces a `1001528 B` font (I thought it did nothing but it seems I was wrong... it does less than this script though)

eroux:
Well, if you see a difference side by side, let's not simplify further then... Gregorio is more about quality output than small pdf or small download size...

eroux:
(let's wait for other comments)

rpspringuel:
I have the same reaction as @henryso when seeing them side-by-side.

eroux:
ok, I think this is ready to be merged then

henryso:
No objections.  Merging.

PR #915, opened 2016-02-13
small header doc improvements

eroux:
This mainly adds a common example so that people can start experimenting in a smoother way. It also tries to translate some sentences into so-called "International English". Feel free to remove all the changes, but I think the example is quite important (it was for me, I had to look for some examples in `gregorio-tets` to understand why my attempts didn't work).

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -1064,7 +1064,7 @@ As an example, you can use
 thus automatically printing the content of the header above the score.
 
 \macroname{\textbackslash grebeforeheaders}{\{\#1\}}{gregoriotex-main.tex}
-Specifies a macro called before the processing of the headers of a score.
+Specifies \TeX{} code processed before the processing of the headers of a score.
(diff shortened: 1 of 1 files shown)


doc/Command_Index_User.tex
```diff
@@ -1055,12 +1055,16 @@ \subsubsection{Headers}
...
+
+\verb=\gresetheadercapture{commentary}{grecommentary}{string}=
+
+\noindent to capture the
+\texttt{commentary} header of gabc files and feed it to \verb=\grecommentary=,
+thus automatically printing the content of the header above the score.
 
 \macroname{\textbackslash grebeforeheaders}{\{\#1\}}{gregoriotex-main.tex}
-Specifies something to be done before a set of headers is processed.
+Specifies a macro called before the processing of the headers of a score.
```
henryso: This can be arbitrary TeX code, not just a macro.

henryso:
Besides that subtle point, I'm fine with these changes.

eroux:
Thanks, fixed

Issue #898, opened 2016-02-10
Completing glyph handling


eroux:
Looking at Nocturnale Romanum:

![nr](https://cloud.githubusercontent.com/assets/60868/12943280/8ac8e7fa-cfe0-11e5-92f9-52ad9484d1c2.png)

there are a few glyphs that Gregorio doesn't handle and that seem to be used in Antiphonale Monasticum project. The main one is "virga strata", but it would be good to handle (in the form where the two pitches are not the same, but keeping both notes in the same element when they have the same pitch), and "pes stratus" too. This also raises the question of how to apply the changes made in #744 for combined glyphs. For instance, the pes stratus would have different forms according to the following note, how should the glyphs be named?

This could also be the occasion to apply #744 completely and revert the direction of the oriscus for the `FlexusOriscusScapus` glyph, as these are just plainly wrong in the Nocturnale Romanum, and we copied them. So should we add some `.nr` variants for the current ones, and fix the normal form?

Note also that NR has a "salicus quassus" figure which looks just plainly wrong, I don't think it should be added to Gregorio...

henryso:
The following are currently possible:

![image](https://cloud.githubusercontent.com/assets/421318/12947085/15d5d264-cfc7-11e5-9177-04c1b726c3f3.png)

eroux:
Thanks! this leaves the following remarks:

- the only problem for virga strata seems to be when the ambitus is one... very strange... maybe we should open a bug issue against 4.1 for this?
- the pes stratus is perfect this way, I didn't find any mean to make the second oriscus in the opposite direction, is there any?
- there should be a way to turn the oriscus of `gO` in the other direction (the glyphs are already drawn IIRC)
- inverting the oriscus of `gOe` (with variants)

henryso:
- What is the problem with virga strata at ambitus one?
- There is currently no way to change the direction of the second oriscus; the current rules are that oriscus can only fuse upwards `(f@ho@j)` or downwards `(i@ho@f)`.  The "exception" here is that a "leading" stemmed oriscus (one that starts a figure) can fuse in either direction.  This, of course, can be changed.
- Inversion of the oriscus in those two figures sounds doable.

eroux:
Forget about the bug, I was trying `gho` instead of `ghO`! Let's not hurry about the fusion change, I don't think it's really a priority, but allowing an inversion of `gO` and `gOe` would be a good feature for 5.0!

henryso:
[...]

PR #917, opened 2016-02-13
Implemented oriscus direction for both normal oriscus and oriscus scapus

henryso:
Along with this, forced direction (with the new rule set) now uses '0' and '1'.
This implements the version 4.1 parts of #898.

I realize there are a lot of changes; this was due to @eroux's request to implement directionality for the oriscus scapus.  I think everything is working properly and have run the coverage and memory testing on this.


eroux:
Sorry, I didn't realize it would be so much work! I thought it would be a line or two...  So thanks a lot! I'll review later tonight

Issue #914, opened 2016-02-13
fusion not working for some forms


eroux:
Fusion doesn't work for some forms, which seems to be when fusing a form like `gi~` from above, for instance

`(-g@i@gi~) (gV@eg~)`

gives

![bug-fusion](https://cloud.githubusercontent.com/assets/60868/13027706/0e160522-d25a-11e5-9e67-e6c041ff7c0a.png)

not fusing the two parts. This used to work in beta1.

PR #918, opened 2016-02-13
Fusion fixes.

### LINKED ISSUE #913: queue length for fused virga seems wrong
Here is the result of `(hV@g@h@gf~)` (which the Antiphonale Monasticum project will actually use):

queuelength

The queue at the beginning has the wrong length... If the problem lies in the font I can fix it, but I admit I'm quite uncertain about the queue length behavior in fusion (especially since I don't have a very precise idea on how fusion works in the C code)...

Also, I don't think there's currently any way to get the same result with the right queue length?

henryso:
Fixes #913 and #914.

Note that this is based on (my) branch fix-898-4.1 (#917), so merging it will merge that one as well.  This also means it is more difficult to review, so I recommend reviewing the commits (a0ccba1 and 23aecae) themselves.  I did this because the test changes compound on the other branch.  If #917 is rejected, I will rebase this against develop.


eroux:
both commits are ok, I think it's safer to merge after #917?

PR #920, opened 2016-02-14
fix #919

### LINKED ISSUE #919: new bar spacings wrong when mora preceeds a bar
Gabc `3(g.) (;) (g) 4(g) (;) (g)` gives the output

bugbars

eroux:
fixes #919 and part of #921

### DIFF
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -925,9 +925,9 @@
       %this calculation is similar to what is done in \gre@calculate@syllablefinalskip, but is only for the text, so we'll use \gre@skip@syllablefinalskip to hold this value for now
       \ifnum#4=1\relax %
         \ifgre@in@euouae %
-          \gre@skip@syllablefinalskip=\gre@space@skip@interwordspacetext@euouae\relax%
+          \gre@skip@syllablefinalskip=\gre@space@skip@interwordspacetext@bars@euouae\relax%
         \else %
-          \gre@skip@syllablefinalskip=\gre@space@skip@interwordspacetext\relax%
+          \gre@skip@syllablefinalskip=\gre@space@skip@interwordspacetext@bars\relax%
         \fi %
       \else %
         \gre@skip@syllablefinalskip=0pt\relax%
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -822,7 +822,7 @@
 }%
 
 \newif\ifgre@newbarspacing%
-\gre@newbarspacingtrue%
+\gre@newbarspacingfalse%
 
 \def\gresetbarspacing#1{%
   \IfStrEq{#1}{new}%
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -957,11 +957,12 @@
       \advance\gre@skip@temp@two by \glueexpr(\gre@skip@temp@three - \gre@skip@temp@one)\relax%
       \gre@skip@temp@one = \gre@skip@temp@three\relax%
     \fi%
-    \gre@debugmsg{barspacing}{text space after accounting for text: \the\gre@skip@temp@one}%
-    \gre@debugmsg{barspacing}{notes space after accounting for text: \the\gre@skip@temp@two}%
+    \gre@debugmsg{barspacing}{space required by text: \the\gre@skip@temp@three}%
+    \gre@debugmsg{barspacing}{allocated text space: \the\gre@skip@temp@one}%
+    \gre@debugmsg{barspacing}{allocated notes space: \the\gre@skip@temp@two}%
     % now we check the bar width
     %as with the text, we start by borrowing the minNotesDistance left over from the previous syllable
-    \gre@skip@temp@three = \glueexpr\wd\gre@box@syllablenotes + \gre@skip@minNotesDistance\relax%
+    \gre@skip@temp@four = \glueexpr\wd\gre@box@syllablenotes + \gre@skip@minNotesDistance\relax%
     %now, if there is a visible bar for this syllable, we calculate this syllable’s equivalent of minNotesDistance
     % gre@count@temp@one holds 0 if next is not a bar, 1 if it is, 2 if it starts with an alteration
     \gre@count@temp@one=0%
@@ -994,14 +995,15 @@
         \fi %
       \fi %
       \gre@debugmsg{syllablespacing}{ space after bar = \the\gre@skip@syllablefinalskip}%
-      \advance\gre@skip@temp@three by \gre@skip@syllablefinalskip%
+      \advance\gre@skip@temp@four by \gre@skip@syllablefinalskip%
     \fi%
-    \ifdim\gre@skip@temp@two < \gre@skip@temp@three\relax%
-      \gre@skip@temp@one = \glueexpr\gre@skip@temp@one + (\gre@skip@temp@three - \gre@skip@temp@two)\relax%
-      \gre@skip@temp@two = \gre@skip@temp@three\relax%
+    \ifdim\gre@skip@temp@two < \gre@skip@temp@four\relax%
+      \gre@skip@temp@one = \glueexpr\gre@skip@temp@one + (\gre@skip@temp@four - \gre@skip@temp@two)\relax%
+      \gre@skip@temp@two = \gre@skip@temp@four\relax%
     \fi%
-    \gre@debugmsg{barspacing}{required space for text: \the\gre@skip@temp@one}%
-    \gre@debugmsg{barspacing}{required space for notes: \the\gre@skip@temp@two}%
+    \gre@debugmsg{barspacing}{required space for notes: \the\gre@skip@temp@four}%
+    \gre@debugmsg{barspacing}{allocated text space: \the\gre@skip@temp@one}%
+    \gre@debugmsg{barspacing}{allocated notes space: \the\gre@skip@temp@two}%
     % now we divide the two distances in half to represent the position of the middle of each element
     \gre@debugmsg{barspacing}{calculate position of middles}%
     \divide\gre@skip@temp@one by 2\relax%
@@ -1009,7 +1011,6 @@
     \gre@debugmsg{barspacing}{distance from previous text to middle of this text: \the\gre@skip@temp@one}%
     \gre@debugmsg{barspacing}{distance from previous notes to middle of this notes: \the\gre@skip@temp@two}%
     % our final position calculation is to make sure that the centers of the text and the bar are not too far from each other
-    \gre@debugmsg{barspacing}{limit by offset}%
     % this isn't really the begindifference (it doesn't account for the width of the text and the bar), but it's close enough that we'll use that register for this calculation
     % this value is positive if the middle of the bar is before the middle of the text
     \gre@dimen@begindifference=\dimexpr(-\gre@dimen@temp@three - \gre@skip@temp@two + \gre@skip@temp@one)\relax%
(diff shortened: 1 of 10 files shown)


eroux:
Thinking back about it, this patch will certainly introduce bad output in the new spacing algorithm, with bar text not being correctly centered.... @rpspringuel do you think you could take a look (you can push on the branch of this PR, it's on the main repo)? The problem should be easier to solve with the changes of this branch, as it shows where the problem comes from

eroux:
Also, I think it's best to wait for this to be properly fixed before releasing rc1, as it involves many changes for users.

eroux:
The current state should be better but a few bugs remain... I'm still on it

eroux:
I'm quite happy with the current state, except for #921. I think it will need careful attention, as it is will modify the code quite a lot, adding many edge cases. I'll stop here for today, @rpspringuel can you please review and fix #921? This will certainly take some time, but that's a higher priority than the release of rc1 today.

rpspringuel:
I'll take a look.

It doesn't look like you've added or modified any tests to cover this bug (just accepted new results).  Does this mean that the existing tests covered it?

tex/gregoriotex-syllable.tex
```diff
@@ -818,7 +822,7 @@
 }%
 
 \newif\ifgre@newbarspacing%
-\gre@newbarspacingfalse%
+\gre@newbarspacingtrue%
```
rpspringuel: For future reference:

The fact that you changed the default and implemented the fix at the same time threw me for a bit, especially since all the test changes were accepted simultaneously in gregorio-test.  I had deliberately been holding off on changing the default on my feature/bar_spacing branch in order to keep it clear (to myself at least) what changes resulted from what fixes.  I, for one, would very much appreciate separate commits for this sort of thing, even in cases where you know that the expectations being accepted as a result of the change in default behavior are wrong because they reveal a bug.

eroux:
You can ignore the PR on tests, as it doesn't contain the latest changes. You can, of course, make five or six different PRs on both gregorio and gregorio-test, differentiating the different steps if you have time.

eroux:
Also, looking at some examples provided by some contacts, my personal taste is that it is quite unpleasant when the text is right of the bar, while it's ok when it's on the left (but that's just my personal taste). I think a good way to handle that would be to split `maxbaroffset` into `maxbaroffsettextright` and `maxbaroffsettextleft`, what do you think?

eroux:
Sorry, I couldn't help myself, I implemented it

rpspringuel:
Well, after some considerable work sorting out what's what, I've run out of time to actually do anything today so I'm going to simply list the things I've observed:

1. My method of offsets assumed that the offsets would be relatively small (though I didn't realize this at the time).  As a result, when the offset limit is set fairly low (and thus the offset can get large), this can result in the item being offset being overlayed with the adjacent syllable.  This appears to be happening more often with the text, but that may simply be a product of the examples in gregorio-test
2. There is a bug in the implementation of asymmetric offset limits as the bar lines are no longer centered between the notes of their adjacent syllables, even when there is no text under the bar.
3. There appears to be something wrong when a manual line break occurs directly after a bar.  I'm seeing several tests which are introducing an overly large space at the beginning of the subsequent line when this happens (e.g. gabc-output/varbraces.tex)

I've pushed a two commits on gregorio-test: the first is accepting new results for tests which appear to be good, the second modifies the bar-spacing-new test to work with the asymmetric offset limits and show the bug with the limits set to 0.  There are still plenty of tests which fail, apparently exhibiting one of the above problems.

eroux:
A few remarks

- I've been thinking about the best way to deal with low values of maxbaroffset, and it wil require a big amount of work to do it properly, for instance in the case where text should be left and would collapse into next syllable, it should first add some space around the bars, but with a maximum of `baroffsetadditionalspacearoundbars` (should be around 1 or 2mm by deafult) so that for short values, the bar stays centered. Then if it's not enough, it should just avoid trying to center the bar, and just use something more simple. Maybe these should be in different macros so that the code is easier to read
- about point 2, that doesn't sound very serious, but thanks for noticing
- about point 3, I think this is due to the order in which spaces and penalties are output, but from my experience, this kind of bug is not easy to deal with and often requires a lot of work

So what I propose is put back the old algorithm by default, merge this, and release rc1 with a KNOW BUG entry about #921, then put all the points here in high priority for this week, so that we have something working for rc2... What do you (all) think?

eroux:
new algorithm is turned off by default, I think this is ready to be reviewed and merged, I'm updating the tests

eroux:
In fact I don't know the best solution, maybe the best is to keep the new algorithm by default and document the bug... I let you decide

rpspringuel:
Going piece by piece on my own branch, I've been able to discover a few things about the issues above:

1. I haven't done anything with this yet.
2. The centering problem I mentioned above (point 2) had nothing to do with the asymmetric offset limits.  Instead there was a sign error in the calculations of the offsets which predated that implementation (as well as the typo which @eroux caught and fixed in 3d691e7).
3. The problem caused by the manual line breaks is probably in the punctum mora fix.  This remains the only thing from @eroux's branch which I haven't ported over to my branch and I'm having trouble generating this bug on my branch (though admittedly I haven't yet exhausted all possibilities).

eroux:
Good! If you decide to fix point `1` before rc1, tell me if you don't understand the algorithm I described (my description is certainly completely obscure). If not, I think a reasonable option would be to put some constraints on the bar and the text for them not to melt into previous notes or text (and keep being separated from these by the minimal space), or doing nothing, that's ok too

rpspringuel:
I'm looking harder at my branch at the moment to see if I can reproduce the line break problem there.  If I can't, then I'll bring in the punctum mora fix, check for it again, and if I find it, start working on a fix.

Assuming that doesn't take all day, I'll start looking at issue 1 later.

eroux:
Great, thanks a lot!

rpspringuel:
Okay, it took a while, but I believe I've isolated the line break bug.  It occurs when the line break is in the same syllable as the bar line (i.e. `(:z)` type constructions).  In these instances the `\GreNewLine` (or `\GreNewParLine`) is part of the "notes" which make up the bar line and thus happens in the middle of the placement of the bar line.  Since they are placed first, the placement of the text associated with the bar then breaks TeX's "ignorance" of beginning of line spaces, causing all of our shifting to finish off the bar syllable to show up, displacing the start of the next line from it's proper position.

I don't yet have a plan for dealing with this bug.

eroux:
Well, the diagnostic is already a very important step! I see two ways out of this:

- the C code could separate the line break from the bar and give it as separate arguments
- the TeX code could (in this context only) override `\gre@penalty` to something not actually setting a penalty but just saving it, and then, if a penalty has been saved, it would be output in the correct place

there are certainly other better solutions

rpspringuel:
I don't know how this would effect other things, but expanding on your first suggestion, what if a line break was always its own syllable?  I.e. `test(::z)` would become functionally identical to `test(::) (z)`.  Could the C code parse something like that?

Other permutations to consider:

* `test(z:)` -> `(z) test(:)`
* `test(:z:)` -> `test(:) (z) (:)`
* `test(gz)` -> `test(g) (z)`
* `test(zg)` -> `(z) test(g)`
* `test(gzg)` -> `test(g) (z) (g)`
* `test(:zg)` -> `test(:) (z) (g)`
* `test(gz:)` -> `test(g) (z) (:)`

I realize that some of these are really strange, but if we go this route we'd need to settle which are allowed (and which aren't) and if the above proposed interpretations are acceptable.

tex/gregoriotex-syllable.tex
```diff
@@ -1075,9 +1116,9 @@
         \gre@skip@temp@two=0 pt\relax%
       \fi %
       \gre@debugmsg{ifdim}{ previousenddifference < 0pt}%
-      \ifdim\gre@dimen@previousenddifference < 0 pt\relax%
+      \ifdim\gre@dimen@temp@three < 0 pt\relax%
         \advance\gre@skip@temp@two by %
```
rpspringuel: This change and all subsequent ones in this file are in the old algorithm, not the new one.  Are they really necessary or did you get a little over zealous with find-and-replace?

eroux: sorry, you're right

eroux: fixed

eroux:
That's an interesting point... I think we should consider it for 5.0 though, as it will certainly have unexpected consequences we don't have time to deal with. For now just patching the C or TeX code to fix this particular problem looks like a better solution. Also, I don't think it's a good idea to modify the structure of the way Gregorio sees gabc, because `test(gzg)` contains the information that the second `g` is inside the `test` syllable, while `test(g) (z) (g)` doesn't. But maybe we could just experiment with fixing this on the pure TeX side? I'll take a look this morning if I have time

eroux:
@rpspringuel I've fixed the problem on a pure-TeX level, the patch is quite small so I think we can keep it this way, tell me if it fixes the problems you encountered (it fixes the cases I've tried)

eroux:
@rpspringuel how do you think we should proceed about that? Do you think you'll be able to fix #921  today or tomorrow? If not, maybe the best option is to release 4.1-rc1 now (after merging this branch)? Do you need some help on fixing #921?

rpspringuel:
I have a fix for #921 in the works, though not the strategy you outlined above.  I should be able to commit it within the next half-hour or so (I'm running final tests on it as I write this).

eroux:
Your fix for #921 gives

bugbars321456987

which can still be improved. You did the exact first step of the technique I was talking about above, so it's very good. The thing is not to make the spacings grow too much, so if you see you have to add more than `baroffsetadditionalspacearoundbars`, you need to just give up on centering the bar between the notes, and get a result like

bugbars321456988

where the constraint is the minimum space between the notes of previous syllable and the bar, and the minimum space between the star and the text of the next syllable. Do you see what I mean?

rpspringuel:
I've been testing the line breaking fix and while what you've done takes care of the penalty, there's still a problem with distance.  When `Z` is used (a ragged line break) the fact that `\GreNewParLine` has an `\hfil` (or something like it, I haven't actually gone into the code to look at exactly what is there yet) in it is causing the bar line width to be reported wrong.  This results in the accompanying text being placed wrong.  I think if we suppress the `\hfil` while boxing, we should get better results.

eroux:
good idea, I'm just pushing a first attempt to finish #921 and I take care of that.

eroux:
I've just pushed something:

- attempt to finis #921, but I think my calculation of `\gre@dimen@temp@five` is wrong (better than before, but still wrong)
- use some `\gre@hskip` instead of `\kern` to have some rubber. rubber around bars are quite important, they should be the major ones actually
- fix https://github.com/gregorio-project/gregorio/commit/b50aa98af399bf0e4091f43bb675c6ecab788622 (was bogus)

eroux:
I've just pushed the change not to break line while boxing

rpspringuel:
I'm having some problems with the new line break fix, but it may simply be the way I cherry picked it into my own work.  I've been trying to sort that out on the side this morning, but had to finish a paper and now have to go to class so I'm not done yet.  I'll continue work later today.

eroux:
Looking at the commits you have in your branch, I think it comes from not cherry-picking https://github.com/gregorio-project/gregorio/commit/96b9990b7a268db4c70d08f8c2c99825acb7ed45 which includes (among others):

- using `\gre@hskip` instead of `\kern`
- using `\ifnum-1=\gre@newlinearg %` instead of `\ifx-1\gre@newlinearg %`, which should solve your problem

rpspringuel:
Ah.   I was deliberately avoiding that commit because it introduced the limit on the added space.  I was trying to keep the changes separate and didn't realize how intertwined the commits were. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

rpspringuel:
Hmm... It appears that simply changing the print order may have fixed not just #930, but also possibly the line break problem.  At the very least, the example of it which I had put into the tests is no longer problematic.  I'm running a more complete set of tests (basically the entire suite of tests with the new spacing algorithm turned on) to see if that's the case.  If I can't find it then b50aa98 and 3413e5f may be unnecessary.  This also makes me wonder if 39ec85b was needed.  How much effort should I put into this line break issue if it appears fixed?  Should I just let it lie and look back at the attempt to finish #921?

eroux:
If there's no problem with linebreak anymore, it's certainly best to revert the corresponding commits yes, but #921 has a higher priority... Also, don't forget to use `gre@hskip` instead of `kern`, once before and once after, if possible on the same skip (but once before and once after is the most important)

rpspringuel:
> If there's no problem with linebreak anymore, it's certainly best to revert the corresponding commits yes, but #921 has a higher priority... 

Then here's what I'll do: If I can find the line break error in another example, then I'll modify the test to show it but leave off trying to fix it until after #921.  If I can't find the error, then I'll work on #921 and come back to the possibility of reverting 39ec85b later.

> Also, don't forget to use gre@hskip instead of kern, once before and once after, if possible on the same skip (but once before and once after is the most important)

I've got that, though it took me two commits to get it right (once for the before, once for the after).  `\gre@hskip` is used for the allocation skip (the big one) both before and after the printing.

rpspringuel:
Question @eroux: when avoiding a collision requires more additional space than we're willing to add, should we give up on adding space entirely and just push the text and bar to avoid the collision, or add what we can and then make up the difference by pushing the text and bar away from the collision?

Also, what should be done if the push creates a collision in the other element?  If we're just pushing, then we could get stuck trying to push back and forth without ever getting a solution.  At some point we might simply have to add space.

Here's how I would design the algorithm:

1. Check for a collision in the text.  If no collision go to 2, otherwise proceed
  - Add space up to a certain limit
  - If the collision still occurs, also push the bar and the text (to maintain their relative positions) away from the collision
2. Check for a collision in the notes.  If no collision go to 3, otherwise proceed
  - Add space up to a certain limit
  - If the collision still occurs, also push the bar and the text away from the collision
3. If you found a collision in 2, then go back to step 1, otherwise finish.

In this way, at each step we're adding some space so eventually we'll have added enough space to get rid of the collision.  Would this algorithm be satisfactory or is there something else you had in mind?

As this involves a loop, it may be necessary to step into Lua to perform the algorithm, but TeX does have a `\loop` primitive which might allow this to be done in pure TeX.  I've not used that primitive before, however, so I'm going to need to experiment with it to get this right.

Finally, to report in, I did not see the line break error in my deeper search.  Since I need some feedback on the above before I can continue work on #921, I'm going to try reverting 39ec85b and seeing if the error appears again.

eroux:
I think this is really too complex, for the short deadline we have we cannot affort that kind of new development (we **must** release rc1 ASAP, and we already spent a week on #921!). Here is the way I propose do that:

1. first step just like it is now
2. if a collision is detected on text, do as it is now: add more text between the bars so that the star is pushed right.
3. if the space you need to add is more than `baroffsetadditionalspacearoundbars`, then give up on technique 2, and do the following:
  - put the text as right as you can (separating it from the text of the next syllable by at least the minimal amount)
  - if it's not enough to align text and bar, move the bar left of the needed amount, keeping it at the minimum distance of the note on the left
  - if it is still not enough, move the text of the bar plus the next syllable right of the needed amount

Do you see what I mean? I can try making a visual explanation if you need, but it's going to take a lot of time (I'm not very good with image manipulation softwares), so please tell me as soon as possible if you need it.

eroux:
Well, actually, let's forget about point 2, we'll implement that in rc2 if we have time. I'm trying to make a patch against your branch, I hope I'll have something by noon (Paris time)

eroux:
@rpspringuel I've implemented point 3 for one of the two cases on the branch `feature/bar_spacing_2` on this repo (the branch is a copy of yours). I'll implement it for the other cases and will fix a few other things

eroux:
In fact, I don't think there's any kind of problem with the other case, I cannot find any example of a bad output, so let's not fix what's working. I've cleaned up the different `\kern` and added a missing `\GreNoBreak`; I also merged `develop` in, and opened #932 (superseding this one)

Issue #922, opened 2016-02-15
new mechanism for hyphen determination


eroux:
This should be a relatively small change, so I think we could do it even after rc1 if you agree. The idea is to make sure that when there is no hyphen, the spacing between the text of the two syllables of the same word is perfect. The means consists of several parts:

- make `maximumspacewithoutdash` 0 by default
- make sure that when outputing the space between two syllables of the same word when there is no hyphen needed, we use `\kern` and not `hskip` (in `\gre@endofsyllable`), so `intersyllablespacenotes` can have its plus and minus removed and become a `dimen`

I think this should be a relatively small change (around 15 lines). What do you think?

The next one I'm not 100% sure about and we can keep it for 5.0:
- introduce `intersyllablespacenotesminus` that will more or less correspond to the minus of `intersyllablespacenotes` but be handled by hand, the idea is that when the distance between the texts of the two syllables is less than this value (very small, around 0.3mm), we just make the notes a little bit closer, this will allow less hyphens to appear in scores

eroux:
What remains for 5.0 is mainly linked with #923: it is to have a coherent rubber at the end of a syllable which is not the final syllable of a word. The particularity here is that this rubber has a `minus` set by the user but adapted to be at most the space between the text of the syllable and the text of the next syllable.

davidweichiang:
I think this was partly accomplished with syllable rewriting (#1098), and #1720 makes syllable rewriting happen after every syllable that has no space or hyphen. So think this could be closed after #1720.

PR #924, opened 2016-02-15
fix 4.1 part of #922

eroux:
4.1 part of #922. There is one bug left, not directly related to this branch: `\gre@rubberpermit` is not handling `@` catcode properly, this results in the default value of `intersyllablespacenotes@alteration` not displayed in the doc pdf. This seems small enough of a bug to be in rc1. Globally this is an important improvement.


henryso:
Would the problem with `\gre@rubberpermit` not handling `@` properly cause a problem for people using `\grechangedim` between scores (i.e., in the document body)?  If so, then we should either fix that or ensure that all dimension names don't have `@` in them.

eroux:
There are a few problems in edge cases like that indeed. The best would be to fix them, I've opened https://github.com/gregorio-project/gregorio/issues/925 for this problem

eroux:
I think I've nailed #925. Please review and merge if satisfactory

PR #929, opened 2016-02-16
Corrected the grouping of repercussive glyphs into a single element.

### LINKED ISSUE #928: Line breaks within tristropha
I think after the fix for #868, at least some elements like the tristropha `(hhh)` and bistropha `(hh)` will unexpectedly break across lines, unless the new syntax for inter-glyph space is inserted, i.e. `(h/!h/!h)`.

texshop_image

(using `4.1.0-beta3-develop-8968c0f-3305`)

henryso:
Fixes #928.

Changes some of the existing tests, indicating that what I thought looked correct was, in fact, not correct.  I also added a test that more obviously tests this fix.

I ran coverage and valgrind tests.  No issues to report.

The change log for this fix should be taken out once 4.1 is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Horizontal episema on initio debilis is now sized correctly (see [#880](https://github.com/gregorio-project/gregorio/issues/880)).
 - Elisions immediately after the vowel are now properly left out of the center (see [#907](https://github.com/gregorio-project/gregorio/issues/907))
 - A horizontal episema on the first note of a torculus deminutus is now placed correctly.  See [#926](https://github.com/gregorio-project/gregorio/issues/926).
-- Corrected determination of repercussive figures (introduced in 4.1.0-beta2).  See [#928](https://github.com/gregorio-project/gregorio/issues/928).
+- Corrected determination of repercussive figures and other edge cases (introduced in 4.1.0-beta2).  See [#928](https://github.com/gregorio-project/gregorio/issues/928) and [#929](https://github.com/gregorio-project/gregorio/issues/929).
 
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Horizontal episema on initio debilis is now sized correctly (see [#880](https://github.com/gregorio-project/gregorio/issues/880)).
 - Elisions immediately after the vowel are now properly left out of the center (see [#907](https://github.com/gregorio-project/gregorio/issues/907))
 - A horizontal episema on the first note of a torculus deminutus is now placed correctly.  See [#926](https://github.com/gregorio-project/gregorio/issues/926).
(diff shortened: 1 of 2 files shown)


rarty:
Thanks for the quick update. I did some more testing and with this fix all repeated glyphs look correct, but I now notice that some other cases also need to be accounted for (with manual non-breaking spaces or algorithmically), like the pressus `(hhg)`:


texshop_image

eroux:
The patch looks ok, but let's wait for the fix of pressus to be added before merging

henryso:
I need rules to implement a fix.  How do I differ that from some complicated set of neumes that *should* break?  Alternately, I can restore the behavior of indiscriminately gluing everything possible into a single element.  @eroux, do you have any suggestions?

eroux:
I think the rules I had in mind were to "glue" glyphs where the first is a single note (punctum or oriscus), and the second starts by this note, let's start with this, I'll ask my monastic contacts if there are more cases. So the basic idea is to handle all forms of *pressi*, *tristrophae*, *trivirgae*, figures like `(h_h_h_)`, etc.

eroux:
Sorry, maybe it wasn't clear, from a technical point of view, the rule `if previous is single note and next starts with the same pitch, let them in the same element` is the most general one, it includes all others, and it can be repeated multiple times in a row (in the case of *tristropha* for instance).

henryso:
So, going by your repetition rule, `(hhhghih)` should be a single element?

eroux:
Good point... First, I don't see why you would keep `hgh` and `ih` in the same element (it doesn't fit anywhere in the rule above)? But the most important is `hhhgh`... to be honest I do not know... maybe we can keep them in the same element? I also thought about a few more rules:

- when a punctum mora ends a glyph, I think we can break the element afterwards
- I've seen the case of `(ivHGFg)` several times, I think that if a single note is after some punctua inclinata descendens (let's not say anything about a punctum before puncta inclinata ascendens), and this note is one pitch above the final puntum inclinatum, and the note would be in an isolated element (not sure about this last condition, you can just skip it if you want), it should be in the same element as the puncta inclinata (and the virga before)

eroux:
(Well, you certainly need rules without "maybe", so let's keep `hhhgh` in the same element, we may reconsider this for rc2)

henryso:
How about `(ivHGFf)`?

eroux:
I'd say leave it together if the ambitus is < 2, `(ivHGFE)`, `(ivHGFf)` and `(ivHGFg)` are one element... that's quite arbitrary, but that's the best I can come up with right now...

henryso:
This is getting fairly complex.  I have to change the glyph determination in order to make the punctum mora thing work.  This will cause a longer delay due to coverage changing.

henryso:
Perhaps I should ask before doing this.  Currently, `(GH.I)` is a single glyph.  Should I break that into two glyphs: `GH` and `I`?

eroux:
Oh ok... why do you need such changes? This doesn't work by just inspecting if there's a punctum mora on the final note of the previous glyph? You can just ignore unrealistic cases such as `ivHG.F`

eroux:
No, you don't need to (our answers crossed, sorry)

henryso:
That should be it.  I merged develop into it and ran the coverage and valgrind tests.  Let me know if I missed something.

eroux:
Looks perfect, thanks a lot!

PR #932, opened 2016-02-18
Feature/bar spacing 2

### LINKED ISSUE #767: space around bars

### LINKED ISSUE #921: maxbaroffset not working correctly
If `maxbaroffset` is set to 0, compiling `{3}(gggggggg.) *(;) 44444{4}(g)` gives

bugbars2

in which the bar is "not perfectly aligned" with the star (to say the least). Solesmes will set `maxbaroffset` to 0 so it needs to work for rc1

### LINKED ISSUE #930: Translation pushing up lyrics under bar at end of line
When a bar syllable with a translation text assigned to it occurs at the end of a line and is followed by a line which does not have any translation text, then the lyric text for the bar syllable is not raised up above the translation text (note, I've tried varying these conditions and this is the only circumstance I can find where this bug occurs). This bug can be seen in the results of the bar-spacing-new test in db97a16. My suspicion is that textlower (the amount of the raise) is getting reset to zero a bit too soon, but I have not yet dug into the code to determine where that might be happening.

This bug also appears to affect only the new spacing algorithm, though I'm not sure why at this point.  The only difference I can see so far between the way translations are handled in the old algorithm and the new is that in the old algorithm the order was set translation, text, centered translation, notes, while in the new algorithm the notes come first.

For instance:
```
<alt>Translation text</alt>
TU[Thou art Peter, and upon this rock](ig) es(ij) Pe(i)trus,(i) *(,) et(i) su(i)per(g) hanc(hg) pe(e)tram[/](e) <i>etc.</i>[more text](::) (Z)

<alt>Punctum mora too close to bar line</alt>
3(g.) (;) (g) 4(g) (;) (g)
```

Produces:
temp

Note that "etc." and "more text" overlap at the end of the first line.

This example is extracted from tex-output/bar-spacing/bar-spacing-new.tex in rpspringuel/gregorio-test@db97a16

eroux:
- fixes #921 and #919
- seems to fix #930
- closes #767

this is good enough for me to go in rc1


eroux:
I've finally discovered a bogus case with the part I didn't change: `{3}3333333333(g) *(;) {}4444(gxgf)`:

bugbars1

and I see you accepted some tests in your branch so I accepted them too, but I find their output a bit strange, especially https://github.com/gregorio-project/gregorio-test/blob/pr-932/tests/gabc-dump/glyphs/divisio_finalis.pdf where `d` and `h` seem way too close... But we can keep these two bugs in rc1, we cannot afford another delay. Please review and merge even if not really satisfactory and please try to release rc1 today.

tex/gregoriotex-signs.tex
```diff
@@ -48,7 +48,11 @@
 % To remedy that, we define \gre@hskip to be \hskip outside a discretionary, and
 % \kern inside a discretionary. This is what these macros do:
 
-\def\gre@falsepenalty#1{}%
+% save asked penalty in the case where we don't really use it (discretionary, etc.)
+\xdef\gre@savedpenalty{0}
+
+% gre@falsepenalty just saves the last seen penalty
+\def\gre@falsepenalty#1{\xdef\gre@savedpenalty{#1}}%
```
rpspringuel: Since we no longer need the penalty delaying stuff, can't these changes be reversed too?

eroux: Sure (you can push in the branch if you want), but this kind of details could be done in an ironing pass after rc1

tex/gregoriotex-syllable.tex
```diff
@@ -213,6 +214,9 @@
   \or %
     % case of divisiofinalis
     \global\setbox\gre@box@temp@width=\hbox{\gre@fontchar@divisiofinalis}%
+  \or %
+    % case of no note
+    \global\setbox\gre@box@temp@width=\hbox{}%
```
rpspringuel: You know, it's occured to me that given that no note syllables have their own type here, it would be possible to modify the implementation to leave the output of the old bar spacing algorithm unchanged.  I realize that we don't want to delay rc1, but is this something I should do for rc2?

eroux: definitely yes, good idea!

rpspringuel:
> I see you accepted some tests in your branch so I accepted them too, but I find their output a bit strange, especially https://github.com/gregorio-project/gregorio-test/blob/pr-932/tests/gabc-dump/glyphs/divisio_finalis.pdf where d and h seem way too close... But we can keep these two bugs in rc1, we cannot afford another delay. Please review and merge even if not really satisfactory and please try to release rc1 today.

I have to admit that I may have been a bit overzelous with accepting test changes in my rush to get through this.  If you can make a list of the ones which you feel have "bad" results, I'll start looking at them after I create rc1.

eroux:
The text that look weird all seem to come from the same bug, which is quite obvious in the `divisio_finalis` test, I think that if we fix this one, the others should be fixed automatically

rpspringuel:
Well, other than what I've posted, I haven't seen any problems.  Thought, admittedly I haven't completed my check of the new test results.  I have to go to school now and won't be back until 4pm EST.  If you're happy, go ahead and merge this and I'll start work on rc1 when I get back.  If you don't merge, then I'll finish my review and merge myself before starting rc1.

eroux:
Ok, let's merge then, this will clean a few things in the issue tracker

rpspringuel:
Turns out I have a late lecture on campus that I forgot about.  It'll probably be around 8pm EST before I can work on making the release.

henryso:
I ran coverage and valgrind tests against this merge.  No issues to report.

PR #934, opened 2016-02-19
Feature/backwardsbar

rpspringuel:
These changes restore the behavior of the old bar spacing algorithm so that scores which use it are unchanged when upgrading to 4.1.

There are, however, some changes which I'm not sure we should keep because I think I'm overriding a bug fix to restore the old behavior.  These are split out into a separate commit (57af33e) in order to make them easier to identify and revert.

Finally, there is one test `gabc-output/edge-cases.gabc` for which I cannot seem to get an old version to pass.  I'd appreaciate it if someone would look that one over and see if they can identify the code changes which are causing it's failure.


henryso:
Should this not be opened against release-4.1?

rpspringuel:
*smacks head*

Of course.

PR #935, opened 2016-02-19
Feature/backwardsbar

rpspringuel:
These changes restore the behavior of the old bar spacing algorithm so that scores which use it are unchanged when upgrading to 4.1.

There are, however, some changes which I'm not sure we should keep because I think I'm overriding a bug fix to restore the old behavior. These are split out into a separate commit (57af33e) in order to make them easier to identify and revert.

Finally, there is one test gabc-output/edge-cases.gabc for which I cannot seem to get an old version to pass. I'd appreaciate it if someone would look that one over and see if they can identify the code changes which are causing it's failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+### Changed
+- The treatment of syllables with no notes has been reverted when the old spacing algorithm is used.  Such syllables are now treated as notes syllables under the old algorithm (as was the case in 4.1.0-beta3 and earlier) and as bar syllables under the new algorithm.
 
 
 ## [4.1.0-rc1] - 2016-02-18
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ Gregorio provides an [`.editorconfig` file](../.editorconfig), using an [editorc
 
 Python files must output no error when inspected by `pylint`.
 
-TeX code must use LuaTeX (more than TeX + eTeX + Omega + PDFTeX) primitives as much as possible, and, when not possible, use code compiling under PlainTeX. All lines inside macro definitions must end with `%` to avoid spurious spaces. To check that no line has been forgotten, please check that `grep -nE '^( |\\(|g|e|x)def)[^%]+$' tex/*.tex tex/*.sty` returns only Lua and metapost code or other lines where `%` is not needed.
+TeX code must use LuaTeX (more than TeX + eTeX + Omega + PDFTeX) primitives as much as possible, and, when not possible, use code compiling under PlainTeX. All lines inside macro definitions must end with `%` to avoid spurious spaces. To check that no line has been forgotten, please check that `grep -nE '^(\s|\\(|g|e|x)def)[^%]+$' tex/*.tex tex/*.sty` returns only Lua and metapost code or other lines where `%` is not needed.
 
 ### Tests
 
--- a/doc/Command_Index_gregorio.tex
+++ b/doc/Command_Index_gregorio.tex
@@ -1094,6 +1094,9 @@ Macro which formats the mode in roman or arabic numerals according to the approp
   \#1 & \texttt{1}--\texttt{8} & The mode to be formated\\
 \end{argtable}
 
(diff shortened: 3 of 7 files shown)


eroux:
Well, all these changes totally make sense to me, thanks a lot! It doesn't impact other tests like `glyph/divisio_finalis`? I admit I hoped it would...

rpspringuel:
IIRC, the changes in 57af33e did effect the spacing around a divisio finalis.

eroux:
Ok, but why aren't these changes visible in https://github.com/rpspringuel/gregorio-test/commit/e82698f6aee0fcc45f1189272ebde103e9ca66bc ?

rpspringuel:
My mistake, those changes effected the spacing around the divisio finalis in the 4 tests I dealt with, but did not effect other tests.  All other tests continue to pass without being changed from what is currently at the top of develop.

What I did to identify these four tests was look back into the history for when the syllable type first changed.  These were the only four tests which were modified at that point.  The other tests were modified later, and so must be the result of a different set of changes.  I'll dig further into the history to see if I can find which changes.

henryso:
I recall that this changed (but perhaps not in the way that concerns you) when I restored the capability to put a custos immediately after a (non-finalis) bar while crawling through the coverage testing.

rpspringuel:
After studying the history in my non-merged branches I've come up with the following:

The divisio finalis test was one of a bunch of tests to change in rpspringuel/gregorio-test@1c996fc.  According to the commit messages, this should correspond to rpspringuel/gregorio@4795d41 rpspringuel/gregorio@fdf34dd, and rpspringuel/gregorio@9ebb98e (though the last of these is only fixing some mistakes made in the first).  These commits implemented the separation between interword spacing around bars and regular interword spacing.  However, the aspect of that which is not in the new bar spacing algorithm should be covered by the changes in 57af33e (in particular the changes in `gregoriotex-spaces.tex`).  As a result I'm stumped as to why the results weren't effected.

eroux:
See https://github.com/gregorio-project/gregorio/commit/57af33ecc6fa63fa31b712c8895c52d4ea3fb302

eroux:
I think we can merge this one too, we'll fix the `glyphs/divisio_finalis` test in another PR

eroux:
I'm not sure it's the right place to report that, but I think there's a problem in https://github.com/rpspringuel/gregorio-test/blob/d90f5e1186e4bb93501caee3bc95a2973f92ab07/tests/tex-output/bar-spacing/bar-vs-nobar.pdf : if you look at the second line:

- the first star should be as left as possible (keeping a minimum distance with the previous "3"
- the second star should be centered (or should it?), with the same minimum space with the previous "3" and following "4"

problems are:

- the second star is not really centered (but maybe it shouldn't?)
- the second star is quite close to the following "4", much closer than the first star is from the previous "3"

So I don't know if the first star is too far from previous "3" or if second star is too close from following "4", but it looks like there's a problem...

What do you think? Anyway, we're quite close to the success of this very important and excellent change, thanks a lot!

rpspringuel:
Looking at it, I think you're right, there does appear to be a small problem with the alignment.  It's too late for me to work on this tonight, but I'll see what I can do in the morning and will attach the change to this PR if I manage to fix it.

rpspringuel:
Found and fixed the problem (2 actually).

eroux:
Thanks a lot!

rpspringuel:
I'm working on the tests that correspond to this PR.  I came up against some failures on which may actually be the desired fixes.  If you were planning on merging, hold off for a moment while I straighten this out.

rpspringuel:
Would some one please check my results so far by checking out this branch and the appropriate test branch and making sure that only bar-spacing-new is failing (there appears to be something wrong with the position of the text under the final divisio finalis in this test)?

henryso:
I checked out both, and indeed only `bar-spacing-new.tex` is failing.

rpspringuel:
Finally sorted that one out.  There's still the problem of the end of line stuff.  Should I try sorting that out here or in another PR?  My first attempts, at this point, will be to try bringing in one or the other of the solutions that we tried (and then reverted) for the other line breaking error we saw earlier.

eroux:
Thanks a lot! If you want we can merge this branch. I'll take a look today at the difference in the `divisio_finalis` test

henryso:
Though I only understand the changes in this pull request at a very high level, if you think this one solves the problems with the one exception noted above, I think it's better to merge and work on that in a separate pull request.

eroux:
let's do that

PR #937, opened 2016-02-19
Allow no-bar syllables to use different offset limits than normal bar…

rpspringuel:
… syllables.


eroux:
Looks good, ok for me to merge

rpspringuel:
I would hold this merge until after #921 is finished as there are going to be conflicts to be worked out.  Further #921 is already giving me enough trouble without introducing this into the mix on top of it.

eroux:
Sure, that's what I was thinking too

eroux:
If you solve the conflicts you can merge if you want

PR #940, opened 2016-02-20
Fix 921 cont

### LINKED ISSUE #921: maxbaroffset not working correctly

rpspringuel:
This should fully fix #921.

Because of how much trouble figuring this out gave me, I've added a whole lot of in line comments so that the next time some one comes back to this they don't have to go through the same trouble.  Please check to make sure they are comprehensible to some one not deep in the weeds of figuring this out.


eroux:
Apart from the small comment on the test PR, I think this is ok to be merged!

Issue #931, opened 2016-02-17
protrusion


eroux:
Looking at some tests for the new bar algorithm:

![aligncoma](https://cloud.githubusercontent.com/assets/60868/13106181/36e08be8-d567-11e5-8511-fc569c8f2e74.png)

I thnk the coma after _Petrus_ shouldn't be taken into account when centering the star. This could be specified in gabc with something like `<pr>` (protrusion): `trus<pr>,</pr>(...)`. This may also impact (with options) the distance at the end of a line (like real protrusion), `interwordspacing`, etc. The ideal would be that users can set up protrusion factors for different symbols (like "protrude coma from half its width", etc. so `<pr>` could taken an optional argument for that.

A direction to explore would be to have several characters automatically protruded (with default factor 0), like `,`, `;`, `:`, `.`, and users would change the protrusion factors of each the TeX side. (forced protrusion in gabc would still be useful with this system). This is the protrusion part of #845 at a larger scale.

eschwab:
I would be in favor of the automatic protrusion. The less things to manually put in the gabc file the better.

eroux:
Sure, what I'm thinking of is to have automatic protrusion, plus the possibility to override it in gabc.

henryso:
Is it required to "close" the `<pr>`?  I realize there's no precedent, but everything did require an explicit close, and this one can only really happen at the end of a word and so is closed by the end of the word.  Not closing the `<pr>` would simplify the parsing.

eroux:
For the sake of coherence, I think it's best if both possibilities are available... or does it cancel the gain of not closing it?

henryso:
It just makes it more complex to parse.  We would have to put in checks that the close occurs on the same syllable and only at the end.  The alternative is a new syntax, but I really don't like that idea, so I guess I'll live with the complexity.

henryso:
I think we were divided evenly between `<pr/>` (@eroux, @jakubjelinek) vs `$pr$` (@henryso, @rpspringuel).  We should resolve that before we start on this.

henryso:
In reference to the previous comment, it was decided that we will support both `<pr>` and `<pr/>`, with `</pr>` being forbidden.

henryso:
[...]

PR #945, opened 2016-02-21
Fix #845

### LINKED ISSUE #845: hyphens in margins
Hyphens should not be allowed to go far from the right limit of the staff lines. Example of undesired behavior:

hyphen

In a first step, the behavior should be the same as when there is no hyphen, in a second step we would differenciate the case where the syllable ends by an hyphen (automatic or not), and allow it to come closer to the lines (a kind of [character protrusion](http://wiki.contextgarden.net/Protrusion)). 

I'm affraid this is no easy fix, as the hyphen that poses problems is added in lua after the paragraph breaking algorithm, so even in a two-pass system, this wouldn't be easy: the first pass would tell the second to add an hyphen, and in some cases, the second would pass the syllable on the line after (because it would have become too big to stay on the line it was on), and thus we would have an instable system... Maybe the best way to deal with this is with a discretionary?

eroux:
Introduce a protrusion factor for hyphens automatically inserted by Lua (so it's a small first step towards #931). The value of 50 should prevent the hyphens from going too far on the right, and shouldn't change the output too much. I'm not entirely sure there's a completely logical explanation to this value, that's just what I found after some experiments. I tried to apply the same protrusion factor to hyphens automatically inserted by TeX, but it changed the tests a lot, in a very surprising way, so it seems unsafe to do it at this stage of the release cycle.

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -962,6 +962,14 @@ Marco to determine how much space the hyphen at the end of a line occupies for t
   & \texttt{zero} & The hyphen is considered to take up no space
 \end{argtable}
 
+\macroname{\textbackslash gresethyphenprotrusion}{\{\#1\}}{gregoriotex-spaces.tex}
+Sets the protrusion factor of hyphens at end of lines. The default value is 50, a good compromise in the tested scores.
+
+\begin{argtable}
+  \#1 & \texttt{number} & The protrusion factor, over 100
+\end{argtable}
+
+Note that a proper interface for setting protrusion factor of punctuation signs might appear in the next version, deprecating this one.
 
 \subsubsection{Clef Visibility}
 
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -982,7 +982,7 @@ Marco to determine how much space the hyphen at the end of a line occupies for t
 Sets the protrusion factor of hyphens at end of lines. The default value is 50, a good compromise in the tested scores.
 
 \begin{argtable}
-  \#1 & \texttt{number} & The protrusion factor, over 100
+  \#1 & \texttt{number} & The protrusion factor, percent
 \end{argtable}
 
 Note that a proper interface for setting protrusion factor of punctuation signs might appear in the next version, deprecating this one. Also, this protrusion factor does not apply to all hyphens (only those inserted by the Lua pass), so use it with caution.
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -985,7 +985,7 @@ Sets the protrusion factor of hyphens at end of lines. The default value is 50, 
   \#1 & \texttt{number} & The protrusion factor, over 100
 \end{argtable}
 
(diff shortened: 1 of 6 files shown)


doc/Command_Index_User.tex
```diff
@@ -978,6 +978,14 @@ \subsubsection{Hyphenation}
   & \texttt{zero} & The hyphen is considered to take up no space
 \end{argtable}
 
+\macroname{\textbackslash gresethyphenprotrusion}{\{\#1\}}{gregoriotex-spaces.tex}
+Sets the protrusion factor of hyphens at end of lines. The default value is 50, a good compromise in the tested scores.
+
+\begin{argtable}
+  \#1 & \texttt{number} & The protrusion factor, over 100
```
henryso: I understand what this means, but it took me a few seconds to "get it".  I think you should say "The protrusion factor as a percentage" or "The protrusion factor, percent".

doc/Command_Index_internal.tex
```diff
@@ -1515,7 +1515,10 @@ \subsection{Flags}
 Boolean which indicates that a hyphen after an empty first syllable should be forced.
 
 \macroname{\textbackslash ifgre@showhyphenafterthissyllable}{}{gregoriotex-syllable.tex}
-Boolean used by \verb=\GreSyllable= to decide if a hyphen should be shown after the syllable.
+Boolean set and used by \verb=\GreSyllable= to decide if a hyphen should be shown after the syllable.
+
+\macroname{\textbackslash ifgre@possibleluahyphenafterthissyllable}{}{gregoriotex-syllable.tex}
+Boolean set by \verb=\GreSyllable= indicating if the lua pass may add an hyphen after the syllable, used by \verb=\gre@calculate@eolshift= for protrusion calculation.
```
henryso: *Very minor*, but Lua should be capitalized, as it is in `Command_Index_User.tex`.

tex/gregoriotex-spaces.tex
```diff
@@ -384,6 +384,15 @@
 
 % dimen keeping the shift computed with next function
 \newdimen\gre@dimen@eolshift
+\newcount\gre@count@protrusion@hyphen@eol %
+
+% protrusion for hyphens inerted by Lua
+\gre@count@protrusion@hyphen@eol = 50
```
henryso: Just for my own edification, we normally put `\relax` after a number to for safety.  Why not here?

henryso:
I don't claim to fully understand everything that contributes to the computation, but this looks fine to me (other than the documentation notes I made to the lines).

tex/gregoriotex-spaces.tex
eroux: I'll add it for the sake of safety, but here (not in a macro), we're not in an environment where we have to avoid end of lines, and there is an end of line after the number, so TeX will see a `number` token, and then a `end of line` token, so it won't search for what's after the number. In an environment where we avoid end of lines, TeX would keep looking for what's after the number, which can lead to errors

eroux:
Thanks! comments are addressed

Issue #944, opened 2016-02-21
dissociate spaces of different bars


eroux:
See point 1 of [this mail](https://www.mail-archive.com/gregorio-users@gna.org/msg03349.html), the minimal space around the bar in `o(g) (,) et(g)` shouldn't be the same as the minimal space around the bar in `o(g) (:) et(g)`. Feel free to close the issue if it's already the case.

rpspringuel:
Currently those spaces are differentiated for bars which occur within syllables (i.e. something like `test(ghghg:ghghg)`), but not for bars in their own syllables.

Should these differentiated spaces be the same as the within syllable spaces (the `spacearound...` set of spaces) or should they be different?

rpspringuel:
Other issues:

The look-ahead argument in `\GreSyllable` and `\GreBarSyllable` doesn't differentiate between divisio minima, minor, maior, and dominica (all are case `11`).  Since this argument is used to determine the skip between the previous syllable notes and the bar, differentiating these spaces will require modifying that argument.

The new bar spacing algorithm currently isn't aware of the type of bar it's setting.  It's just a box with a certain width as far as it is concerned.  Some method of telling one bar from another will be required to determine the after bar spacing.

eroux:
To answer your questions:

1. the spaces should be differenciated, maybe with `@isolated` suffix (not a very good one, feel free to change it)
2. let's add more lookahead arguments, ok
3. it doesn't seem to change the arguments of grebarsyllable, as they need to be the same as gresyllable, the best solution is certainly to have a `\GreSetBarType{x}` written by the C code in `#8`

what do you think?

rpspringuel:
> it doesn't seem to change the arguments of grebarsyllable, as they need to be the same as gresyllable, the best solution is certainly to have a `\GreSetBarType{x}` written by the C code in `#8`

That's a possibility, but I was thinking more on the TeX side.  Currently `\gre@writebar` (which is aware of what kind of bar it is working on) puts the appropriate spaces in when the bar is in the middle of a syllable, but doesn't put any in when it's not.  I was thinking that it could put in the appropriate spaces when the bar is in its own syllable too.

eroux:
Why not indeed

eroux:
Starting on branch fix-944, feel free to push on it directly too

eroux:
[...]

PR #947, opened 2016-02-22
part of 944

eroux:
This is a small (but relatively independant) part of #944. This very slightly modifies tests (because the spacing algorithm was using `\GreCPPunctumDeminutus` instead of the correct glyph for bars... I'm not sure why it did that...).

Also, I made the space between the bars of divisio finalis tunable. I think this could be merged even without the rest of #944.


henryso:
Seems fine, other than the need to resolve conflicts (probably just the change log).

eroux:
Thanks!

PR #952, opened 2016-02-22
Fix 944

### LINKED ISSUE #944: dissociate spaces of different bars

eroux:
fixes #944. Mostly renaming, no big change in the algorithm, except the one proposed by @rpspringuel in https://github.com/gregorio-project/gregorio/issues/944#issuecomment-186901725 .

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -67,7 +67,7 @@ The spaces around bars are now:
 They have variants for the new bar spacing algorithm only:
 
 - `bar@xxx@standalone@text` (where `xxx` is `virgula` or `minima`, etc.) are used when the bar is in a standalone syllable with text
-- `bar@xxx@standalone@text` for bars in a standalone syllable without text
+- `bar@xxx@standalone@notext` for bars in a standalone syllable without text
 
 ### Hyphens going right of staff lines
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -54,17 +54,20 @@ A new algorithm has been implemented but is turned off by default. We hope this 
 
 If you do give it a try, please be sure to report any bugs you find.
 
-#### Spaces difference according to presence of text
+#### Renaming of spaces
 
-The following spaces have been added:
+The spaces around bars are now:
 
-- `spacearoundsmallbartext` - for the space around virgula and divisio minima with text underneath
-- `spacearoundminortext` - for the space around divisio minor with text underneath
-- `spacearoundmaiortext` - for the space around divisio maior with text underneath
-- `spacearoundfinalistext` - for the space around divisio finalis with text underneath
-- `spacebeforefinalfinalistext` - for the space before a divisio finalis at the end of a score
+- `bar@virgula`
+- `bar@minima` (both replacing `spacearoundsmallbars`)
+- `bar@minor` (replacing `spacearoundminor`)
+- `bar@maior` (replacing `spacearoundmaior`)
+- `bar@dominican` (new)
 
-By default, these are sized one half millimeter larger than their "non-text" counterparts.  This may cause minor spacing changes in your existing scores.  Adjust them as necessary to get the look you want.
+They have variants for the new bar spacing algorithm only:
+
+- `bar@xxx@standalone@text` (where `xxx` is `virgula` or `minima`, etc.) are used when the bar is in a standalone syllable with text
+- `bar@xxx@standalone@text` for bars in a standalone syllable without text
 
 ### Hyphens going right of staff lines
 
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -358,26 +358,25 @@
       \gre@calculate@glyphraisevalue{#1}{0}{}%
       %here we need some tricks to draw the line before the custos (for the color)
       \setbox\gre@box@temp@width=\hbox{%
-      % we type a hskip and the we type the custos
-      \gre@skip@temp@four = \gre@space@skip@spacebeforeeolcustos\relax%
-      \hskip\gre@skip@temp@four %
-      \gre@pickcustos{#1}\relax %
+        % we type a hskip and the we type the custos
+        \gre@hskip\gre@space@skip@spacebeforeeolcustos\relax %
+        \gre@pickcustos{#1}\relax %
       }%
       \gre@dimen@temp@three=\wd\gre@box@temp@width %
       % we make \wd\gre@box@temp@sign contain the width of a custos
       \setbox\gre@box@temp@sign=\hbox{%
-      \gre@pickcustos{#1}\relax %
+        \gre@pickcustos{#1}\relax %
       }%
       \gre@localrightbox{%
-      \ifgre@showlines %
-        \ifnum#1<\gre@pitch@belowstaff\relax %
-          \gre@additionalbottomcustoslineend %
-        \else\ifnum#1>\gre@pitch@abovestaff\relax %
-          \gre@additionaltopcustoslineend %
-        \fi\fi %
-      \fi %
-      \raise \gre@dimen@glyphraisevalue\relax%
-      \copy\gre@box@temp@width %
+        \ifgre@showlines %
+          \ifnum#1<\gre@pitch@belowstaff\relax %
+            \gre@additionalbottomcustoslineend %
+          \else\ifnum#1>\gre@pitch@abovestaff\relax %
+            \gre@additionaltopcustoslineend %
+          \fi\fi %
+        \fi %
+        \raise \gre@dimen@glyphraisevalue\relax%
+        \copy\gre@box@temp@width %
       }%
     \fi %
   \fi %
@@ -1983,13 +1982,12 @@
 
 %macro to end a line with a divisio finalis
 \def\GreFinalDivisioFinalis#1{%
-  \GreLastOfScore %
-  \GreNoBreak %
   \GreBarSyllable{\GreSetThisSyllable{}{}{}{}{}}{}{}{1}{\GreSetNextSyllable{}{}{}{}{}\GreLastOfLine}{}{16}{}{%
     \ifgre@newbarspacing\else %
       \gre@hskip\gre@space@skip@bar@finalfinalis %
       \GreNoBreak %
     \fi %
+    \GreLastOfScore %
     \GreDivisioFinalis{0}{}%
     #1%
   }%
@@ -1998,9 +1996,8 @@
 
 %macro to end a line with a divisio maior
 \def\GreFinalDivisioMaior#1{%
-  \GreLastOfScore %
-  \GreNoBreak %
   \GreBarSyllable{\GreSetThisSyllable{}{}{}{}{}}{}{}{1}{\GreSetNextSyllable{}{}{}{}{}\GreLastOfLine}{}{16}{}{%
+    \GreLastOfScore %
     \GreDivisioMaior{0}{}%
     #1%
   }%
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1981,7 +1981,7 @@
     \gre@hskip\gre@space@skip@bar@finalfinalis %
     \GreNoBreak %
   \fi %
-  \GreBarSyllable{\GreSetThisSyllable{}{}{}{}{}}{}{}{1}{}{}{0}{\GreLastOfLine}{%
+  \GreBarSyllable{\GreSetThisSyllable{}{}{}{}{}}{}{}{1}{}{}{16}{\GreLastOfLine}{%
     \GreDivisioFinalis{0}{}%
     #1%
   }%
@@ -1996,7 +1996,7 @@
     \gre@hskip\gre@space@skip@bar@maior %
     \GreNoBreak %
   \fi %
-  \GreBarSyllable{\GreSetThisSyllable{}{}{}{}{}}{}{}{1}{}{}{0}{\GreLastOfLine}{%
+  \GreBarSyllable{\GreSetThisSyllable{}{}{}{}{}}{}{}{1}{}{}{16}{\GreLastOfLine}{%
     \GreDivisioMaior{0}{}%
     #1%
   }%
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1616,8 +1616,14 @@
   \relax%
 }%
 
-% emits text if #1 is 0
-\def\gre@bar@text#1{\ifcase#1 \or text\fi}%
+% #1 is #2 of gre@writebar
+% #2 is #3 of gre@writebar
+% emits the suffix of the space
+\def\gre@bar@space@suffix#1#2{%
+  \ifcase#1\ifgre@newbarspacing %
+    \ifcase#2 @standalone@notext\or @standalone@text\fi %
+  \fi\fi %
+}%
 
 \newcount\gre@count@shiftaftermora
 \gre@count@shiftaftermora=5
@@ -1686,7 +1692,7 @@
 }
 
 %a macro to write a bar
-%% 1: the type of the bar : 0 for virgula, 1 for minima 2 for minor, 3 for major, 4 for finalis and 5 for the last finalis
+%% 1: the type of the bar : 0 for virgula, 1 for minima 2 for minor, 3 for major, 4 for finalis and 5 for the last finalis, 6 to 13 for dominican bars
 %% 2: is 0 if it is in a syllable containing only this bar, 1 otherwise
 %% 3: is 0 if there's no text under the bar or 1 if there is text under the bar
 %% 4: macros that may happen before the skip after the bar (typically GreVEpisema)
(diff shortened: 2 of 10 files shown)


henryso:
Are these renames for spaces that are all new to 4.1?

eroux:
no, it changes `spacearoundxxx` into `bar@xxx`

eroux:
Do you mean that these changes should appear in `UPGRADE.md`?

henryso:
If all the renamed spaces are new in 4.1, then there's no need to deprecate or mention in UPGRADE.md.  However, if you are renaming an old space, both need to happen.  I'll admit that I've lost track of the starting point of these changes, so I'll take your word for it if you say they are all new spaces.

eroux:
I've updated `UPGRADE.md`

UPGRADE.md
```diff
@@ -54,17 +54,20 @@ A new algorithm has been implemented but is turned off by default. We hope this
...
+- `bar@minima` (both replacing `spacearoundsmallbars`)
+- `bar@minor` (replacing `spacearoundminor`)
+- `bar@maior` (replacing `spacearoundmaior`)
+- `bar@dominican` (new)
 
-By default, these are sized one half millimeter larger than their "non-text" counterparts.  This may cause minor spacing changes in your existing scores.  Adjust them as necessary to get the look you want.
+They have variants for the new bar spacing algorithm only:
+
+- `bar@xxx@standalone@text` (where `xxx` is `virgula` or `minima`, etc.) are used when the bar is in a standalone syllable with text
+- `bar@xxx@standalone@text` for bars in a standalone syllable without text
```
rpspringuel: Copy/paste error.  The second line shouldn't have `@text`.

rpspringuel: Or rather, it should be `@notext`.

tex/gregoriotex-spaces.tex
```diff
@@ -961,42 +965,7 @@
   % now we check the bar width
   %as with the text, we start by borrowing the minNotesDistance left over from the previous syllable
```
rpspringuel: This comment should be updated to account for the deletion below.

tex/gregoriotex-signs.tex
```diff
@@ -1726,118 +1732,123 @@
...
+    \ifgre@newbarspacing %
+      \gre@count@temp@one=1\relax %
+    \fi %
+  \fi %
   \ifcase#1 % 0 : virgula
-    \ifnum#2=1\relax %
-      \gre@skip@temp@four = \csname gre@space@skip@spacearoundsmallbar\gre@bar@text{#3}\endcsname\relax%
+    \ifnum\gre@count@temp@one=1\relax %
+      \gre@skip@temp@four = \csname gre@space@skip@bar@virgula\gre@bar@space@suffix{#2}{#3}\endcsname\relax%
       \gre@hskip\gre@skip@temp@four %
```
rpspringuel: The stretchability of these skips is creating a problem in the new bar spacing algorithm because it assumes that the box containing a bar has a fixed size, but with these stretchable skips inside it doesn't.  Can these be changed to non-stretchable dimensions?

eroux:
@rpspringuel can you test the current state when you have time? I think it's ok... The only problem left for me is the way text under bar is handled at end of line... as the following syllable is on the next line, it looks as if the text was placed in a random position... I'm thinking to set `maxbaroffset`s to 0 when there is a forced end of line (we cannot detect that for automatic end of line, at least not reasonably in 4.1. What do you think?

eroux:
I've updated the tests. I think there's still something wrong in tex-output/bugs/fix-845, there is still a space after the double bar at the end of line 2...

eroux:
About the bug in fix-845, the strange (but kind of good) thing is that with old bar spacing algorithm it's even worse...

rpspringuel:
Things I've noticed as I look at the results:

1. If the bar syllable is in the same word as the following syllable, a hyphen may need to be added to the bar syllable's text.  We currently don't do anything about that.
2. When a line is manually ended with bar syllable with text and in a justified fashion (i.e. something like `test(::z)`) the bar lines overlap with the end of line custos and the text sticks out into the margin.  (See example below.)
3. I'm seeing what are probably the same test failures that @henryso mentioned in the test PR.

```
% !TEX TS-program = LuaLaTeX+se
\documentclass[12pt]{article}

\usepackage[autocompile,debug=barspacing]{gregoriotex}

\begin{document}
\gresetbarspacing{new}
\gabcsnippet{(c3) <alt>Translation text</alt>
TU(ig) es(ij) Pe(i)trus,(i) *(,) et(i) su(i)per(g) hanc(hg) pe(e)tram[/](e)
test(g)
test(g)
test(g)
<v>\hbox to 0pt{\rule{0.4pt}{1cm}}etc.\hbox to 0pt{\rule{0.4pt}{1cm}}</v>(::z)
test(g)
}

\end{document}
```

tex/gregoriotex-syllable.tex
```diff
@@ -919,20 +927,53 @@
...
+      % the text extends past the notes, so we need to get to the end of the text
+      \kern-\gre@dimen@enddifference%
+      \GreNoBreak%
     \fi%
-    %finally we end by adding the penalty
-    \gre@debugmsg{barspacing}{after bar penalty: \greendafterbarpenalty}%
-    \gre@penalty{\greendafterbarpenalty}%
+    \ifgre@eolshiftsenabled%
+      %this shift will shorten the line somewhat to allow the lyrics to extend under the custos should TeX decide to insert a line break after the syllable being ended
+      \gre@calculate@eolshift{\dimexpr(\gre@dimen@enddifference-\gre@skip@bar@lastskip)\relax }%
```
rpspringuel: Aren't you double counting `\gre@skip@bar@lastskip` by including it here and in the `\kern` statements below?

eroux: No, the idea is that `\gre@calculate@eolshift` must know the real enddifference in order to decide the eolshift to apply, so we have to pass it correctly, the rest is not really connected. In most cases eolshift will return 0 anyway.

eroux:
About hyphens of text associated with bars only, I think we can keep it for later, I don't see any use case where this would be useful. Many tests are not accepted yet, indeed. Thanks for the small example demonstrating a problem, I'll take a look.

tex/gregoriotex-syllable.tex
rpspringuel: This would also imply that `begindifference` isn't quite right either.  Does it make sense then to fix the calculations of `enddifference` and `begindifference` in `calculate@parposition` so that it produces the true values?  This would necessitate adjusting the kerns above which move between the bar and text.

eroux: I think we can do without it... it will just make things more complex... If we go in that direction, the best would be that the C code is writing the type of bar, otherwise things will be too complex for no benefit

eroux: this case is really complex to handle, I'll come up with a solution in 30mn or so

eroux:
Ok, end of lines should be better now...

rpspringuel:
The example I posted does look much better now.  Will you be updating the test expectations?

eroux:
@henryso indeed, it would be best if they were `dimen`, but there should be some code factorization and reorganization in gre@writebar for this to happen

eroux:
@rpspringuel, @henryso what do you think about the idea of https://github.com/gregorio-project/gregorio/pull/952#issuecomment-187681988 ?

rpspringuel:
I do not think it to be an unreasonable solution, but not having examined this case in detail would want to see some examples of the before (especially cases which are deemed "problematic") and after to assess whether it is the right one.

eroux:
@rpspringuel can you just proofread this last documentation commit (especially the part about the plus and minus value)? It seems to correspond to what I see in my experiments, but I'm not 100% sure

eroux:
Well, maybe I'm just inventing problems... we'll see if users report this problem... and we can deal with these problems in a 4.1 update

henryso:
@eroux To answer your request for opinion, I'm afraid I don't understand the logic there enough to give a constructive opinion.  I'm struggling to keep up with your pace (both of you) with these changes.

eroux:
tests accepted!

eroux:
last change doesn't impact tests

eroux:
Well, I let you review but I think we're ready for rc2 once this is merged...

eroux:
Oh, one last question: should the new bar spacing algorithm be enabled by default? I don't have a very strong opinion on that, but I believe it's good to keep the old algorithm by default, encouraging people to try the new one, and make the new one by default in 5.0, making the old one deprecated. What do you think?

rpspringuel:
If we can work out the bugs (and we're making good progress on that), then I don't have any problem making the new algorithm the default in 4.1.

However, I wouldn't mark the old algorithm as deprecated until 5.0 and only then if, after polling the user list, it was determined that no one was using it.

If there are any known bugs left in the new algorithm, when it comes time to release 4.1, then I wouldn't make it the default under those conditions and would delay the deprecation of the old algorithm appropriately (again with the same caveat about usage).

rpspringuel:
I'm seeing two test failures: tex-output/bar-spacing/bar-spacing-new and tex-output/bugs/fix-845.  Both look like reasonable results to me.

henryso:
If the burden of maintaining the old algorithm is not too heavy, I think we should leave it in indefinitely, but promise no fixes or changes to it.  It will either die out by itself in time or it will be evident that someone is using it.

rpspringuel:
Looking for bugs I noticed the following (non-bugs):

1. no-bar syllables at the end of a line look kind of odd because it places the text just under the end of line custos (e.g. run gabc-output/edg-cases with the new bar spacing algorithm on and look at the end of the first line).  Is this something we want to worry about, or should we just advise people to use manual line breaks to avoid this situation if they don't like what they see?
2. A final divisio finalis is a bit further from the syllable which preceeds it than a bar line (including a divisio finalis) any where else in the score).  See gabc-output/eol-reverse with the new bar spacing algorithm on to see and example (look at the last two lines).
3. We might want to increase the spaces around the bars with text a bit.  If gabc-output/glyphs/divisio-finalis is run with the new bar spacing algorithm, we get the "h" coming in towards the "d" in a way that @eroux thought was undesirable.
4. When a bar line occurs at the end of a line before a custos, it appears to be a bit tight to the custos.  We might want to consider using a slightly wider space under these circumstances.  See gabc-output/glyphs/horizontal_episema_tuning with the new bar spacing algorithm active for an example.

Note: I don't consider any of the above blocking, even for the final 4.1 release.  They are simply things I've noticed as I work my way through the whole battery of tests with the new spacing algorithm turned on.

It's late and I need to go to bed.  I'll resume this in the morning.

eroux:
To answer your points:

1. I don't see anything wrong in `gabc-output/edge_case`, we'll see if the output correspond to users' expectations
2. that's a bit strange indeed, but I think it's due to the last line being stretched at its maximum (at least I think so), and bars have big stretching values, so I don't think there's anything wrong. I'll try to take a look later today
3. the problem with the divisio_finalis test was that it was different from before when using the old algorithm. I don't think the output is particularly bad, that looks like what's expected with the spacing values
4. do you mean `spacebeforeeolcustos@bars`? Why not, that sounds like an easy change. I like this output though... so I let you implement it

Thanks a lot for the careful review!

eroux:
There is one last thing that annoys me: at a manual end of line, we do the computations of the bar spacing with a non-0 `nextbegindifference`, which makes absolutely no sense... A change is necessary on the C side in order to fix that, I'm not really sure yet what's the best option..

henryso:
Does the `nextbegindifference` problem bother you enough that it needs to be done in 4.1, or can that be pushed to later?  We are getting very close to the deadline and we don't want something like the ill-fated 2.4.3 to go into TeX Live.

eroux:
I really think it is a problem yes, I'm experimenting right now, the solution seems quite easy... I noticed a bug though: `GreLastOfScore` should be called before the final glyph, not at the level of the syllable, otherwise no custos will be printed if there is a line change in the last syllable

eroux:
Ok, I've pushed the new way to insert `\GreEndOfLine`, it improves the output of `fix-845` a lot. I didn't push the tests yet, i'm working on the problem with `\GreEndOfScore`

eroux:
@henryso, would you have a little time to work on the `\GreLastOfScore problem`? The idea is to call only `\GreLastOfLine` at syllable level:

```
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -3463,24 +3463,18 @@ static void write_syllable(FILE *f, gregorio_syllable *syllable,
     if (syllable->next_syllable) {
         fprintf(f, "{\\GreSetNextSyllable");
         write_text(f, syllable->next_syllable->text);
-        if (end_of_line) {
+        if (end_of_line || !syllable->next_syllable) {
             fprintf(f, "\\GreLastOfLine");
         }
-        if (!syllable->next_syllable) {
-            fprintf(f, "\\GreLastOfScore");
-        }

```

but calling `\GreEndOfScore` before the final element of the final syllable (except if the last syllable is `\GreFinalDivisioFinalis`, in which case nothing needs to be changed)

henryso:
I'll see what I can do.  Does one of the existing test show the problem?

eroux:
Yes, take a look at `tests/gabc-output/queuelength2`, there are not custos at the end of a few lines at the end while there should be

eroux:
Well, that was a hardcore debugging session, but many things improved this morning for bars at end of lines! This should improve the coherence of the output in this area! I think once `\GreLastOfScore` is fixed and https://github.com/gregorio-project/gregorio/pull/952#commitcomment-16289788 is addressed, we'll be ready for rc2. I think the fix-845 is still not perfect, but I won't have time today to fix it (it's already much better than before), we can keep it for later...

henryso:
There is a possible bug when I put `\GreLastOfScore` before the last element.  The custos is in the wrong place.  This will take a bit longer.

eroux:
Sure, no problem!

rpspringuel:
I've finished my detailed review of the results of last nights run and didn't spot any additional problems with the bar spacing algorithm.

I'm going to hold off on doing anything with `spacebeforeeolcustos@bars` until we get some feed back from the users.  It may just be that my expectations are too governed by the old algorithm at the moment.

Once the fix @henryso is working on is implemented and tested, I'm ready to merge this and create rc2.

eroux:
@rpspringuel before relasing rc2, could you address https://github.com/gregorio-project/gregorio/pull/952#commitcomment-16289788 ? it's not about `spacebeforeeolcustos@bars`, it's more about replacing `0` by `maxbaroffsettextleft@eol` and `maxbaroffsettexright@eol` in eol cases, that should be easy enough

henryso:
I think the problem with the wrong custos has something to do with the point at which the line breaks.  The output looks like:
```
\GreSyllable{\GreSetThisSyllable{}{g}{}{g}{}}{}{\GreUnstyled}{1}{\GreSetNextSyllable{}{}{}{}{}\GreLastOfLine}{105:0:1}{16}{}{%
\GreGlyph{\GreCPFlexusOpenqueueOneDeminutus}{6}{6}{9}%
{%
\GreGlyphHeights{6}{5}%
}{}{105:2:3}%
\GreEndOfElement{2}{0}%
%
%
\GreNewParLine %
%
%
\GreEndOfElement{2}{0}%
\GreGlyph{\GreCPFlexusOpenqueueOneDeminutus}{6}{4}{9}%
{%
\GreGlyphHeights{6}{5}%
}{}{105:8:9}%
\GreEndOfElement{0}{0}%
\GreGlyph{\GreCPPunctum}{4}{4}{0}%
{%
\GreGlyphHeights{4}{4}%
\GreAdditionalLine{\GreOCaseFinalPunctum}{0}{3}%
}{}{105:11:12}%
\GreEndOfElement{2}{0}%
%
%
\GreNewParLine %
%
%
\GreEndOfElement{2}{0}%
\GreLastOfScore\GreGlyph{\GreCPFlexusOpenqueueOneDeminutus}{4}{9}{9}%
{%
\GreGlyphHeights{4}{3}%
\GreAdditionalLine{\GreOCaseInitialVirga}{0}{3}%
\GreAdditionalLine{\GreOCaseFinalDeminutus}{0}{3}%
}{}{105:15:16}%
}%
%
\GreEndScore %
\endinput %
```

The salient argument of `\GreGlyph` is `#3`, the height of the next glyph, which looks correct.  The output looks like:

image
image

The strangest thing here is that the custos before the final flexus is at height 5!

I think the problem is in TeX, and given the time pressure, do either of you have any clues where I should look?

eroux:
From my (distant) point of view, it looks more like a problem with the automatic changes of bottom space across compilations... what makes me think of this problem is that the first custo doesn't have any ledger line, which would be ok if it was at its correct place... so maybe the problem is that automatic space change make the custos lower than it should?

henryso:
The first custos should be at height 6, on the lowest line of the regular staff.  It's the second custos which should be at height 4 and have a ledger line.  I'm close to being out of time to work on this until much later in the day.  If I get to that point, I'll push what I have and make a note here.

eroux:
Is there a difference between first and second compilation with your example?

henryso:
Yes, after the first pass, both custos are on the bottom line, which is correct for the first custos, but not for the second.

eroux:
Wow, that's getting weirder...

henryso:
I put debugging in `\GreNextCustos`, and it *is* being called as I expect, so I don't know why the second custos is simply in the wrong place.

henryso:
I'm out of time for now, so I've pushed what I have.  I accepted the changed tests except for queuelength2 which exhibits the problem described above.  You may find it enlightening to enable debugging on `custos` and `lineheight`.

eroux:
Ok, all tests pass now. @rpspringuel once you've created `maxbaroffsettextxxx@eol` we can merge this and release rc2

rpspringuel:
I've made the change on my computer, but have been unable to push the change because the Internet has gone out at home.  Hopefully it'll be restored before I get back home after class. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

rpspringuel:
Since this is still open, I've submitted the results of my doc check against it.  Please look it over to make sure I haven't written anything stupid and then merge if acceptable.  I'll run the def check (which generally has more false positives and thus takes longer to work through) after rc2.

It's too late tonight for me to build the release candidate, but I will do so tomorrow.

eroux:
Looks good, thanks! Merging

PR #963, opened 2016-02-24
Put GreLastOfScore before the final element (rather than syllable).

henryso:


eroux:
Thanks a lot for this!

PR #966, opened 2016-02-25
Fix #889

### LINKED ISSUE #889: Example with no margins
An Abbey producing many scores with Gregorio (and certainly many other people) have a workflow in which they produce PDFs of just the scores with no margins with TeX and then import the PDF in InDesign. The problem is that it's not obvious how to produce a pdf which has the precise vertical size of the score with TeX. A small research didn't allow me to find anything, but I got [a very straightforward solution from stackexchange](http://tex.stackexchange.com/a/291664/33382). So I think there could be a template to do that in `examples/`, and it could be used for scribus too.

eroux:
This looks safe enough to add in 4.1... what do you think?


eroux:
(doesn't have to be in rc2)

henryso:
Since this is example code, as long as the example works (which it seems to), then I'm fine with including this, but I'd feel better about it if @rpspringuel is able to vet it.

eroux:
sure, there's no emergency. This code (with adaptations) should be used in production on saturday in a French abbey producing a lot of books, both for sale and for internal use of other abbeys... so if you have comments feel free to express them!

rpspringuel:
I'm not opposed to this per se, but wonder if the examples folder in the distribution is the right place for it.  I've been playing with an idea of and on for a section of the website which would be dedicated to this sort of thing.  Take a look at https://github.com/rpspringuel/gregorio-project.github.io/tree/tipsntricks, particular the changes introduced in https://github.com/rpspringuel/gregorio-project.github.io/commit/6f071615b74c170f0e79e0f28ad4df8ca4ab2f26 and https://github.com/rpspringuel/gregorio-project.github.io/commit/63ab93faa612346bab8fa2c4158122fdfc7e3fc5 to see what I've got so far.  I have two other entries which I wanted to complete before posting anything for review, but have had to put them on the back burner as we pushed for release.

eroux:
That's a good idea! I still think the example folder is useful, for tests. So what should we do with this example? I propose to let it in `examples/` while your tipsntricks branch is not online... What do you think?

eroux:
Well, trying to reduce the problem of the "9" being cut, it seems that it's a bug in commentary, but I cannot understand it right now, it's certainly best to fix it before merging this pull request...

rpspringuel:
I agree that the examples folder is useful, as it provides an identical known working project on every installation; something which can be very useful when trying to debug someone's system.  I definitely wouldn't want to get rid of that.

If you're worried about losing track of this example, then I don't have a problem with putting it into `examples` for now, with the understanding that it will become part of the Tips'n'Tricks portion of the website.  However, seeing as I have spring break this coming week, I'm going to make a push to finish the Tips'n'Tricks stuff so that we can have that live on the website sooner rather than later.

eroux:
ok, we can wait and put this on tipsandtricks then. Would you have time to look at the issue with commentary? no emergency though

eroux:
I think that with https://github.com/gregorio-project/gregorio-project.github.io/pull/29, https://github.com/gregorio-project/gregorio/commit/da525e19f10c1b51d5140944a02227aa2c3632df doesn't make sense anymore. Should I revert it? I think the two first commits of this PR still make sense.

rpspringuel:
I agree.  Reverting da525e1 but keeping the other two makes sense.

eroux:
This should be ok to merge now

eroux:
I'm fixing a final bug, please don't merge before next commit

eroux:
Should be good now. This doesn't change test because the font Alegreya doesn't have any italic correction, but this improves the rendering of elision and fixes the shift in commentary. @rpspringuel, see https://github.com/rpspringuel/gregorio-project.github.io/commit/7e5671b9994844b1d88d3aca9e719a367b038a9a#commitcomment-16378819

rpspringuel:
Off hand I'd say this looks good.  I'll run the tests to make sure, and will merge if no flags come up.  Given the time, that will probably be in a few hours as I'll have to get ready for Vespers soon and the tests take some 30 or 40 minutes on my machine.

eroux:
Sure, no problem. Thinking back about it, maybe adding italic correction to commentary all the time may be a mistake... It's definitely useful for elision and mode, and for commentary in some cases (like when there is no margin), but for documents with margin, it may be best to let italic go into the margin... So I reverted the change of commentary style, sorry!

henryso:
I was thinking about this, and if the purpose of the "marginless" document is so that it could be pasted into another document which presumably has margins, it might be good to show how the marginless example could be modified to have a small margin on the right to accommodate for the italic correction.

PR #971, opened 2016-02-27
Get new bar spacing algorithm to respect `\gresetlastline` setting.

rpspringuel:
This solution is a bit of a hack as I simply insert a call to `\GreNewLine` into the code for a final finalis (or maior).  It works and doesn't modify too much code, but if someone has a better idea as to how to fix it, I'm all ears.

I don't have time to check the tests at the moment (and incorporate a new one), but I wanted to get this out for comment sooner rather than later.  I'll take care of the tests later tonight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
-
+### Fixed
+- New bar spacing algorithm will now respect `\gresetlastline`.
 
 ## [4.1.0-rc2] - 2016-02-25
 ### Fixed
--- a/tex/gregoriotex-signs.tex
+++ b/tex/gregoriotex-signs.tex
@@ -1991,7 +1991,7 @@
     \GreDivisioFinalis{0}{}%
     #1%
     \ifgre@justifylastline%
-      \GreNewLine%
+      \def\gre@newlinearg{0}%
     \fi%
   }%
   \relax%
@@ -2004,7 +2004,7 @@
     \GreDivisioMaior{0}{}%
     #1%
     \ifgre@justifylastline%
-      \GreNewLine%
+      \def\gre@newlinearg{0}%
     \fi%
   }%
   \relax%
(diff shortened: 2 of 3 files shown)


eroux:
Thanks! I'll inspect that further, but I think there should be a better way of doing this...

rpspringuel:
I've found a slightly less hacky way of pulling this off.

eroux:
Here's how I would do (based on vanilla 4.1):

```
diff --git a/tex/gregoriotex-syllable.tex b/tex/gregoriotex-syllable.tex
index 30b0b86..9db0f23 100644
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -946,8 +946,16 @@
       \fi %
       \kern\glueexpr(-\gre@dimen@eolshift)\relax%
     \fi%
-    % if no end of line:
-    \ifnum\gre@newlinearg=-1\relax %
+    % gre@count@temp@one is "forced end of line or last of score"
+    \gre@count@temp@one=0\relax %
+    \ifnum\gre@newlinearg=-1\else %
+      \gre@count@temp@one=1\relax %
+    \fi %
+    \ifgre@endofscore %
+      \gre@count@temp@one=1\relax %
+    \fi %
+    % if no end of line or end of score
+    \ifnum\gre@count@temp@one=0\relax %
       % add the penalty
       \gre@debugmsg{barspacing}{after bar penalty: \greendafterbarpenalty}%
       \gre@penalty{\greendafterbarpenalty}%
@@ -971,10 +979,10 @@
         \GreNoBreak %
         \kern\gre@skip@nextbegindifference\relax%
       \fi%
-    \else %
+    \else\ifgre@endofscore\else %
       \gre@debugmsg{barspacing}{calling new line with argument: \gre@newlinearg}%
       \gre@newlinecommon{\gre@newlinearg}%
-    \fi %
+    \fi\fi %
   \else%
     %
     % Old spacing algorithm
```

What do you think?

rpspringuel:
This change seems to work in most cases.  In testing it out I noticed that tex-output/bugs/fix-845 changes as a result, though, frankly, I don't know why.  That test should still be operating under the old algorithm.

eroux:
It operated under the new algorithm. The test is broken actually, it doesn't test what it's supposed to, but that will be for a future pull request...

rpspringuel:
Oh, well, in that case I think your proposal is working fine.  I've submited the change.

eroux:
Thanks, merging!

PR #980, opened 2016-02-29
Adding `interwordspacetext@bars@notext` and `interwordspacetext@bars@…

### LINKED ISSUE #978: interwordspacetext not taken into account for bars
With the following example:

```tex
\documentclass[11pt]{article} % use larger type; default would be 10pt
\usepackage{fontspec}
\usepackage{graphicx}
\usepackage{geometry}
\geometry{a4paper}
\usepackage{gregoriotex}
\usepackage{libertine}
\begin{document}
\gresetbarspacing{new}
\gresetlastline{ragged}
\gresetinitiallines{0}
\gabcsnippet{(c3) am<v>\hbox to 0pt{\hss\rule{0.1mm}{3mm}}</v>(g) <v>\hbox to 0pt{\rule{0.1mm}{3mm}\hss}</v>E(g.) am<v>\hbox to 0pt{\hss\rule{0.1mm}{3mm}}</v>(g) (::) <v>\hbox to 0pt{\rule{0.1mm}{3mm}\hss}</v>E(j)}
\end{document}
```

I get

bugbars

in which the space between 'am' and 'E' seems to be less than `interwordspacetext` (between the texts of the syllables around the bar)... @rpspringuel do you have time to inspect that? (this is quite high priority)

rpspringuel:
…notext@euouae`

Fixes #978

### DIFF
--- a/tex/gsp-default.tex
+++ b/tex/gsp-default.tex
@@ -221,12 +221,14 @@
 %
 % minimal space between letters of different syllable texts for text around bars
 % (new bar spacing algorithm only)
-% Note that the plus and minus values of interwordspacetext@bars or interwordspacetext@bars@euouae
-% are what will be used around bars in the new bar spacing algorithm
 \grecreatedim{interwordspacetext@bars}{0.15 cm}{scalable}%
 % minimal space between letters of different syllable texts for text around bars,
 % euouae context
 \grecreatedim{interwordspacetext@bars@euouae}{0.15 cm}{scalable}%
+\grecreatedim{interwordspacetext@bars@notext}{0.3 cm}{scalable}%
+% minimal space between letters of different syllable texts for text around bars,
+% euouae context
+\grecreatedim{interwordspacetext@bars@notext@euouae}{0.3 cm}{scalable}%
 % rubber length that will be added around bars in new bar spacing algorithm
 \grecreatedim{bar@rubber}{0 cm plus 0.22787 cm minus 0.02 cm}{scalable}%
 % additional space that will appear around bars that are preceded by a custos and followed by a key.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Changed
 - `spacebeforeeolcustos` has been increased, aproximately doubling its size.  It's also been made fixed by default, but is still capapble of accepting a rubber value should you so desire.
 
+### Added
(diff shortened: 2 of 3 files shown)


tex/gsp-default.tex
```diff
@@ -221,12 +221,14 @@
...
-% Note that the plus and minus values of interwordspacetext@bars or interwordspacetext@bars@euouae
-% are what will be used around bars in the new bar spacing algorithm
 \grecreatedim{interwordspacetext@bars}{0.15 cm}{scalable}%
 % minimal space between letters of different syllable texts for text around bars,
 % euouae context
 \grecreatedim{interwordspacetext@bars@euouae}{0.15 cm}{scalable}%
+\grecreatedim{interwordspacetext@bars@notext}{0.3 cm}{scalable}%
+% minimal space between letters of different syllable texts for text around bars,
+% euouae context
+\grecreatedim{interwordspacetext@bars@notext@euouae}{0.3 cm}{scalable}%
```
eroux: I wound tend to make it a big bigger, something like 0.35 or even 0.4... But it's not that important, the main issue is that it's now possible to tune it!

eroux:
Apart from the small comment (don't change the space value if you're already running the tests, we can change it in a future version), it looks fine, thanks a lot!

rpspringuel:
I've run and posted the updated tests already.  If you want me to change the value, it will require updating the tests again.

eroux:
No problem, let's merge it like this. (I'm counting on the italian Antiphonale Monasticum project to get some fine tunings for all these spaces, we'll see what they come up with!)

eroux:
Sorry, I think I tend to merge a bit too fast... Did you try with the example of https://github.com/gregorio-project/gregorio/issues/978 ? The output is better, but the second space is still significantly smaller than the first... do you understand why?

rpspringuel:
I did try and got exactly what I expected to get:

temp

The lower set of lines are placed manually outside of the score for the purposes of showing what I expected the spaces to be.  As you can see they match what is in the score.  If the space is too narrow, then you simply need to increase `interwordspacetext@bars@notext`.  Here is it increased to 0.5cm:

temp

rpspringuel:
By default `interwordspacetext` is 0.38cm, so it should be larger than the default I choose for `interwordspacetext@bars@notext` (which is 0.3cm, or twice `interwordspacetext@bars`).

eroux:
Oh sorry, I "misread" the value of `interwordspacetext` (though here it must be stupidity, not misreading). Well, I think `interwordspacetext@bars@notext` should be at least 0.38 then, let's make it 0.4. Sorry for my stupidity!

PR #991, opened 2016-03-01
Changing default to new bar spacing algorithm

rpspringuel:
Note: I had to move the block of code to after the definition of `\GreBarSyllable` in order for the `\let` statement to work properly.


henryso:
This looks fine to me.  As for the tests, comparing the before/after, I can't qualitatively say one is better than the other, so I suppose they are fine as well.

rpspringuel:
I was a little surprised by that.  Most every test changed, and in some cases I can say the new result looks better (or at least more compact), but there are also changes in the other direction.  I suppose that could be due to the artificiality of many of the tests (the ones that most closely resemble real scores are the ones where I think there is most clearly an improvement), but it may also be a sign that something isn't quite right yet in the algorithm or with the default spacings we've chosen for all the new spaces.

Still, I think that out best bet at this point is to make this the default and then respond to any bugs reported.

henryso:
I'm not disagreeing.  I'm just not a particularly good judge of things in this case.

PR #993, opened 2016-03-02
Fix (::Z)

eroux:
This is really a tiny fix, one small test change


eroux:
Sorry, I pushed the first commit in the main branch...  I'm doing this in a hurry...

rpspringuel:
The changes look good.  I'm running the tests to double check, but will probably merge a little later today.

eroux:
Ok thanks, the changes should be noticeable only in twisted edge cases, I produced one quite artificially, but I cannot really reproduce it... it should prevent a few bugs from appearing in other (real) edge cases though

PR #1000, opened 2016-03-03
Fixes #984

### LINKED ISSUE #984: line not ragged before euouae block in some cases
Compiling

```tex
\usepackage{libertine}
\begin{document}
\gresetbarspacing{new}
\gresetbreakbeforeeuouae{ragged}
\gresetlastline{ragged}
\gregorioscore{a--prophetae-predicaverunt-MWE3}
\end{document}
```

with gabc

```
name: Prophetae predicaverunt;
office-part: antiphona;
mode:1;
%%
(c4) Pro(d)phé(ddo>)tæ(c) pre(d)di(f)ca(e)vé(d)runt(dc~) *() na(f)sci(g) De(h)us(g) Sal(h>)va(g)tó(f_0fg)rem(f_) (,) de(f) Vír(f)gi(g)ne(e) Ma(f)rí(d.)a.(d.) (::) <eu>E(h) u(h) o(g) u(f) a(gh) e</eu>(ggo>) (::)
```

one gets

bug2

where the second line is stretched quite a lot, comparing to a version with a `Z`:

bug3

eroux:


eroux:
@henryso there's still one test not passing in a very specific case that requires a change in the C code, it's the very last line of `tex-output/eol-euouae/eol-euouae.tex`. The idea is that now when the line before euouae must be ragged, we break the line in `GreNextSyllableBeginsEUOUAE`; the problem is when there is `g(g) (::) (z) <eu>E(g) u</eu>(g)`, in which case we must not break the line. Could you make a second argument to `GreNextSyllableBeginsEUOUAE` which value would be `1` in the case where there is any line break before euouae already, `0` otherwise? If you have time, just the C part is enough, I'll make the TeX and documentation. Otherwise I'll do it tonight.

henryso:
This should not be a pull request against master.  I will see if I have time this morning to look into it.

eroux:
Sorry, thanks for noticing

PR #1001, opened 2016-03-03
Fix 984

eroux:


henryso:
I made the change you requested.  I think it does what it's supposed to, but I haven't made any TeX changes, so obviously the tests break.

tex/gregoriotex-main.tex
```diff
@@ -1644,15 +1644,18 @@
 \newif\ifgre@in@euouae
 \gre@in@euouaefalse
 
+\newif\ifgre@firstin@euouae
+\gre@in@euouaefalse
```
rpspringuel: As is this line is a repeat of 1645.  I think you meant `\gre@firstin@euouaefalse`

eroux: thanks, addressed

eroux:
should be ok now

rpspringuel:
There are merge conflicts now, but other than that this looks good.

eroux:
should be ok now

PR #1003, opened 2016-03-03
Fixing double error message

rpspringuel:
I put the code marking the deprecated distances in the wrong place.  As a result it raised both the deprecation warning/error and the unrecognized distance error.  This change makes it so that only the first occurs when the user attempts to change a deprecated distance.


eroux:
Looks good to me!

PR #1004, opened 2016-03-03
Feature/prep 4.2

rpspringuel:
While we haven't decided exactly what form the next release will take.  This would be the first step for a 4.2 release (should we decide to go that route).

Obviously bug fixes for 4.1 (i.e. progress towards 4.1.1) wouldn't incorporate these changes, but those PRs should be against the release-4.1 branch and probably won't impact this code (though #1003 shows that this is possible) so I'm fairly sure it's safe to merge this into develop now.

Merging this code will apply the effects of #1003 to the develop branch.

If/when we decide to go for a 5.0 release, then additional changes will be required to move the initial related code to OBSOLETE status.  One of the commits here makes sure that all the comments marking such code are consistent so that they are easy to find when the time comes for that.


eroux:
ok for me!

PR #1006, opened 2016-03-04
Added \GreLastSyllableBeforeEUOUAE into the syllable before EUOUAE.

### LINKED ISSUE #988: There should be a \GreLastOfLine-like macro in the syllable before an EUOUAE
See https://github.com/gregorio-project/gregorio/issues/984#issuecomment-190784677

This macro should use the same calculation as to whether the `\gre@newlinecommon{1}` is inserted in `\GreBeginEUOUAE` to determine whether it should call `\GreLastOfLine`

henryso:
Fixes #988.

I implemented my understanding of the fix, but no output tests changed, so I'm unsure if I did it correctly.  Please review and let me know what I missed.  Otherwise, if this is satisfactory, please merge it.


eroux:
Looks good, thanks!

PR #1009, opened 2016-03-04
tune spacings around custos at end of score

### LINKED ISSUE #990: New bar spacing algorithm pushes custos into margin
I was going through the process of making the new algorithm the default and found this during the test review:

temp

The test is tex-output/bugs/fix-43.

eroux:
Fixes #990


eroux:
This is not perfect and these spacings should probably be tuned futher, but at least it fixes the bug and gives nicer spacings than before

henryso:
This looks OK to me, but I'm terrible at judging the "beauty" of spaces.

Issue #1023, opened 2016-03-07
UPGRADE.md and empty syllables


henryso:
Since empty syllables are now bar syllables, perhaps something should be noted in UPGRADE.md describing this (and the fix if you don't like the current result).

eroux:
Indeed. Also, maybe empty syllables could raise warnings? There's no defined behavior for that and the output may change from version to version...

henryso:
Do we still want empty syllables to raise warnings?  That's not done yet.

eroux:
I think it would be a good idea yes... but we can leave it for 4.2.

henryso:
Ok, reopening this issue then.

henryso:
I think we need to try to get things printed to stderr from gregorio to print when doing auto-compiles from Lua.

henryso:
As I start coding this and looking at the tests, I find that we are using a no-note syllable expressly for some cases of #767.  We even emit a special `\GreNoNoteSyllable` in this case.  Is that really undefined behavior?  Aren't we, by doing this, defining the behavior?  Or am I missing something?

rpspringuel:
I don't think that the no note syllable is undefined.  At least as I thought about it in implementing the new spacing algorithm, the no note syllable is a type of bar line (i.e. should use `\GreBarSyllable`), call it a divisio nihil.  This was a deliberate choice in order to account for some of what I was seeing in the examples in #767.

The issue is that this is different from the old spacing algorithm, where the no note syllable was treated as a type of note (i.e. used `\GreSyllable`), call it a punctum invisum.  IIRC, this old behavior wasn't necessarily anything intentional, just the result of the fact that gregorio made the simple assumption that whatever wasn't a bar syllable, was a note syllable.  One could, therefore, consider this behavior to have been undefined, though it did operate in a predictable fashion.

The use of `\GreNoNoteSyllable` is the intermediary which allows the code to negotiate between these two different ways of treating the syllable.  It is not meant to imply that no-note syllables are their own type, but rather allows them to be treated as one type (punctum invisum) in the old algorithm and as something else (divisio nihil) in the new.

Further, this aliasing does not have to be tied to the old and new algorithms, though our user function changes them automatically.  With a simple `\let` statement, the end user could switch how a no-note syllable is treated without changing the bar spacing algorithm.

To that end, I don't think errors or warnings are needed here, but rather documentation explaining how things are defined and how the user can change them.

henryso:
[...]

PR #1028, opened 2016-03-11
Log gregorio messages

henryso:
The idea was noted in https://github.com/gregorio-project/gregorio/issues/1023#issuecomment-194902969

Since it appears we don't need further implementation on #1023, I am submitting this as its own pull request.

### DIFF
--- a/doc/Command_Index_internal.tex
+++ b/doc/Command_Index_internal.tex
@@ -48,6 +48,11 @@ Macro that handles obsolescence errors.
   \#2 & string & name of the correct macro to use\\
 \end{argtable}
 
+\macroname{\textbackslash gre@allowdeprecated@asboolean}{}{gregoriotex-main.tex}
+Macro that evaluates \texttt{true} or \texttt{false} depending on whether
+deprecated usage is allowed.  This is used to pass the value of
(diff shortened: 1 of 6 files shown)


eroux:
Looks good, thanks! Did you test it in an unfriendly environment by tuning some kpse restrictions? It would work, it's just in case...

henryso:
It works fine for normal scores.  I use the same filename transformation (but using `.glog` instead of `.gtex`), so if kpathsea allows write to the `.gtex` file, it should allow write to the `.glog` file.

Snippets, on the other hand, pose a problem.  We use `os.tmpname()` to get a filename for writing the gabc from the snippet.  LuaTeX allows this, and the "read anything" typical permission for kpathsea will allow these to be read.  The current implementation (in this pull request) also uses `os.tmpname()` to get a filename for the log file.  However, the kpathsea inside the gregorio executable will reject a write to this file.

Should I disable the logging when using snippets?

henryso:
I think I've come up with a workaround.  I'll try to prepare it tonight.

henryso:
This version works on a restricted kpathsea environment unless read access is too limited to load the gregorio package.

eroux:
Thanks!

PR #1032, opened 2016-03-14
Rename `deprecated` in CHANGELOG and UPGRADE.

eschwab:
The `deprecated` option was added in fc11bc2f786a944f9c135b60c0b0502d2883e07e

It was renamed to `allowdeprecated` in 56d4814a06ef7948afa334ad5c96b8450ab9ada7

But CHANGELOG and UPGRADE were not changed. This all happened before `4.0` so I don't think anything else needs to be mentioned in the CHANGELOG.


henryso:
Thanks for noticing.

PR #1037, opened 2016-03-16
Implemented torculus starting with oriscus using fusion.

### LINKED ISSUE #1013: "syntaxic sugar" for some fused glyphs
Solesmes uses the glyph `(go@hd~)` (a torculus liquescens with first note being an oriscus). The only way to get it now is through fusion, but it would be best if `(gohd~)` had the same output (using fusion behind the scenes). Maybe the same principle could be applied to the `pes stratus` of https://github.com/gregorio-project/gregorio/issues/898 ? Anyway, this is not a top priority.

henryso:
Fixes #1013.

All current tests pass and I added a case for this new figure.


henryso:
Which tests evidence the fix?  Perhaps you should add the example from the issue?

eroux:
Thanks a lot!

PR #1040, opened 2016-03-17
Set glyph id and heights for custos glyphs.

### LINKED ISSUE #961: low Custos overlapping with text (edge case)
In gabc-output/lines/2-lines, 3-lines and 5-lines the low notes an the lyrics overlap in the third line.  Is this because of they are custos, or because there's a problem with the line height calculations?

henryso:
Fixes #961.

The updated test results look correct to me, but please confirm.  I'm not sure I handled the end-of-line custos correctly, so double-check that, please.


eroux:
This looks good, thanks a lot!

PR #1041, opened 2016-03-19
Wrote computed results to the .gaux file rather than measurements.

### LINKED ISSUE #987: Table difference check for determining when the gaux file should be saved is too fragile
See https://github.com/gregorio-project/gregorio/issues/984#issuecomment-190780675

henryso:
Fixes #987.

Output appears to be stable.


eroux:
Looks good to me, ok to be merged. Thanks a lot!

PR #1043, opened 2016-03-19
Added a tunable space for puncta inclinata at the unison.

### LINKED ISSUE #1042: ability to tune space between two unison puncta inclinata
I'm not really sure about the space inserted between the puncta inclinata of `(GG)` for instance, but it should be tunable on its own.

henryso:
Fixes #1042.

I set the `punctuminclinatumunisonshift` to the same as `beforepunctainclinatashift`, which is what is currently used in this case.  No current tests break, and though punctum cavum inclinatum at the unison is already tested, I added a test specifically for the non-cavum case.


eroux:
Looks perfect, thanks a lot!

PR #1047, opened 2016-03-20
Determined a podatus followed by a virga as a single element.

### LINKED ISSUE #1045: improving scandicus determination
gabc `(eghv)` should be considered as only one element (a scandicus), and thus no line break should occur between `eg` and `hv`. The scandicus can have two forms for Solesmes: `eg!hv` and `eghv`, depending on the original adiastematic notation.

henryso:
Fixes #1045.

All current tests pass, so I added a test to exercise this code.


eroux:
Wonderful, thanks a lot! Ok to be merged

PR #1030, opened 2016-03-13
Added missing directional oriscus figures

henryso:
This includes full fusion support and covers the open items of #898 and its duplicate #972.

Note: I based this off of #1028, so merging this will merge that as well.  This is because I was forgot to change my branch when I started work.  If #0128 is to be rejected, I will rebase these changes.

Note 2: While I was coverage testing this change, I found a bug when passing a filename with no extension to gregorio, which I fixed.

I got somewhat confused during implementation, so any help testing with new and interesting combinations would be appreciated.  I have updated the tests to my satisfaction.

If this is found satisfactory, please merge it.

PR #1048, opened 2016-03-21
Added the ability to set the minimum number of notes to keep together at the start and end of a syllable.

### LINKED ISSUE #1044: minimal grouped notes at syllable extrema
A request from Solesmes: it would be best to leave the `x` first notes of a syllable together (meaning: keeping them on the same line), and the `y` last notes too. These values should be configurable on the TeX side. I think it should not be too difficult by adding two arguments to GreEndElement, indicating the position of the note from the beginning and from the end...

henryso:
Fixed #1044.

In adding the documentation for the new TeX macros, I was forced to fix document compliation breakages caused by #1030.  I also incidentally added the missing gabc sequences for completeness.

The gabc-gtex tests obviously break for this, and some of the existing gabc-output tests are good enough to show this change at work, so I just accepted the results.


eroux:
Looks perfect, thanks a lot!

Issue #1039, opened 2016-03-16
text-only mode


eroux:
It would be useful for some Abbey if GregorioTeX had a text-only mode that just prints the text, with hyphens between each syllable and just a simple space between words. The idea is to typeset a document in this mode and give the result to someone who would proofread hyphenation.

I think this should be simple enough by overriding `\GreSyllable` and `\GreBarSyllable` with a simple `\GreTextOnlySyllable`.

rpspringuel:
This is mostly already possible via existing settings.  If you generate a score with the following settings:

```
\gresetinitiallines{0}
\gresetclef{invisible}
\gresetlines{invisible}
\gresethyphen{force}
\grechangestaffsize{1}
\grechangedim{interwordspacetext}{0.5em}{fixed}
\grechangedim{interwordspacetext@bars}{0.5em}{fixed}
```

the results are pretty much as you describe.  The notes are still technically being printed, just very small to the point of being almost (but not quite) unnoticeable.  If this isn't sufficient, then I could add a flag which makes the notes truly invisible.  

A parallel flag to make the lyrics invisible would then complete the set, allowing the user to toggle the visibility of pretty much every element of the score.

eroux:
Interesting, thanks for that!

It would be best if the notes could disappear completely indeed, but maybe just handling `\grechangestaffsize{0}` correctly should be enough.

End of lines should also not be handled in this mode, but it's not a big problem, we can close this even if they're not handled.

PR #1053, opened 2016-03-27
Feature/hidden

rpspringuel:
Addresses #1039.


eroux:
Thanks a lot! That looks good, ok for me to be merged, maybe you could just add a test that also completely hides the staff lines, key, custos, etc. and puts hyphens between all syllabes? As it's the main use case, I think it's a good idea to have a test for it, even though it's not a feature per se, but a mixture of many.

rpspringuel:
Done.  See rpspringuel/gregorio-test@6f5921f.

eroux:
Perfect, thanks a lot!

PR #1060, opened 2016-04-03
Prevented an element cut before an isolated stropha.

### LINKED ISSUE #1056: preventing linebreak between anything and isolated stropha
Gregorio should not cut lines before an isolated stropha. Isolated stropha means here that it's not in a bi/tristropha. And of course this rule applies only to stropha not beginning a syllable. I admit I'm a bit puzzled by the way we should deal with that... should it be merged in the element before it? Why not... What do you think? (that's a request from my contact at Solesmes)

henryso:
Fixes #1056.


eroux:
Looks perfect, thanks!

PR #1063, opened 2016-04-07
Fixing problems in `\gre@changestyle` definition under PlainTeX

rpspringuel:
Need to lead with `\expandafter` when defining a function name which is specified with `\csname`


eroux:
Thanks!

Issue #962, opened 2016-02-24
Distance inheritance


rpspringuel:
Given the proliferation of circumstantial distances, it might be useful if we set up a system whereby one distance could be specified as inheriting its value from another.

Something like `\greinheritdim{bar@virgula@standalone@notext}{bar@virgula@standalone@text}` would indicate that the user wants `bar@virgula@standalone@text` to inherit the value of `bar@virgula@standalone@notext` so changes to the latter would effect the former.

This could also be implemented as a feature of `\grechangedim` but the third argument (`scalable` or `fixed`) becomes redundant in this case.  Perhaps it would be better if that argument was made optional if we decide to go that route.

henryso:
How about `\grechangedim{bar@virgula@standalone@notext}{bar@virgula@standalone@text}{inherited}` ?

rpspringuel:
Hadn't thought of that.  The only potential problem I can see is what do we do if a user enters something like: `\grechangedim{bar@virgula@standalone@notext}{bar@virgula@standalone@text}{scalable}`?  Would raising an error be enough?

henryso:
I think raising an error would be enough in that case.

eroux:
I thought about an inheritance system based on the name (`foo@bar` would automatically inherit from `foo` if it exists), but that's just too complex. @henryso 's proposal looks good to me

PR #1066, opened 2016-04-10
Inherited distances

rpspringuel:
Addresses #962
Adds `inherited` option for third argument in `\grecreatedim` and `\grechangedim`, allowing the user to link one distance to another.


eroux:

PR #1067, opened 2016-04-10
Fix gprocess to work on my system.

avbop:
I don't know if this is necessary elsewhere, but I was getting errors and this change fixed them. The main change is switching from `luainputenc` to `inputenc` in line 173. I also added another run of `lualatex` for the line height calculations.


avbop:
Actually, I can remove inputenc altogether; it gives a warning if included.

eroux:
Not sure it's used much, but thanks a lot! It looks good to me, ok for the merge

Issue #1065, opened 2016-04-08
Debug messages overlapping keyword problem


rpspringuel:
Right now we have two debug message categories `barspacing` and `spacing`.  If one sets the package option `debug=barspacing`, `debug=`{general,barspacing}`, or any other list containing `barspacing` then the `spacing` messages will also be printed to the log because the check to see if a message is to be printed is `\IfSubStr{keyword}{\gre@debug}` and `spacing` is a substring of `barspacing`.

To get around this, either keywords should not be set such that they are substrings of or contain other keywords (meaning that `barspacing` needs a new name), or we should redefine the mechanism such that only "whole word" matches are allowed.

On the path of the second, I would propose the following:

1. After `\gre@debug` is set, the string it contains is manipulated to remove braces (xstring provides ` \StrRemoveBraces` which can do this) and add a comma to the beginning and end of the string.  This would make it so that `\gre@debug` looks like this: `,general,barspacing,otherkeyword,`
2. The check to see if a message should be printed then looks for the presence of `,somekeyword,` rather than just `somekeyword`.

eroux:
Your solution looks good, I don't see any problem with it

PR #1068, opened 2016-04-10
Allow only whole word matches when checking debugging keywords

rpspringuel:
Addresses #1065
Since this is a completely internal change, there is no CHANGELOG entry.
Likewise, this doesn't change the way we use these functions, so there is no need to change their documentation either.


eroux:
Looks OK to me, thanks a lot!

Issue #1072, opened 2016-04-12
remove DS_STORE files from the dist archive


eroux:
There are some `._.DS_Store` and `.DS_Store` in the dist archive, especially in the `contrib/TeXShop` directory, that should be avoided...

rpspringuel:
They must be getting swept up by rules which grab the contents of a directory, rather than a particular file.  Is there a way to add a `*DS_Store` to a list of files excluded from the dist archive?

rpspringuel:
A little research later:

At least for `zip` adding `-x "*/\.DS_Store"` should exclude all `.DS_Store` files.  Likewise adding `-x "*/\._.DS_Store` should get rid of the `._.DS_Store` files.  I also might guess that both can be exculded with the single option: `-x "*/\*.DS_Store"`

I would assume that the other compression commands in the make file have a similar option that could be added to them (though I don't have time to research that right now).

rpspringuel:
Okay after some more research I have two possible fixes:

1. Add ``dist-hook: rm -f `find $(distdir) -name ".DS_Store"` `` to the top level `Makefile.am`.  This hook is executed after the `distdir` is built, but before it is turned into the archive and should remove all files named `.DS_Store` inside the `distdir`.
2. Modify `Makefile.am` in `contrib` so that `TeXShop` is listed as a `SUBDIRS` entry and then add a `Makefile.am` to the `TeXShop` directory which lists the contents of said directory in it's own `EXTRA_DIST` variable.  This should prevent any `.DS_Store` files from being added to the `distdir` in the first place.

Do you have a preference as to which solution is used?

Also, it should be noted that the only `.DS_Store` file that I can find when I build the distribution tarball on my machine is in the `contrib/TeXShop` directory.  Further, I can find no `._.DS_Store` files.

henryso:
I prefer the latter, but whatever works, works.

eschwab:
Fixed by #1073

PR #1073, opened 2016-04-15
Feature/manifest

rpspringuel:
This implements the manifest solution to #1072.


rpspringuel:
#1074 shows the other solution.  It does not conflict with this one, but only one is needed to solve #1072.  Merge the preferred solution and close the other.

eroux:
I think my preference goes to this one

PR #1076, opened 2016-04-16
Renamed parmesan to granapadano.

### LINKED ISSUE #1075: renaming parmesan into granapadano
Packaging GregorioTeX for ctan, I realize that keeping the name parmesan is not the best idea I ever had, as it can only cause problems with original parmesan from Lilypond...  So the `parmesan.*` files should be renamed `greparmesan.*`, and maybe a hack should be done so that documents calling the `parmesan` font continues to work. I'm just not sure there is any single document using `parmesan`, so that second part is not mandatory.

henryso:
Fixes #1075.

No tests change.


contrib/system-setup.bat
```diff
@@ -113,8 +113,8 @@ greextra.ttf ^
 gregorio.ttf ^
 gregorio-op.ttf ^
 gresym.ttf ^
-parmesan.ttf ^
-parmesan-op.ttf ^
+granapadano.ttf ^
+granapadano-op.ttf ^
 gregall.ttf ^
```
rpspringuel: I would leave the parmesan names here and in the system-setup.command below.  That way when it comes to diagnosing problems, we can see if the old file names are returning results (indicating possible version conflicts).  That's why `gregoriotex-ictus.tex` is still in the list.

eroux:
Ok for me, thanks a lot!

henryso:
I added in the deprecated substitution to granapadano when parmesan is selected.  I missed that in the original commit.

tex/gregoriotex-main.tex
```diff
@@ -1654,17 +1654,22 @@
...
-    \gre@loadgregoriofont%
-    \gdef\GreCPVirgaReversaAscendensOnDLine##1{##1}%
-  \else %
-    \xdef\gre@gregoriofontname{#2-#1}%
-    \expandafter\xdef\csname gre@fontfactor@#2-#1\endcsname{#3}%
-    \gre@loadgregoriofont%
-    \gdef\GreCPVirgaReversaAscendensOnDLine##1{\GreCPVirgaReversaLongqueueAscendens}%
-  \fi %
+  \IfStrEq{#2}{parmesan}{%%% DEPRECATED for removal in 5.0
+    \gre@deprecated{parmesan font}{granapadano font}%%% DEPRECATED for removal in 5.0
```
rpspringuel: I might add quotes around `parmesan` and `granapadano` in order to make it clear to the user what the appropriate substitution is.

PR #1077, opened 2016-04-17
new ctan target to Makefile

### LINKED ISSUE #1071: new "ctan" target to make or install-gtex.sh
There should be a "ctan" target to make, creating a `gregorio-x.y.z-ctan.zip` containing:

- gregorio-x.y.z-bin.tar.bz2 with all the files of the dist archive, minus:
  - `install-gtex.sh`, `VersionManager.py`, `.editorconfig`, `.gitignore`, `def_check.sh`, `doc_check.sh` and `.indent.pro` files
  - the `tex/`, `doc/`, `fonts/`, `windows/`, `macosx/`, `debian/` and `contrib/` directory (well, it might have been easier to list what should be in there actually)
- `tex/`, `fonts/` and `doc/` directories, with the same content as in the dist archive, minus the `Makefile.*`
- the `*.md` files
- a `gregorio-x.y.z.tds.zip` as produced by `./install-gtex.sh tds`

eroux:
After quite a few discussions, this is how I got the update zipball on [CTAN](https://www.ctan.org/pkg/gregoriotex).

This also slightly changes the place where a few files are installed, so maybe there are a few more things to adapt?

Fixes #1071

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 # Copyright (C) 2006-2015 The Gregorio Project (see CONTRIBUTORS.md)
-# 
+#
 # This file is part of Gregorio.
 #
 # Gregorio is free software: you can redistribute it and/or modify
@@ -28,3 +28,9 @@ make-fonts:
 	make -C fonts fonts
 
 export TOP_LEVEL_MAKE = true
+
+gregoriotex.tds.zip: install-gtex.sh
+	./install-gtex.sh tds
+
+ctan: dist install-gtex.sh
+	./build-ctan.sh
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -28,3 +28,9 @@ make-fonts:
 	make -C fonts fonts
 
 export TOP_LEVEL_MAKE = true
+
+gregoriotex.tds.zip: install-gtex.sh
+	./install-gtex.sh tds
+
+ctan: dist gregoriotex.tds.zip
```
henryso: This should actually also depend on build-ctan.sh

eroux:
The package made it to [ctan](https://www.ctan.org/pkg/gregoriotex), so it should be safe to merge

PR #1085, opened 2016-05-02
Texlive2016

rpspringuel:
The first step in achieving compatibility with TeXLive 2016.  Still problems with color to work out, but that may have to wait for the xcolor package to get updated.


eroux:
Looks good to me, thanks!

rpspringuel:
Although most of this now appears in the master branch, I've used this branch as an intermediate to reconcile the conflicts created by the 4.1.2 release being created after the 4.2.0 beta.  I'm thus merging it so that those conflicts are taken care of.

PR #1088, opened 2016-05-07
Added a new glyph for ascending puncta inclinata.

### LINKED ISSUE #856: differenciate glyph of punctum inclinatum for ascendens figure
The current output of puncta inclinata ascendens with greciliae is not really satisfactory:

inclinata

the reason is that the shape of the punctum is tuned to look nice in descendens figure. It would be good to be able to tune it for ascendens too.

I think it's not necessary to tune deminutum or ascendens forms as they shouldn't appear in ascendens figures. When there is a kind of ambiguity (only one punctum, etc.), the descendens figure should be taken.

henryso:
Fixes #856.

This obviously changes some tests, and I added a test for the `1` and `0` shaping arguments.


eroux:
Looks perfect to me, thanks a lot!

henryso:
Any objections to merging this?  I'd like to base the build/install script changes on the changes here.

eroux:
ok, let's merge

PR #1091, opened 2016-05-08
Add support for hn and hp pitches to nabc.

jakubjelinek:
As gabc now supports n and p pitches, this adds corresponding change to nabc as well.
While nabc has no staves, higher pitches than m are occassionally useful, plus consistency is also good.
Shall I submit this against develop or release-4.2?


eroux:
Thanks! As the patch is tiny, I think this can go into 4.2, but before it's merged, can you add a changelog entry?

jakubjelinek:
Created #1092 instead then.

PR #1094, opened 2016-05-08
Release 4.1.2

rpspringuel:
This PR cherry-picks just the packaging and TL2016 changes from 4.2.  It's designed to be completely compatible with the 4.1.1 binary so that this can be uploaded to CTAN and create a fully working installation in TL2016 without requiring the TeXLive people to upload and rebuild the binaries.

Proper testing requires you to install the 4.1.1 binary (i.e. do a full install from the `v4.1.1` tag) and then use `install-gtex.sh` to install just the GregorioTeX files from the tip of this PR (use `make ctan` to build all the necessary files).  You also, naturally, need to be on the matching branch in gregorio-test.


rpspringuel:
There are a handful of tests which do not currently pass under TL2016 after these changes.  However, these can be made to pass by the following work arounds:

1. To solve the `Undefined color` issue, modify `gregoriotex.sty` to remove the optional argument `[named]` from the `\color` command in `\gresetlinecolor`.
2. Instead of using the `geometry` package to change to `a4paper`, use the document class option instead (about 18 tests need this change).

With these two changes, all tests pass.  The second is a temporary change that will no longer be needed once the `geometry` package gets updated.  The first should be in the same boat, but there's been no confirmation that the `xcolor` package will be updated.  Since I can get around it with a simple change in our own package, should I make that change?

henryso:
For workaround 1, besides working around the `Undefined color` issue, what other effects does it have?  I'm inclined to let the `xcolor` folks fix it, but you know more about these things than I do.

For workaround 2, since the change is in the test rather than in gregorio, I think we can just make them if this is not fixed quickly enough.

rpspringuel:
> For workaround 1, besides working around the Undefined color issue, what other effects does it have? 

Well, since the `\color` command no longer specifies the `named` pseudo model, one should be able to use any color expression that `xcolor` recognizes instead of just colors which have previously been defined with `\definecolor`.  I had previously forced the use of the two step process (define a color name and then use it) in order to distinguish keep color definition problems separate from GregorioTeX problems.  I suppose, then, that this bug is showing that the exact opposite is happening.

rpspringuel:
The xcolor author got in touch today and sent me a potential fixed version of `xcolor.sty`.  I'm double checking it, but it appears to be working.  I thus expect the xcolor fix to appear in TeXLive 2016 and we won't need to use that workaround.

rpspringuel:
I've completed my checks and can confirm that the patched version of xcolor will fix our problems.

rpspringuel:
Seeing as the last remaining workarounds are test alterations, if there are no objections, then I think this is ready to be merged, 4.1.2 released, and uploaded to CTAN.

henryso:
No objections from me.

eroux:
Sure, let's do that!

PR #1095, opened 2016-05-09
Add 29 new gregall glyphs, replace po-1pp2su1sux1 and po-1su1sux1

jakubjelinek:
glyphs with slightly smaller and nicer looking ones.
If it isn't too late, I'd like also this change in, it just adds 29 new glyphs and replaces 2 (well, actually one plus derived one) with nicer glyph.  I've verified GregorioNabcRef builds fine and manually verified all the changes in there.


henryso:
I don't think it's too late for 4.2.  I also don't feel I can adequately review this pull request other than to suggest adding something to the change log.

eroux:
Same here, once there's a changelog entry (it doesn't have to list all the new glyphs), this can be merged in 4.2

jakubjelinek:
Like this?

eroux:
Perfect, thanks! Merging

Issue #1080, opened 2016-04-27
Skip fonts install option in install-gtex.sh


rpspringuel:
I normally keep my repository very clean by running `make maintainter-clean` after installation.  As a result, the font ttf files are not usually present.  However, this creates a problem when I'm testing changes which only affect the tex files, as `install-gtex.sh` tries to install the font files (which aren't present) and thus fails before running `texhash`.  It would be useful, therefore, if `install-gtex.sh` had an option which would allow one to skip installing the font files.

henryso:
#1097 will allow you to do something like...

```
SKIP=fonts,docs,font-sources ./install-gtex.sh system
```

... to skip installation of fonts, documentation, and font sources.

PR #1097, opened 2016-05-10
Build script changes

henryso:
Modified build.sh to build docs by default.
Added more build targets to the root Makefile.
Added fine-grained install control to install-gtex.sh.
Disabled FORTIFY when building for debug with -O0.

This change requires the corresponding change in gregorio-project/gregorio-test#216.


henryso:
I didn't realize there was an issue filed, but this change also covers #1080.

eroux:

PR #1099, opened 2016-05-10
Fix CHANGELOG: grechangeformat --> grechangestyle

eschwab:
006ad0d5d830cefa7127aa636084cfab01406088 Introduced the name change.

I think the deprecated section of 4.0 is correct with `grechangeformat`.


rpspringuel:
Thanks for catching this.

PR #1101, opened 2016-05-12
Added support for flattened oriscus shapes at ambitus one.

### LINKED ISSUE #1079: oriscus queue length
A request from Solesmes for their font: they want to have a different "queue" length for their oriscus depending on the previous or next note. By "queue" I mean the shape of the oriscus, nothing to do with the stem. Examples:

- the oriscus of `(goi)` would have a long "queue" on the left, and a long queue on the right, merging with the joining line
- the oriscus of `(goh)` would have a long "queue" on the left, and a short queue on the right, because a long queue would merge with the second note and would not be elegant

Until now, this is purely a font problem that can be solved relatively easily. The main problem is with `!` and `@`, as these should make the oriscus use a short queue in case of an ambitus of one. This means duplicating all glyphs starting or ending with an oriscus, and use the long or short form according to what's around. I cannot find another solution though. What do you think?

henryso:
Fixes #1079.

Some tests change, but I had to increase the resolution of the comparison to see any diffs.


henryso:
In order to use this feature for a font, you need to add a `flattened oriscus` key (set to true) to the json configuration of the font.  See `fonts/greciliae.json`.

The following glyphs need to be added to the base font:
- DescendensOriscusFlattened
- AscendensOriscusFlattened
- DescendensOriscusFlatTop
- DescendensOriscusFlatBottom
- AscendensOriscusFlatTop
- AscendensOriscusFlatBottom
- DescendensOriscusLineTLFlatBottom
- DescendensOriscusLineBLFlatBottom
- DescendensOriscusLineTRFlatTop
- DescendensOriscusLineBRFlatTop
- AscendensOriscusLineTLFlatTop
- AscendensOriscusLineBLFlatTop
- AscendensOriscusLineTRFlatBottom
- AscendensOriscusLineBRFlatBottom

The names should be fairly self-explanatory, the new tag at the end is:
- Flattened - flattened on both top and bottom
- FlatTop - flattened on top
- FlatBottom - flattened on the bottom

eroux:
Wow, this really is great, thank you so much! The only change I would do is lower the whole top of the `AscendensOriscusFlatTop` by 6 units (thus decreasing the whole height) so that the link with the note after is even nicer, but otherwise it's perfect!

henryso:
Should all the flattened tops be lowered or just ascendens oriscus?

eroux:
I think the ascendens is enough

henryso:
To me it looks funny if only the ascendens tops are lowered `(h@gof) (f@goh)` looks uneven, so I'm going to do the descendens tops as well.

eroux:
Ok! Maybe it's a good idea to lower the bottom a little bit, so that the height doesn't change too much...

henryso:
I'll push the update so you can take a look to see if the bottom should be changed.  It looks OK to me, but I'm not a good judge.

eroux:
Looks good to me! thanks a lot!

rpspringuel:
All the tests pass for me, so I think this is mergable.

PR #1104, opened 2016-05-14
Update debian packaging

anthonyfok:
First of all, please allow me to express my gratitude to your wonderful work of Gregorio!
It has been a godsend to me and to the communities in the past few years.

After a yearlong break from Gregorio, and failing to get gregorio (2.3-1) to work on Debian "unstable" yesterday, I began trying my hands on refreshing the Debian package.  I am so happy to find Gregorio alive and well, and already at version 4.1.2!

What a coincidence!  It was about the same time that TeX Live 2016 (pre-test) entered into Debian "unstable" and it finally dawned on me that the Debian `texlive-binaries` and `texlive-music` packages already contain `gregorio` and `gregoriotex` respectively.  So, a separate `gregorio` package might not be needed any more, but perhaps we still need it for cutting-edge versions.  :-)

Anyhow, this pull request contains some updates to the Debian packaging in the `debian/` directory.

A diversion (using `dpkg-divert`) is probably needed for `/usr/bin/gregorio` to override the same executable in Debian's `texlive-binaries` (TeX Live 2016) before the package is fit for upload to Debian "unstable".  But, oh well, I will work on it some other day.

Last but not least, thank you so much for bringing Gregorian chant to life!
God bless you all!


debian/gregoriotex.install
```diff
@@ -1,6 +1,6 @@
-tex/gregorio*.tex /usr/share/texmf/tex/luatex/gregoriotex
-tex/gsp-default.tex /usr/share/texmf/tex/luatex/gregoriotex
-tex/gregorio*.lua /usr/share/texmf/tex/luatex/gregoriotex
-tex/gregorio*.sty /usr/share/texmf/tex/luatex/gregoriotex
```
eroux: We recently changed that I believe, see  the new `install-gtex.sh`

eroux:
Thanks a lot for this! I made a small comment but I don't know that much about debian packaging, so I trust you did things well. I think `dpkg-divert` is a good idea indeed. Thanks again!

debian/gregoriotex.install
anthonyfok: Thank you for the reminder!  I have now updated the pull request to use `install-gtex.sh` instead.

henryso:
@rpspringuel Which branch should this go against?  My gut says 4.2, but I'm not sure.

rpspringuel:
I would make this against `relaese-4.2`.  The debian packaging stuff isn't needed by the TeXLive of CTAN folks (in fact it's specifically excluded from the CTAN upload) so there's no need to try and rush these changes in there.  Further, with 4.1.2 appearing in TeXLive 2016 already, this is needed mostly for those looking to update beyond that, and 4.2 will be the next release.

rpspringuel:
Oh, and if pulled against 4.2 then the renaming of the documentation pdfs needs to be taken into account (i.e. they no longer have `-${FILEVERSION}` in the file name).

henryso:
@anthonyfok Do you mind making these changes against the release-4.2 branch instead, taking into account @rpspringuel's notes?

anthonyfok:
@rpspringuel and @henryso: Great idea!  Yes, I would be happy to make these changes against the release-4.2 branch instead, and.... done!  (For the time being anyway.)  Please continue our discussions at PR #1106.  Thanks!

PR #1105, opened 2016-05-14
Update debian packaging for the "release-4.2" branch

anthonyfok:
As discussed in PR #1104, updates to `debian/` shall go into the newer `release-4.2` branch, hence this new pull request.  :-)

Note that a few more changes need to go in before this can be used for uploading to Debian archive, especially the need add diversion to /usr/bin/gregorio so that newer `gregorio` packages can override /usr/bin/gregorio in the TeX Live 2016 `texlive-binaries` package which has just entered Debian sid (unstable) distribution.

----

References:
* https://www.debian.org/doc/debian-policy/ap-pkg-diversions.html
* https://wiki.debian.org/Adding%20and%20removing%20diversions
Other yet-to-add change may include update debian/changelog, though probably unnecessary if the user is using `debian/build-deb.sh`, which adds its own debian/changelog entry on the fly anyway.

----

Speaking of which, `debian/build-deb.sh` fails to complete on the latest Debian sid (unstable) with TeX Live 2016 (pre-test?) while trying to build GregorioRef.pdf, stopping at the following error:

```
! LaTeX Error: Option clash for package xcolor.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.42 \definecolor
               {lightgray}{gray}{0.9}
 406 words of node memory still in use:
   2 hlist, 1 rule, 1 dir, 5 glue, 4 attribute, 50 glue_spec, 4 attribute_list,
 2 write, 1 pdf_colorstack nodes
   avail lists: 2:13,3:2,4:2,5:2,7:3,8:1,9:2
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on GregorioRef.log.
```

which Akira Kakuto reported in http://www.tug.org/pipermail/tex-live/2016-April/038140.html.  Should we add

```tex
\RequirePackage{luatex85}
\RequirePackage[pdftex]{xcolor,graphicx}
```
as Akira suggested, or should we just wait xcolor is updated?

Many thanks!


anthonyfok:
Mea culpa!  Mea culpa!  Mea maxima culpa!  I have based the pull request against the wrong branch!  Will do it again.  Sorry!

Issue #1110, opened 2016-05-17
easing justified text at end of final line


eroux:
The following example:

```tex
\documentclass[11pt]{article} %%
\usepackage{fontspec} %%
\usepackage{gregoriotex} %%
\gresetlastline{justified}
\begin{document} %%

a\label{reprise}\par

\newcommand{\Ruber}[1]{\textit{\textcolor{gregoriocolor}{#1}}}

\gabcsnippet{%
(c4) Com(g)mu(h)ni(h)cán(h)tes,(h) et(h) di(h)em(h) sa(h)cra(h)tís(h)si(h)mum(h) ce(h)le(h)brán(h)tes,(h) (,) quo(h) U(h)ni(h)gé(h)ni(h)tus(h)  +(;) (g+) ([ev:\hfil]) <v>\Ruber{Sequitur p. \pageref{reprise}.}</v>{}()%
}%

\end{document}
```

![bug](https://cloud.githubusercontent.com/assets/60868/15318632/37bfdf80-1c27-11e6-935f-0876539b8821.png)

in which the right alignment is not correct.

rpspringuel:
That would be a bug in the end of line shifts.  `\greseteolshifts{disable}` makes the problem disappear.  Further, giving the last "syllable" a note, also fixes the problem, so it is particular to lines which end with a `\GreNoNoteSyllable`.  It's too late tonight, but I'll look at the possibility of killing the eolshift in these cases tomorrow.

rpspringuel:
I have a fix, but it's revealing a potential problem with the clef change discretionary.  Take a look at `/gabc-output/glyphs/clef_change`.  If you compile the gabc file, you'll see that the clef change which occurs between `test4` and `test5` puts an empty `BarSyllable` at the end of the line (after `test4`).  This prevents the lyrics in the previous syllable (i.e. `test4`) from sliding over under the end of line custos.  My fix accentuates this, pushing `test4` even further away from the end of the line.  This doesn't seem to be desirable to me, but I think the problem is with the underlying behavior: what is put in the discretionary.  Should that empty `BarSyllable` be there?  Shouldn't there be some recognition that there isn't anything to be printed and the before break argument be left blank?

PR #1116, opened 2016-05-22
Eolateos

rpspringuel:
This fixes the problem indicated in #1110, but makes another problem more apparent.

### DIFF
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -1205,7 +1205,9 @@
       \GreNoBreak%
     \fi%
     \ifgre@eolshiftsenabled%
-      \ifgre@endofscore\else%
+      \ifgre@endofscore%
+        \kern\glueexpr(-\gre@skip@bar@lastskip)\relax%
+      \else%
         %this shift will shorten the line somewhat to allow the lyrics to extend under the custos should TeX decide to insert a line break after the syllable being ended
         \gre@calculate@eolshift{\dimexpr(\gre@dimen@enddifference-\gre@skip@bar@lastskip)\relax }%
         \gre@debugmsg{barspacing}{End of line shift: \the\gre@dimen@eolshift}%
--- a/tex/gregoriotex-syllable.tex
+++ b/tex/gregoriotex-syllable.tex
@@ -1091,7 +1091,7 @@
       \GreNoBreak%
     \fi%
     \ifgre@eolshiftsenabled%
-      \ifdim\wd\gre@box@syllablenotes>0pt\relax%
+      \ifgre@endofscore\else%
         %this shift will shorten the line somewhat to allow the lyrics to extend under the custos should TeX decide to insert a line break after the syllable being ended
         \gre@calculate@eolshift{\dimexpr(\gre@dimen@enddifference-\gre@skip@bar@lastskip)\relax }%
         \gre@debugmsg{barspacing}{End of line shift: \the\gre@dimen@eolshift}%
@@ -1099,8 +1099,6 @@
           \gre@dimen@eolshift=\gre@skip@bar@lastskip\relax %
         \fi %
         \kern\glueexpr(-\gre@dimen@eolshift)\relax%
-      \else%
-        \gre@dimen@eolshift=0pt%
       \fi%
     \fi%
     % gre@count@temp@one is "forced end of line or last of score"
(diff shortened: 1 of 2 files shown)


rpspringuel:
See https://github.com/gregorio-project/gregorio-test/pull/220/commits/1c0530479f8e6d54f70330f57183535e2ac33628 for the problematic test.  You'll know that `test4` has been pushed well away from the end of the line.  This is because there is an empty syllable between it and the end of the line.

rpspringuel:
After doing some further testing, I'm unable to collapse the space more than a nominal amount by removing the empty syllable.  As a result, I think any fix for that issue is going to require significantly more work and should be tracked separately.  If others agree, then this is ready to merge.

eroux:
Ok for me to merge this and open another issue for the empty syllables in discretionaries (corresponding to the wrong test)

tex/gregoriotex-syllable.tex
```diff
@@ -1091,13 +1091,17 @@
...
     \fi%
     \ifgre@eolshiftsenabled%
-      %this shift will shorten the line somewhat to allow the lyrics to extend under the custos should TeX decide to insert a line break after the syllable being ended
-      \gre@calculate@eolshift{\dimexpr(\gre@dimen@enddifference-\gre@skip@bar@lastskip)\relax }%
-      \gre@debugmsg{barspacing}{End of line shift: \the\gre@dimen@eolshift}%
-      \ifdim\gre@dimen@eolshift<\gre@skip@bar@lastskip %
-        \gre@dimen@eolshift=\gre@skip@bar@lastskip\relax %
-      \fi %
-      \kern\glueexpr(-\gre@dimen@eolshift)\relax%
+      \ifdim\wd\gre@box@syllablenotes>0pt\relax%
```
eroux: Well, if the change is about the final syllable, what about replacing it with a check for the final syllable?

rpspringuel: I tried using `\ifgre@endofscore` first, but didn't get any results.  Using this test did.

Spurred by your comment, I just did some further testing (taking the sample and turing it into a gabc file instead of a snippet, change pushed to the test PR) and the reason that using `\ifgre@endofscore` doesn't work is because `\GreLastOfScore` (which sets that flag to true) is never called.  If some one could look into why the command line tool is not inserting that call, I'd appreciate it.

henryso: Which test?  If you're talking about gregorio-project/gregorio-test@1c05304, it looks to me like `\GreLastOfScore` is being generated for that test.

henryso: Never mind.  I re-read your comment and will see if I can figure out what's going on.

henryso: It's happening because the last "element" in the score is a texverb rather than a real element, and the code only looks one ahead.  What is the correct behavior?

1. Put `\GreLastOfScore` on the last real element, or
2. Put `\GreLastOfScore` on the last "thing" of the score, regardless of what it is, or
3. Something more complex than that (please explain).

eroux: Very good question... I think option 1 makes sense.

rpspringuel: @henryso, can you define what you mean be "real element"?

eroux: I think it's "not a texverb"

henryso: A "real element" is something that has (or could have) notes in it.  Examples of things that are not "real elements": spaces `(/)`, texverb `([ev:...])`, no-line-break area `(<nlba>)`, above lines text `([alt:...])`, clef `(c4)`, bar `(::)`, end-of-line `(Z)`.

rpspringuel: Okay then, I agree with @eroux, option 1 makes sense.

henryso: I will try to work on this tonight and open a pull request against @rpspringuel's repository.

eroux: great, thanks!

henryso: The fix is ready.  It fixes some problems in our test suite that I suppose we never noticed.

rpspringuel:
Okay, I've merged in @henryso's change and now am running into another problem.  With this change in place, if the score ends with a bar line and the last line should be justified, the bar line doesn't get pushed all the way to the right hand side of the staff.  I'm going to have to work out what is causing this and how to get around it, but don't have the time to do that right now.

rpspringuel:
I think I've fixed the problem.  Please review and merge if acceptable.

henryso:
Looks good to me.

PR #1119, opened 2016-05-25
Rearranged syllable text to improve ligaturing

### LINKED ISSUE #1098: ligatures and syllabation
The word *efficiamur* is usually syllabized (is that English?) as `ef-fi-ci-a-mur`, but with fonts containing an `ffi` ligature, the "gabcization" depends on the context:

* if the notes are small enough not to add hyphens between the first two syllables, it should be `e()ffi()ci()a()mur()` in order to get the `ffi` ligature correctly
* if an hyphen must be put between the first two syllables, it should be `ef()fi()ci()a()mur()`

I have no clue on the best way to handle this in a simple way... do you?

henryso:
Fixes #1098.

This is drawn against the `fix-1098` branch in the main repository, so feel fry to fix any issues you find.

After all these changes, no test outputs changed, but I did add an inherited dimension to one without changing its output while testing the hashing algorithm against spaces.


eroux:
Thanks a lot! That's a very impressive change from a technical point of view! And many scores will look much better thanks to that! The only thing I think would be necessary before merging would be a mechanism to make this behavior optional, so that if there's a bug somewhere (with document stability for instance), users can disable this behavior

henryso:
Should this be on by default or off?

eroux:
We can turn it on by default

henryso:
I implemented the toggle.  In testing it, I realized that I had accidentally deleted one line which broke the entire algorithm, so that line is now restored.  This breaks one test which moves some letters by a minuscule amount.

eroux:
Thanks! I think this is ok to be merged, we'll see how users react to beta2!

rpspringuel:
I can confirm a full set of tests passing.

PR #1120, opened 2016-05-25
Defining small caps for PlainTeX

rpspringuel:


henryso:
I wonder if it might just be better to issue a warning or an error if `\sc` is not defined?

rpspringuel:
That would be more in line with what we do for `\GreUnderline`.  However, we would then need to modify the defaults styling for modes, which currently use `\sc` (and thus raise an `Undefined macro` error).

henryso:
Do you think it would be too incoherent to have that default be different in PlainTeX ?

rpspringuel:
It would mean that mode numbers are lowercase roman numerals by default.  Personally I think that looks odd, but it wouldn't be terribly incoherent.

henryso:
How about using roman-majuscule as the default for the mode number system in plain TeX, then?

rpspringuel:
That makes some sense.  I'll see what I can do about implementing that.

rpspringuel:
Changes implemented.  No tests change.  Does anyone think I need to add tests related to this?

henryso:
Probably a single test with a mode header is good enough, but I think a test is a good idea.

henryso:
GregorioRef should be updated for this change, where it describes the default mode stye and where it describes the mode number system.

rpspringuel:
Simple tests added.  Let me know if that's not enough.

henryso:
I would add a gabc with mode header to those tests, exercising the difference in mode number system.

rpspringuel:
Done.

henryso:
Looks good.

Issue #1038, opened 2016-03-16
setting parskip, lineskip, baselineskip and lineskiplimit for scores


eroux:
I didn't inspect that with enough care yet, but the way I thought vertical spaces were set was thought `spaceabovelines`. So I thought the stretching was done here. The problem is that this space is set in a box, so it's basically never stretched. Also, the way it interferes with the different lineskips of TeX can certainly lead to some uncoherent vertical spaces. So there should be a way to do it properly by setting

* `\parskip`
* `\lineskip`
* `\baselineskip`
* `\lineskiplimit`

for scores and not mislead users into thinking `spaceabovelines` may solve their problems.

PR #1129, opened 2016-05-28
Score specific values for line spacing parameters

rpspringuel:
Addresses #1038.

Default values have been chosen based on a 12pt article.  Let me know if you want these tweaked.

### DIFF
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -47,6 +47,10 @@
 \newskip\gre@emergencystretchsave%
 \newcount\gre@widowpenaltysave%
 \newcount\gre@clubpenaltysave%
+\newskip\gre@skip@parskip@save%
+\newdimen\gre@dimen@lineskip@save%
+\newdimen\gre@dimen@baselineskip@save%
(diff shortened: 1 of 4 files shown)


henryso:
I'm assuming no tests changed?

tex/gregoriotex-spaces.tex
```diff
@@ -1457,6 +1473,9 @@
     {spaceporrectuspunctummorashift}{\gre@rubberfalse}%
     {raresignshift}{\gre@rubberfalse}%
     {divisiofinalissep}{\gre@rubberfalse}%
+    {lineskip}{\gre@rubberfalse}%
```
eroux: Why using non-stretchable dimen instead of skips?

rpspringuel: The default values for those three led me to believe they were dimensions, but after a bit more research I realize they are skips whose stretch and shrink are 0 by default.  I'll fix that.

rpspringuel:
@henryso no tests change.

rpspringuel:
So I've modified a test to show what increasing the parameters does, though I make no claim that said modification attempts to be systematic about it.

PR #1130, opened 2016-05-29
Release 4.1

rpspringuel:
Missed an occurrence of the bug. This gets the custos to reappear.

### DIFF
--- a/.gregorio-version
+++ b/.gregorio-version
@@ -1,4 +1,4 @@
-4.1.3
+4.1.4
 
 *** Do not modify this file. ***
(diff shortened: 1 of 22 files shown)


rpspringuel:
Note, changes in 4.2 make the code changes in this PR unnecessary on the 4.2 branch, though the CHANGELOG needs to be updated.  I've already taken care of that.  We're going to have to watch the merges to make sure we don't have code reversions when 4.2 is merged into master.

henryso:
I merged master into release-4.2, so hopefully there should be no problem merging it back.

PR #1151, opened 2016-06-16
fix #1149

### LINKED ISSUE #1149: bars sometimes appearing at the beginning of lines
This one is super difficult to reproduce, I don't manage to isolate it in a small example, but I have one complex example so I'll fix it

eroux:
same here, no time to run the tests


henryso:
None of the tests change from this fix, but as per #1149, this is difficult to reproduce.  Merging.

Issue #1153, opened 2016-06-21
Nabc vertical position should not depend on the pitch of the gabc notes


jakubjelinek:
As has been reported on gregorio-users, when a score line contains l or higher notes, the <alt> baseline seems to be moved above (verified already in 4.1), which is not bad for <alt>, but undesirable for nabc, which has a way of positioning the neumes vertically.
Wonder if it is \gre@dimen@additionaltopspacealt and \gre@space@count@additionaltopspacealtthreshold that matters in this case.  If yes, wonder if we could not then duplicate/modify the \GreSetTextAboveLines and \def\gre@typesettextabovelines macros as well as add additionaltopspacenabc* variants of the above (use Nabc or nabc instead of Text and text), so that it could be configurable differently for nabc.  And by default use higher treshold.
Could this still be done for 4.2?

henryso:
Copied from gregorio-project/gregorio-test#242, the following changes to the documentation are necessary:

- `additionaltopspacenabcthreshold` should be added as a new count to Command_Index_User.tex
- `\gre@typesettextabovelines` should be updated (to document the `#2` argument) in Command_Index_internal.tex
- `\GreSetNabcAboveLines` should be added to Command_Index_gregorio.tex

PR #1154, opened 2016-06-21
Copy'n'paste approach for issue #1153.

jakubjelinek:
I've actually tested only 4.1 backport of this change so far.
The commit also includes one pasto change -
\gre@space@count@additionaltopspacethreshold=2
...
\gre@space@count@additionaltopspacethreshold=0
where the latter IMHO should have been
\gre@space@count@additionaltopspacealtthreshold=0

I've also noticed the Copyright years weren't updated this year, shall that be done before 4.2 is released?

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -815,6 +815,35 @@
   \relax %
 }%
 
+% similarly, but for typesetting Nabc neumes above lines
+
+\def\GreSetNabcAboveLines#1{%
+  \gdef\gre@currenttextabovelines{%
+    \gre@typesetnabcabovelines{#1}%
+    \gdef\gre@currenttextabovelines{}%
+    \relax %
+  }%
+}%
+
+% typesets the nabc above the line
+\def\gre@typesetnabcabovelines#1{%
+  \gre@style@abovelinestext%
+  \gre@debugmsg{spacing}{Raise nabc: \gre@space@dimen@abovelinestextraise}%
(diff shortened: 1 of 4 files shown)


jakubjelinek:
Thinking about it more, I believe we must be also measuring the maximum height of the alt text for inter-staff spacing, with different baselines for alt text and nabc neumes I bet we also need to measure separately the maximum heights of the alt text vs. maximum height of nabc neumes and use the corresponding baseline for those.  Or is that not the case?
As for the copy'n'paste, I think only the \gre@typesetnabcabovelines macro is large enough that it would be perhaps useful to use just one macro with an extra argument instead.

henryso:
The calculation (not that I understand it completely) that you speak of is `\gre@calculate@additionalspaces` in `gregoriotex-spaces.tex`.  I think it's being put into `\gre@dimen@additionaltopspacealt`.

jakubjelinek:
`\gre@dimen@additionaltopspacealt` I already replace with `\gre@dimen@additionaltopspacenabc`
for \GreNabcAboveLines.  I've tried Sr. Maria's testcase copied 3 times and the spacing between the top of the nabc neumes and previous line of text is the same when \GreNabcAboveLines and \GreTextAboveLines is used for nabc, so I think with the cleanup change I've just pushed it should DTRT and not be too ugly anymore (but I'm never sure about \relax and % at the end of lines).

eroux:
This looks ok to me

henryso:
I'll merge this and review tests some time today.

PR #1158, opened 2016-06-23
fix #1152

### LINKED ISSUE #1152: end of score introducing wrong penalties
With the TeX file from #1139 and the following gabc file:

```
name: Urbs fortitudinis;
%%
(c3) a(i)pe(i)rí(j)te(ihz) por(i>)tas,(g_f_3!gwh) (,) qui(h)a(hhog) no(f)bís(g_h)cum(f) De(fe)us,(d) al(ef)le(f)lú(e.)ia.(e.) (::) <eu>E(i) u(i) o(j) u(i) a(ih!/gggggggggggggggggggggg) e(iio>) </eu>(::)
```

the last note is alone on a line. It should be completely impossible. The problem is that the `.gtex` contains `\GreLastOfScore` at the beginning of the last syllable, so in the `\gre@endofglyphcommon` introduces a `\gre@space@count@finalpenalty` which allows a break.

eroux:
This should fix #1152. I'm not 100% sure of the change but it doesn't seem to break anything and fixes the bug... Please review


henryso:
This seems to be the correct solution to me.

PR #1160, opened 2016-06-23
Bump Copyright year.

jakubjelinek:
Dunno what exact rules we want to use, either Copyright years are bumped only when the file is touched first time during year, but then we might need 2014-2015, 2017-2018 etc. separate ranges, or what various projects due is in January all copyright years are updated (e.g. GCC has a python script to do that) and only single ranges are used from the year the file has been touched first till the year the copyright has been bumped last time.


henryso:
Thanks.

PR #1161, opened 2016-06-24
Refactored hyphen macros.

### LINKED ISSUE #1140: calling \GreHyph with an argument
It would be useful sometimes to be able to reduce the size of the hyphen, but depending on the space between the syllables: having a long hyphen when the space is very big, and a shorter one when the space can be shrinked. I think the best way to do that is by
- calling `\GreHyph` more consistently when inserting an hyphen (sometimes `-` is used)
- calling `\GreHyph` with an argument (ignored by default?) being the width between the two syllables

henryso:
Fixes #1140.

No tests change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Staff lines after a two-line initial should now be sized correctly in a more consistent way (see [#1141](https://github.com/gregorio-project/gregorio/issues/1141)).
 - Space at the end of line was not always correct when a line is cut in the middle of a word (see [#1155](https://github.com/gregorio-project/gregorio/issues/1155)).
 - In rare cases, the very last bar or glyph of a score could appear alone at the beginning of the final line (see [#1152](https://github.com/gregorio-project/gregorio/issues/1152)).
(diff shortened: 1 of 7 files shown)


eroux:
Thanks, I think it's a good basis! But the main point for this feature is to allow the user to change the hyphen (font+char) according to the space the hyphen has... So I think it would be helpful to have a high level macro doing this (redefining `\gre@hyphen` in fact) a difficulty is to allow the user to change the hyphen automatically added in the Lua pass... Do you see what I mean? (I'm not sure it's 100% clear in my head)

eroux:
In fact, in an ideal world, the second argument of `\gre@hyphen` would not be ignored in the case of hyphens added by Lua at the and of a line, it would be `enddifference+width of (custos+space)`... That requires the eolshift code to tell Lua which hyphen should be used though, but that could be done with an attribute... Wdyt?

henryso:
I honestly don't know enough about this stuff to comment in any meaningful way.  Let me know what I should do and I'll try to do it.

eroux:
Hmmm... it's a bit blur in my mind too... maybe I can just continue the PR ? I have a few ideas on how to do what I think would be best...

henryso:
Ok, I will pull this change into its own branch.

PR #1163, opened 2016-06-25
Refactored hyphen macros.

### LINKED ISSUE #1140: calling \GreHyph with an argument

henryso:
Fixes #1140.


henryso:
@eroux The fix-1140 branch is ready for you.

PR #1166, opened 2016-06-27
Fix 1138

### LINKED ISSUE #1138: spaceafterlineclef@close
It would be useful to differenciate `spaceafterlineclef` when the notes are "close" to the key or not, for instance in `(c4) (a)` the note is far, and in `(c3) (h)` the note is closed. There are a few corner cases (the second note should be taken in the case of a pes), but it can be made in a very simple way.

eroux:
Fixes #1138.

### DIFF
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -4034,18 +4034,35 @@ static int first_note_near_clef(const gregorio_score *const score) {
         clef = score->first_voice_info->initial_clef;
         if (!clef.secondary_line && !clef.flatted && score->first_syllable
                 && score->first_syllable->elements) {
-            gregorio_element *element = score->first_syllable->elements[0];
+            const gregorio_element *element = score->first_syllable->elements[0];
             if (element && element->type == GRE_ELEMENT) {
-                gregorio_glyph *glyph = element->u.first_glyph;
+                const gregorio_glyph *glyph = element->u.first_glyph;
                 if (glyph && glyph->type == GRE_GLYPH) {
-                    signed char clef_pitch = LOW_LINE_PITCH
+                    const signed char clef_pitch = LOW_LINE_PITCH
                         + ((clef.line - 1) * 2);
-                    gregorio_note *note = glyph->u.notes.first_note;
-                    if (glyph->u.notes.glyph_type == G_PODATUS && note->next) {
-                        note = note->next;
+                    const gregorio_note *low_note = glyph->u.notes.first_note;
+                    const gregorio_note *high_note = low_note;
+                    switch (glyph->u.notes.glyph_type) {
+                    case G_PODATUS:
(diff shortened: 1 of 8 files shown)


src/gregoriotex/gregoriotex-write.c
```diff
@@ -4028,6 +4028,33 @@ static void suppress_expansion(FILE *const f, const char *text)
...
+            if (element && element->type == GRE_ELEMENT) {
+                gregorio_glyph *glyph = element->u.first_glyph;
+                if (glyph && glyph->type == GRE_GLYPH) {
+                    signed char clef_pitch = LOW_LINE_PITCH
+                        + ((clef.line - 1) * 2);
+                    gregorio_note *note = glyph->u.notes.first_note;
+                    if (glyph->u.notes.glyph_type == G_PODATUS && note->next) {
+                        note = note->next;
+                    }
+                    if (!(note->u.note.pitch > clef_pitch - 4
```
henryso: I think it would be better to change this to:
```
if (note->u.note.pitch < clef_pitch - 3
        || note->u.note.pitch > clef_pitch + 3) {
```

eroux: Indeed. I wonder if something could also be checked for high keys and above line things...

src/gregoriotex/gregoriotex-write.c
```diff
@@ -4028,6 +4028,33 @@ static void suppress_expansion(FILE *const f, const char *text)
...
+        if (!clef.secondary_line && !clef.flatted && score->first_syllable
+                && score->first_syllable->elements) {
+            gregorio_element *element = score->first_syllable->elements[0];
+            if (element && element->type == GRE_ELEMENT) {
+                gregorio_glyph *glyph = element->u.first_glyph;
+                if (glyph && glyph->type == GRE_GLYPH) {
+                    signed char clef_pitch = LOW_LINE_PITCH
+                        + ((clef.line - 1) * 2);
+                    gregorio_note *note = glyph->u.notes.first_note;
+                    if (glyph->u.notes.glyph_type == G_PODATUS && note->next) {
```
henryso: This probably needs to check *both* notes of the podatus, as well as both notes of any flexus deminutus.

eroux: correct yes

eroux: don't hesitate to apply these changes directly, I won't have time today...

eroux: In fact I think both notes of clivis should be checked whatever the liquescentia, because of the stem...

henryso: Perhaps the first two notes of a porrectus then, for the same reason?  Anything else we're missing?

eroux:
Looks good to me... do you know if it changed any test?

eroux:
no test changed

PR #1172, opened 2016-07-03
Added the ability to change a dimension for a given line in a score.

### LINKED ISSUE #1156: tuning `spacelinestext` on a per-line basis
In some rare cases, `spacelinestext` (including the adjustments gregorio makes) are not optimal, and a tuning must be done by hand, but I don't think it is possible with the current state...  This would be needed in case where, for instance, a very low note meets a very high letter (with an accent).

Looking at [this thread](https://mail.gna.org/public/gregorio-users/2016-06/msg00033.html) I think the same should be done for the other vertical spaces such as `abovelinestextraise` or the different future nabc heights.

My feeling is that a macro could just store these user values in a lua table, emptied at each end of score, and these could be used in the lua second pass. From this very general point of view it seems easy so it could fit into 4.2, but if it's more complex, this could go into 5.0.

henryso:
Fixes #1156.

No tests change, but I added one to exercise the new feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - `moraadjustment` and `moraadjustmentbar` did not scale when changing factor, they do now.
 
 ### Added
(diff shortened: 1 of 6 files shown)


henryso:
I added the spaces used by `\gre@generatelines`.  Did I miss anything?

Issue #844, opened 2016-01-28
best way to represent optional notes?


eroux:
The Antiphonale Monasticum Project would like to represent some optional notes, they're thinking about two solutions:

- extending the cavum system to many other glyphs (virga, pes, clivis, torculus and porrectus)
- using brackets before and after groups of optional notes, where the brackets have different heights and vertical placements according to the notes (algorithm currently unknown!)

Both solutions seem quite complex, especially the first one... What do you think would be best? Do you have other ideas?

henryso:
If we use the cavum system and generate all the glyphs, it *will* overrun the BMP PUA.  Alternately, we can do it with fusion.  A large part of the difficulty in either case, is drawing the base glyphs.

Depending on the algorithm, the bracket solution could also be fairly complex, but the difficulty here is really the placement algorithm and drawing the brackets (metapost?).

As far as which is better, this is clearly a style decision, and given there's no precedent, I don't think we can really make an appropriate decision.  Perhaps both should be implemented in the long term, and of course the project-selected solution in the short term.

eroux:
For cavum glyphs, drawing all possible glyphs would indeed be a rather bad decicion... I was thinking about only drawing the cavum part and place it as signs, I think it could be plugged (not without work of course) into current architecture?

I agree that both solutions would be best in the long term. The Antiphonale Monasticum project is not really decided yet, but they seem to prefer the cavum option...

henryso:
Placing the cavum on the note head doesn't sound too difficult.  Placing it on the downward stroke of the porrectus will be somewhat more complex.

eroux:
I admit I don't really see why: the sign placement already has so many options, I think it would be relatively easy, for instance, to draw a vertical episemus in the middle of the porrectus stroke? Wouldn't this be basically the same?

henryso:
For the porrectus stroke, the complication is vertical positioning since it's not just below the pitch or something like that; it's probably best handled by creating one glyph per porrectus stroke (I think we have ten of them), and then some glue logic in the cases to select the right glyph depending on stroke/case.

henryso:
One more thing to think about: there is an option to see the lines behind the "hole".  Drawing a white glyph on top of a black one will not allow this to happen.

eroux:
[...]

PR #1175, opened 2016-07-04
Added support for square brackets around notes.

henryso:
Part of the implementation for #844.

No tests change, but I added a new one to exercise the new feature.


eroux:
Wonderful, thanks a lot! The only thing I would change is to put the extremities of the brackets further from the staff lines (closer to the middle of the lines)... What do you think? (it must be changed in the font right?)

henryso:
I don't think it's as simple as a font change, but I'll try.

eroux:
Ok

henryso:
It looks like we'll need three:

- a short bracket for typical cases where the extrema are both on a space
- a long bracket for typical cases where the extrema are both on a line
- the current size bracket for cases where one extremum is on a space and one is on a line (but shifted up or down accordingly); this case would also handle weird cases below and above the staff.

This is turning out to be a bit complex, so it may take me some time.

henryso:
I made the changes described above.  Please review.

eroux:
I think this is really perfect, thank you very much!

Issue #1182, opened 2016-07-10
changing default spaces


eroux:
Many of the current spaces have not been carefully chosen (by me), and I think changing them is only possible for quite big projects as it requires to understand how the spacing work, which is not obvious. So having better default spaces should be quite beneficial to many. The Antiphonale Monasticum projects allowed us to take the spaces they use, I'll make a pull request and we can discuss them

eroux:
Here are the spaces used by the italian Antiphonale Project:

```tex
\grechangedim{annotationraise}{-2mm}{scalable}
\grechangedim{beforeinitialshift}{2mm}{scalable}%
\grechangedim{afterinitialshift}{2mm}{scalable}%
\greseteolhyphen{normal}

\grechangedim{moraadjustment}{0.05 cm}{scalable}%
\grechangedim{moraadjustmentbar}{0.05 cm}{scalable}%
%\grechangedim{alterationadjustmentbar}{0.07 cm}{scalable}%
\grechangedim{bar@virgula@standalone@notext}{0.2 cm}{scalable}%
\grechangedim{bar@minima@standalone@notext}{0.2 cm}{scalable}%
\grechangedim{bar@minor@standalone@notext}{0.2323 cm}{scalable}%
\grechangedim{bar@dominican@standalone@notext}{0.2323 cm}{scalable}%
\grechangedim{bar@maior@standalone@notext}{0.2323 cm}{scalable}%
\grechangedim{bar@finalis@standalone@notext}{0.2323 cm}{scalable}%
\grechangedim{bar@finalfinalis@standalone@notext}{0.29169 cm}{scalable}%
\grechangedim{bar@rubber}{0 cm plus 0.025cm minus 0.025 cm}{scalable}%
\grechangedim{bar@minima@short}{0.12 cm plus 0.05 cm minus 0.00469 cm}{scalable}%
\grechangedim{bar@virgula@short}{0.13 cm plus 0.05 cm minus 0.00469 cm}{scalable}%
%\grechangedim{bar@finalfinalis@standalone@notext}{0.2323 cm}{scalable}%
\grechangedim{interwordspacetext@bars@notext}{0.19 cm}{scalable}%
\grechangedim{interwordspacetext@bars@notext@euouae}{0.18 cm}{scalable}%
\grechangedim{interwordspacetext}{0.17 cm plus 0.05 cm minus 0.05 cm}{scalable}%
[...]

PR #1183, opened 2016-07-10
#1182: new spaces

eroux:
Spaces from the Antiphonale Monasticum project


henryso:
As one who is aesthetically challenged, I cannot really speak to the "beauty" of these spaces, but I will ask if we should really be changing them this late in 4.2.

eroux:
As you prefer, another option is to send them to the list and make them the default only starting from 5.0... what do you think?

henryso:
I would certainly feel more comfortable doing that.

henryso:
@eroux Perhaps we should send this to the list now and make the changes in 5.0?

eroux:
Yes, I'm just waiting a few days for a better version of the spaces

PR #1194, opened 2016-07-28
Copyright notices and installation manifests for documentation

### LINKED ISSUE #1188: Problem with Docs
In working on the installation over TeXLive 2016 issues, I've noticed a problem with the docs which I don't understand.

```
GregorioRef Warning: missing GABC for notdeferror: ...xmf-dist/tex/luatex/luaotfload/fontloader-2016-06-16.lua:15598: attempt to index field 'spacing' (a nil value)
.
<template> ...nd \hskip \z@ plus.5fill\relax \egroup \egroup 
                                                  \begingroup \CT@setup \CT@...

l.54 ...tTable{'greciliae','gregorio','granapadano'}
```

If I switch back to TL2015, then the error goes away, so the error is almost certainly related to TL2016.  Can anyone here further narrow this down so that I can report it in the right place?

henryso:
~~An attempt to fix #1188.~~

~~This works around the font loader issue, allowing the document to be built under TeX Live 2016.~~

~~However, doing it this way causes a segfault under (Gentoo's packaged version of) TeX Live 2015 unless the luatex font cache is cleared between runs of lualatex or if I `\input` the Dominican font table before the "normal" font table.  Since this doesn't happen under 2016, the bug is obviously fixed.~~

~~Is it good enough to abandon building the document under 2015, knowing that we build it anyway for distribution?  If so, then I think this is ready for review/merge.~~

Edit: this pull request now only adds the copyright notices and corrects the installation manifests.

### DIFF
--- a/doc/Appendix_Font_Tables.tex
+++ b/doc/Appendix_Font_Tables.tex
@@ -1,5 +1,23 @@
 % !TEX root = GregorioRef.tex
 % !TEX program = LuaLaTeX+se
+%
+% Copyright (C) 2006-2016 The Gregorio Project (see CONTRIBUTORS.md)
+%
+% This file is part of Gregorio.
+%
+% Gregorio is free software: you can redistribute it and/or modify
+% it under the terms of the GNU General Public License as published by
+% the Free Software Foundation, either version 3 of the License, or
+% (at your option) any later version.
+%
+% Gregorio is distributed in the hope that it will be useful,
+% but WITHOUT ANY WARRANTY; without even the implied warranty of
+% MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+% GNU General Public License for more details.
+%
+% You should have received a copy of the GNU General Public License
+% along with Gregorio.  If not, see <http://www.gnu.org/licenses/>.
+%
 \begin{landscape}
 
 \section{Font Glyph Tables}\label{glyphtable}
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -1,5 +1,23 @@
 % !TEX root = GregorioRef.tex
 % !TEX program = LuaLaTeX+se
+%
+% Copyright (C) 2006-2016 The Gregorio Project (see CONTRIBUTORS.md)
+%
+% This file is part of Gregorio.
+%
+% Gregorio is free software: you can redistribute it and/or modify
+% it under the terms of the GNU General Public License as published by
+% the Free Software Foundation, either version 3 of the License, or
+% (at your option) any later version.
+%
+% Gregorio is distributed in the hope that it will be useful,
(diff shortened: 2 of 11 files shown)


doc/Makefile.am
```diff
@@ -29,6 +30,16 @@ NABCSRCFILES = GregorioNabcRef.tex veni.gabc
 GregorioRef.pdf: $(SRCFILES)
 	$(MAKE) $(AM_MAKEFLAGS) -C ../src gregorio
 	../src/gregorio -o factus.gtex $(<D)/factus.gabc
+	texlua scorefonttable.lua greciliaeR ../fonts/greciliae.ttf \
```
eroux: I'm not sure I can find the `scorefonttable.lua` script, did you forget to add it?

henryso:
Sorry about that.  Added.

eroux:
Thanks! I think a license indication (and a small sentence on what they do) would be good at the top of the new files. Apart from that, maybe a `KNOWN ISSUES` section in the `README.md`  file could also indicate that the documentation cannot be built with TeXLive 2015

henryso:
There's no license indication at the top of any of the other documentation source files.  Should I add one?

eroux:
Oh, sorry I didn't notice that... well, yes, I think it would be best

henryso:
Done.

eroux:
Thanks! Ok for me!

henryso:
@rpspringuel Is this solution good enough for you?

doc/Makefile.am
```diff
@@ -18,7 +18,8 @@
 # GregorioRef.tex must be the first source file
 SRCFILES = GregorioRef.tex Command_Index_gregorio.tex \
 		   Command_Index_internal.tex Command_Index_User.tex \
-		   Gabc.tex Appendix_Font_Tables.tex GregorioRef.lua factus.gabc
+		   Gabc.tex Appendix_Font_Tables.tex GregorioRef.lua factus.gabc \
+		   scorefonttable.lua extrafonttable.lua
 NABCSRCFILES = GregorioNabcRef.tex veni.gabc
```
rpspringuel: There are a couple of other places where the new files may needed to be added to manifests.  Off-hand, I think the installers for Mac and Windows might need them (though, since the new files are lua file, they may be picked up by wild cards).  You should also check `install-gtex.sh`.

rpspringuel:
Other than my comment, this works for me.  I should be able to check the installers later tonight if you're not sure where (or whether) the files need to be added.

henryso:
I made the changes based on your recommendations.  Please check the installers when you have a chance.

henryso:
I rolled back the workaround and re-implemented the copyright notices and manifest changes.  Please review and merge if satisfactory.

eroux:
Ok for me, thanks a lot!

eschwab:
Fine for me.

PR #1195, opened 2016-07-30
fixing #1189

### LINKED ISSUE #1189: first note alignment with c and f clefs
In the following example:

```tex
\usepackage{libertine}
\grechangestaffsize{16}
\pagestyle{empty}
\begin{document}
\gabcsnippet{(f3) Res(f) est(e) ad(f)mi(h)rá(gf)bi(e)lis(f) vir(i)go(h) ve(g)ne(f)rá(hg)bi(f)lis(e) (z0c4) pa(c)rit(e) sed(d) in(h)tác(gfg)ta.(f.) (d+::f3) Si(f)ne(e) vi(f)ri(h) có(gf)i(e)tu,(f) ple(i)na(h) San(h)cto(f) Spí(hg)ri(f)tu,(e) gé(c)ni(e)trix(f) est(h) fac(gfg)ta.(f.) (::)}
\end{document}
```

giving

buggregorio

the first note of the lines is at the same distance of the clefs, but when there is a mix of c and f clefs, the consequence is that they are not aligned. I see a few ways to fix it:

- one is manual: the user calls `\gregoriobolalignonfclefs` and the next score has all beginning of lines aligned as if they were f clefs
- one is automatic (but optional): calling a function `\GreScoreContainsFClef` at the beginning of each score containing an f clef
- one is even more automatic with two pass: the first detects if an actual f clef appears in the score, then the second adjusts things

I think the second one is the best... what do you think? (it should be optional though)

eroux:
This should fix #1189, currently only the TeX part is here

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -1008,6 +1008,26 @@ Macro to determine whether Gregorio\TeX\ should apply a small shift at the begin
   & \texttt{disable} & The shifts are not applied.
 \end{argtable}
 
+\macroname{\textbackslash grebolshiftcleftype}{\{\#1\}}{gregoriotex-spaces.tex}
+Macro to determine how notes should be left aligned in the case where clefs of different widths appear in the same score.
+
+\begin{argtable}
+  \#1 & \texttt{largest} & The notes are aligned as if all clefs had the width of the largest clef (default)\\
+  & \texttt{current} & The notes are aligned on the current clef, which leads to unaligned notes. This was the default of Gregorio < \texttt{5.0}.
+\end{argtable}
+
+\macroname{\textbackslash grelocalbolshiftcleftype}{\{\#1\}}{gregoriotex-spaces.tex}
+Equivalent of \verb=\grebolshiftcleftype= but valid only until the next end of a score, and with more options. This can be used inside gabc for corner cases like different alignment on a score taking two pages.
+
+\begin{argtable}
+  \#1 & \texttt{largest} & same as in \verb=\grebolshiftcleftype=\\
+  & \texttt{current} & idem\\
+  & \texttt{f} & force left alignment of notes as if all clef were f clef\\
+  & \texttt{c} & idem with c clef\\
+  & \texttt{fb} & idem with flatted f clef\\
+  & \texttt{cb} & idem with flatted c clef\\
+\end{argtable}
+
 \macroname{\textbackslash gresetlastline}{\{\#1\}}{gregoriotex-main.tex}
 Macro to determine whether the last line of the score should be justified or not.
 
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -1017,15 +1017,15 @@ Macro to determine how notes should be left aligned in the case where clefs of d
 \end{argtable}
 
 \macroname{\textbackslash grelocalbolshiftcleftype}{\{\#1\}}{gregoriotex-spaces.tex}
-Equivalent of \verb=\grebolshiftcleftype= but valid only until the next end of a score, and with more options. This can be used inside gabc for corner cases like different alignment on a score taking two pages.
+Equivalent of \verb=\grebolshiftcleftype= but valid only until the next end of a score, and with more options. This can be used before a score or even inside a \verb=<v>verbatim</v>= in gabc for corner cases like different alignment on a score taking two pages.
 
 \begin{argtable}
-  \#1 & \texttt{largest} & same as in \verb=\grebolshiftcleftype=\\
-  & \texttt{current} & idem\\
-  & \texttt{f} & force left alignment of notes as if all clef were f clef\\
-  & \texttt{c} & idem with c clef\\
-  & \texttt{fb} & idem with flatted f clef\\
-  & \texttt{cb} & idem with flatted c clef\\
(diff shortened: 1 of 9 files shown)


henryso:
I added the C part.  Running the tests, I think the results look as they should, given the code, but the fix-1192 test shows a weird edge case, where the clef change to a larger clef occurs on the last (in this case, the only) line, adding space to the first clef which really doesn't align it with anything below.  I presume something similar would occur if the clef changes from C to F to C on the same line.

eroux:
Thanks a lot! Indeed the current mechanism doesn't cover the kind of edge cases you're describing, but they should be very rare and easy to fix in gabc. I'll run the tests

henryso:
Perhaps how to fix this edge case in gabc should be documented.  I'll admit I have no idea how you'd do that.

eroux:
Strange: trying to run the tests, I have a lot of message like

```
_: ligne 18 : printf: 0.999984: nombre non valable
```
and all tests fail, although the `diff-*.png` show no visible difference... any idea where this may come from?

henryso:
Must be the test scripts.  I'll look into it... Maybe a locale issue.

henryso:
It's definitely a locale problem.  `compare` is returning 0.999984 and your locale most likely is expecting 0,999984.  I'll have to think about the best way to fix this one.

henryso:
I set LC_ALL=C in gregorio-test.sh.  Let me know if that helps.

eroux:
that solved the problem, thanks!

eroux:
Ok, I think this is ready to review

doc/Command_Index_User.tex
```diff
@@ -1008,6 +1008,26 @@ \subsubsection{End of Line Behavior}
...
+\macroname{\textbackslash grebolshiftcleftype}{\{\#1\}}{gregoriotex-spaces.tex}
+Macro to determine how notes should be left aligned in the case where clefs of different widths appear in the same score.
+
+\begin{argtable}
+  \#1 & \texttt{largest} & The notes are aligned as if all clefs had the width of the largest clef (default)\\
+  & \texttt{current} & The notes are aligned on the current clef, which leads to unaligned notes. This was the default of Gregorio < \texttt{5.0}.
+\end{argtable}
+
+\macroname{\textbackslash grelocalbolshiftcleftype}{\{\#1\}}{gregoriotex-spaces.tex}
+Equivalent of \verb=\grebolshiftcleftype= but valid only until the next end of a score, and with more options. This can be used inside gabc for corner cases like different alignment on a score taking two pages.
```
henryso: Since this is the user documentation, I would explicitly mention that this should be used in a `<v>texverb</v>` in gabc.

doc/Command_Index_User.tex
```diff
@@ -1008,6 +1008,26 @@ \subsubsection{End of Line Behavior}
...
+
+\begin{argtable}
+  \#1 & \texttt{largest} & same as in \verb=\grebolshiftcleftype=\\
```
henryso: Similarly, as this is user documentation, I would repeat the descriptions here rather then refer to another macro.

tex/gregoriotex-signs.tex
```diff
@@ -142,11 +142,12 @@
     ]%
 }%
 
-
 %% marcro to define the clef that will appear at the beginning of the lines
```
henryso: Since you're here, you might as well correct the spelling of macro.

tex/gregoriotex-signs.tex
```diff
@@ -183,18 +193,10 @@
...
-%% #4: 0: no space after, 1: normal space after, 2: short space after
-%% #5: if 3, it means that we must not put a flat after the clef, otherwise it's the height of the flat
-%% #6: the type of the secondary clef : c or f
-%% #7: the line of the secondary clef (1 is the lowest)
-%% #8: if 3, it means that we must not put a flat after the secondary clef, otherwise it's the height of the flat
-\def\gre@typeclef#1#2#3#4#5#6#7#8{%
-  \setbox\gre@box@temp@width=\hbox{%
+
+% sets \gre@box@temp@width with the clef, the arguments are the same as \gre@typeclef
+\def\gre@boxclef#1#2#3#4#5#6#7#8{%
```
henryso: This macro needs to be added to Command_Index_internal.tex.

tex/gregoriotex-spaces.tex
```diff
@@ -420,6 +428,77 @@
   \relax %
 }
 
+% the width of the current
+\newdimen\gre@dimen@clefwidth@current\relax%
```
henryso: These dimensions need to be added to Command_Index_internal.tex.

tex/gregoriotex-spaces.tex
```diff
@@ -1212,9 +1291,6 @@
 % begindifference is the difference between the beginning of the text and the beginning of the notes. Warning : it can be negative.
 \newdimen\gre@dimen@begindifference\relax%
 
-% the width of the clef
-\newdimen\gre@dimen@clefwidth\relax%
```
henryso: This dimension needs to be removed from Command_index_internal.tex if it's there.

henryso:
I can't review the C code since I wrote it, but I'm done reviewing the TeX code and documentation.

eroux:
done, thanks

henryso:
Looks fine to me.  If you're OK with the C code, then I think this can be merged.

eroux:
Thanks!

PR #1203, opened 2016-08-11
Added fix for bad \hbox, thanks to @OldClaudio.

### LINKED ISSUE #1202: glitch in \gre@setinitial
Report from @OldClaudio:

The start of `\gre@setinitial` (in `gregorioriotex-main.sty` file), that is responsible of actually inserting the commentary over the first staff is as such:

```
\def\gre@setinitial#1{%
    % see comments on this function to see what it does
    \gre@debugmsg{commentary}{Time to calculate the true raise.}%
    \gre@calculate@commentarytrueraise%
    % we start by printing the commentary
    % when doing this we place it inside a box as which is linewidth wide with an \hfill before to get it to align to the right of the page
    % we then place that box inside one of 0 width so that our cursor doesn't move when we place the commentary
    \ifvoid\gre@box@commentary\else%
        \gre@debugmsg{commentary}{I am printing a commentary box}%
        \raise\gre@dimen@commentarytrueraise\hbox to 0pt{\hbox to \gre@dimen@linewidth{\hfill\box\gre@box@commentary}}%
    \fi%
```

The line before the last one (`\raise...`) appears to me to be wrong, to the point that the log file reports a bad hbox exceeding the linewidth by another linewidth.

In facts `\raise` raises the external hbox by `\gre@dimen@commentarytrueraise`. Such box is defined to be 0pt wide; its contents is another hbox which is declared to be `\gre@dimen@linewidth` wide. But the external hbox does not contain any glue to shrink to 0pt. I think the statement should be modified by adding the `\hss`  special glue that has a zero natural width and a first order infinite stretch and shrink components; I suggest:

```
  \raise\gre@dimen@commentarytrueraise\hbox to 0pt{\hbox to \gre@dimen@linewidth{\hfill\box\gre@box@commentary}\hss}%
```

Remember that `\hss` is an (almost primitive) glue value used also by plain TeX; therefore there is no problem with the users who use luatex instead of lualatex to typeset their documents while using gregoriotex

henryso:
Fixes #1202.

The tests don't change, but I don't know what to do to specifically test this fix.

Please review and suggest a test or merge if satisfactory as-is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 [Unreleased][unreleased]
 ### Fixed
 - When the note after an oriscus is at the same pitch, the oriscus will now point downwards by default (see [#1177](https://github.com/gregorio-project/gregorio/issues/1177)).
+- With thanks to Claudio Beccari (@OldClaudio), adding a commentary no longer generates a bad `\hbox` during TeX processing (see [#1202](https://github.com/gregorio-project/gregorio/issues/1202)).
+
 
 
 ## [4.2.0-rc2] - 2016-07-05
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -428,7 +428,7 @@
   % we then place that box inside one of 0 width so that our cursor doesn't move when we place the commentary
   \ifvoid\gre@box@commentary\else%
     \gre@debugmsg{commentary}{I am printing a commentary box}%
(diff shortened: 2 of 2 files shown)


rpspringuel:
There's no visual test that can be done to test this fix.  It eliminates a overfull box warning, so one would have to test for it in the log.  I suppose, also, that if one could get LuaTeX to fail on all warnings, then one could also write a test for which this was the only warning (and thus would fail pre-fix and pass post-fix).

eroux:
Looks good, thanks a lot!

PR #1204, opened 2016-08-12
Appended FILE_VERSION to the gregorio executable file name.

### LINKED ISSUE #1197: Installation over TeXLive 2016
When installing Gregorio 4.2 over a TeXLive 2016, one ends up with two versions of the executable on one's system and depending on one's method of usage and PATH settings (both generally and for specific programs) the wrong version may end up getting used.

For the full blown installers, I'm configuring them to move the TL2016 executable aside, but preserve it so that during uninstallation it can be restored to its original location.

I also plan on adding a version check of the installer into the version check scripts which are executed during package loading.  Thus even if the above doesn't work, the user will be alerted to the problem in an understandable manner.

I also may add a `install-exec-hook` so that the installation from source using `make install` works like the full blown installers.

henryso:
Fixes #1197.

I think I've covered everything *except* the system-setup scripts.  It's possible to modify those to look for any version of the gregorio executable on the PATH, but I'm not entirely sure how to do that from the Windows batch file, and it might not be a good idea to make this particular script too complicated.

The test harness was also updated to handle this.

Please review and merge if satisfactory (inasmuch as it covers).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - With thanks to Claudio Beccari (@OldClaudio), adding a commentary no longer generates a bad `\hbox` during TeX processing (see [#1202](https://github.com/gregorio-project/gregorio/issues/1202)).
 
 ### Changed
-- In order to facilitate installation alongside TeX Live, the version number is now appended to the gregorio executable file name.  If you are running the executable directly in your custom scripts, you will need to change them to include the version number.  If you are not using the TeX-Live-packaged version of Gregorio, you will probably need to use the `--shell-escape` option when compiling your `.tex` files.  See UPGRADE.md and [#1197](https://github.com/gregorio-project/gregorio/issues/1197) for more information.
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
 - When the note after an oriscus is at the same pitch, the oriscus will now point downwards by default (see [#1177](https://github.com/gregorio-project/gregorio/issues/1177)).
 - With thanks to Claudio Beccari (@OldClaudio), adding a commentary no longer generates a bad `\hbox` during TeX processing (see [#1202](https://github.com/gregorio-project/gregorio/issues/1202)).
 
+### Changed
+- In order to facilitate installation alongside TeX Live, the version number is now appended to the gregorio executable file name.  If you are running the executable directly in your custom scripts, you will need to change them to include the version number.  If you are not using the TeX-Live-packaged version of Gregorio, you will probably need to use the `--shell-escape` option when compiling your `.tex` files.  See UPGRADE.md and [#1197](https://github.com/gregorio-project/gregorio/issues/1197) for more information.
```
eroux: Maybe it's worth specifying that the TeXLive 2016 binary is still unsuffixed

eroux:
apart from my small comment, looks good to me!

rpspringuel:
Thanks for doing this.  I've been so busy around the monastery that I haven't had time to really work out what I was doing wrong.  Hopefully I will find some time this weekend to take care of the rest of the work.

rpspringuel:
I've sumbitted a PR with some clean-ups that I had done in my own work on this issue.  I think that with those this should be good to merge.

henryso:
Merging this.  Per @rpspringuel, changes to system-setup will come later.

Issue #1199, opened 2016-08-06
TeX Live 2016 test breaks


henryso:
Under TeX Live 2016 (but not TeX Live 2015), the `plain-tex/PopulusSion/PopulusSion.tex`, `plain-tex/braces/over_braces.tex`, and `plain-tex/text-styles/text-styles.tex` tests fail with:
```
infwarerr: The package is already loaded, stopped
) (/usr/local/texlive/2016/texmf-dist/tex/generic/oberdiek/ltxcmds.sty
ltxcmds: The package is already loaded, stopped
)
! Undefined control sequence.
<argument> ...degraphics \@undefined \else \@ifpackageloaded 
                                                  {pst-pdf}{}{\@ifpackageloa...
\AtBeginDocument #1->#1
                       
l.1271   }
        
? 
! Emergency stop.
<argument> ...degraphics \@undefined \else \@ifpackageloaded 
                                                  {pst-pdf}{}{\@ifpackageloa...
\AtBeginDocument #1->#1
                       
l.1271   }
        
End of file on the terminal!

!  ==> Fatal error occurred, no output PDF file produced!
```

The custom spaces also appear to be broken under TeX Live 2016.  See the `gabc-output/glyphs/spaces.gabc` test:

![image](https://cloud.githubusercontent.com/assets/421318/17458604/15965d2e-5be4-11e6-92b9-46a0e712a51d.png)

rpspringuel:
The `plain-tex` tests are failing due to some sort of package loading error with graphicx and xcolor.  I can reproduce them with the following simple document:

```
%%%%%%%%%%%%%%%%%%
%% TeXLive 2016 compatibility
\ifdefined\pagewidth                   \let\pdfpagewidth       \pagewidth                  \fi
\ifdefined\pageheight                  \let\pdfpageheight      \pageheight                 \fi
%%%%%%%%%%%%%%%%%%

\pdfpagewidth210mm
\pdfpageheight297mm

\edef\greoldcatcode{\the\catcode`@}
\catcode`\@=11

\input graphicx.tex % for \resizebox
\input color.sty%

\catcode`\@=\greoldcatcode

\color{red}
Hello!
\bye
```

This only occurs under LuaTeX.  pdfTeX handles the above document fine.  We're thus probably looking at a LuaTeX bug.  @eroux, where was it that we were supposed to report such bugs again?

eroux:
[...]

PR #1207, opened 2016-08-16
System setup

rpspringuel:
This PR does two things:

1. The system-setup scripts are modified to account for multiple versions of gregorio residing on the same system (which happens when you upgrade gregorio in a TeXLive2016 installation).  The scripts not only show the locations of all versions of each file, but should also grab the version numbers out of them (where appropriate) so that you can see which is where.  This finishes #1197
2. PlainTeX package loading order is modified to work around the bug noticed as a result of #1199.  The bug in color/graphicx has been reported, but this means we don't have to wait for the fix for our tests to be working again.

It should be noted that the modified Windows batch file makes use of `where` which means that it is no longer 100% compatible with XP.  [Microsoft claims that XP knows `where`](https://technet.microsoft.com/en-us/library/cc753148(v=ws.11).aspx), but based on [other information I've found](http://stackoverflow.com/questions/304319/is-there-an-equivalent-of-which-on-the-windows-command-line) the 32-bit version of XP doesn't actually know the command ([though it can be added](http://dllexedown.com/bbs/search.php?sfl=wr_subject&stx=where.exe&sop=and&gr_id=&mininum=0&maxnum=10000&onetable=04_xp64_en)).  Given that XP is now 2+ years past end of life, I'm comfortable dropping full support for it at this point.

### DIFF
--- a/contrib/system-setup.bat
+++ b/contrib/system-setup.bat
@@ -169,21 +169,33 @@ for %%G in (%files%) do (
 	)
 )
 
-:: Unversion files
-set files=gregorio-vowels.dat ^
-gregoriotex-ictus.tex ^
-gsp-default.tex ^
-greciliae.ttf ^
+:: Font Files
+set files=greciliae.ttf ^
 greciliae-op.ttf ^
 greextra.ttf ^
 gregorio.ttf ^
 gregorio-op.ttf ^
+granapadano.ttf ^
+granapadano-op.ttf ^
+gregall.ttf
+
+for %%G in (%files%) do (
+	echo ##### %%G >> %output%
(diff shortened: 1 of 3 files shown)


rpspringuel:
Just realized that I can grab the version of the font files, at least in a unix environment.  I'll have to see if I can do the same for Windows.

rpspringuel:
Turns out that `otfinfo` is one of the binaries bundled in TeXLive so the procedure for getting the font info on Windows is the same as it is on Unix.

henryso:
I'm only eyeballing it, but this looks OK to me.  Merging.

Issue #1209, opened 2016-08-18
Release 4.1.5?


rpspringuel:
Should we create a 4.1.5 release?  Several of the fixes that are pending or going into 4.2 would be compatible with 4.1 and thus could be pushed to CTAN.

If we were to create a 4.1.5 I would include the following:

* #1201 
* The fixes for #1199 (94c4288 and ca8b226 and a local commit on my machine)
* 54bc8b8
* 30bd15d
* f96b2c2
* Documentation mods from fc54014

henryso:
I tend to favor this, but I think we should have a long-lived "ctan" branch for it, so that when 4.2 becomes master, we still have a "live" branch for what's on CTAN.

eroux:
Ok for me too, I don't really have an opinion on the branches, do as you prefer

rpspringuel:
I created a PR against master with the changes I proposed above.  I also named the branch `ctan` as suggested.  For now said branch is just in my fork, but if we go through with this I'll push it into the main repository.

rpspringuel:
I'm coming across something weird in terms of backwards compatibility with TL2015.  I'm still trying to isolate the problem and will let you know when I've figured out the cause.

rpspringuel:
Okay, the backwards compatibility issues were localized to my machine and not symptomatic of the release itself.

PR #1201, opened 2016-08-08
Reach the translation glyph node by traverse_id()

anthonyfok:
LuaTeX-ja prepends custom whatsit nodes to store e.g. text direction,
breaking hard-coded paths like startnode.head.next.head.next.head
in the center_translation() function.

This patch uses node.travese_id() to find our desired nodes instead.

Fixes #1180

Issue #1180, opened 2016-07-08
Translation centering with [/] fails to work when LuaTeX-ja is loaded


anthonyfok:
Hi!

Sorry, it is me again with yet another LuaTeX-ja related issue, this time with centered translation with `[/]`.  To test, I added a translation for one word in PopulusSion.gabc in the gregorio-test/tests/plain-tex/luatex-ja directory:

```diff
--- a/tests/tex-output/PopulusSion/PopulusSion.gabc
+++ b/tests/tex-output/PopulusSion/PopulusSion.gabc
@@ -17,7 +17,7 @@ style: modern;
 
 % The syntax in this part is called gabc. Please refer to http://home.gna.org/gregorio/gabc/#basis
 
-Pó(c3eh)pu(g)lus(h) Si(hi)on,(hgh.) *(;) ec(hihi)ce(e.) Dó(e.f!gwhhi)mi(h){n}us(h) vé(hi)ni(ig//ih)et(h.) (,) ad(iv./hig) sal(fe)ván(ghg)das(fg) gen(e_f_e_)tes(e.) :(:)
+Pó[People](c3eh)pu(g)lus[/](h) Si(hi)on,(hgh.) *(;) ec(hihi)ce(e.) Dó(e.f!gwhhi)mi(h){n}us(h) vé(hi)ni(ig//ih)et(h.) (,) ad(iv./hig) sal(fe)ván(ghg)das(fg) gen(e_f_e_)tes(e.) :(:)
 
 et(e) au(eh)dí(hhi)tam(i) fá(kjki)ci(i)et(i) Dó(ij)mi(ihi)nus(iv./hiHF) (,) gló(h!i'j)ri(ji!kvJI)am(ij) vo(j.i!jwk)cis(ji) su(i_j_i_)æ,(i.) (;) in(e) læ(e)tí(e!f'h)ti(h)a(hi!jVji)
```

`luatex PopulusSion.tex` runs fine until near the end:

```
(./PopulusSion-4_1_4.gtexerror: ...exlive/texmf-dist/tex/luatex/gregoriotex/gregoriotex.lua:325: attempt to index field 'head' (a nil value)
.
\GreEndScore ...re@localrightbox {}\fi \hfil \par 
                                                  \ifgre@keeprightbox \globa...

l.725 \GreEndScore
                 %
? 
) [1{/var/lib/texmf/fonts/map/pdftex/updmap/pdftex.map}])
[...]

PR #1210, opened 2016-08-18
Ctan

rpspringuel:
PR showing what #1209 would do.


henryso:
It's hard to review this because most of the file changes are copyright dates and version bumping, but these have all be reviewed elsewhere, so as long as the tests look OK, I think this is good.

eroux:
ok for me too

rpspringuel:
Since I don't have the time to take care of this today (Sunday), on Monday I'll merge this and publish the release.  I'll also prepare a new release candidate for 4.2 at the same time so that we can get people to test the installer changes there.

anthonyfok:
Hi @rpspringuel, thank you so much for merging #1201.  But, if it is at all possible, please "stop the press!" and wait just a little bit longer (say an hour or two?) before you actually release 4.1.5: I found a bug in my code, in that when `\setmainjfont` is used to set the Chinese/Japanese font, the user is likely to run into this crash:

```
! error:  (node lib): lua <node> expected, not an object with type nil
```

For example, "Adobe Ming Std" and "AR PL UKai TW" work fine, but STZhongSong, Fandol, cwTeXKai etc. would crash... for some reasons, with some fonts, the glyph node isn't found within the hbox, but instead one level deeper.

I have a 3-line, patch, not perfect, but at least no more crashing.

I will write more at #1180 or #1201.

Thanks again!

PR #1211, opened 2016-08-19
Suppressed \GreLastOfScore if only no-element syllables follow it.

### LINKED ISSUE #1205: problems when final syllable has no notes
The following example:

```
\documentclass[a5paper, 12pt]{book}
\begin{document}
\gresetinitiallines{0}
\gabcsnippet{(c4) Ma(i)rí(kko>)a!(j.) R(::) Salve Mater.()}
\end{document}
```

produces the strange result:

buggregorio

I think this is because of the way `\GreLastOfScore` is handled...

henryso:
Fixes #1205.

I'm still unsure if this causes a problem related to #1116.  @rpspringuel Would you be able to check?

No tests break, and I added the test from the issue.

Other than that, please review and merge if satisfactory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Fixed
 - When the note after an oriscus is at the same pitch, the oriscus will now point downwards by default (see [#1177](https://github.com/gregorio-project/gregorio/issues/1177)).
 - With thanks to Claudio Beccari (@OldClaudio), adding a commentary no longer generates a bad `\hbox` during TeX processing (see [#1202](https://github.com/gregorio-project/gregorio/issues/1202)).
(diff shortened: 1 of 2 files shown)


eroux:

rpspringuel:
I'm not sure what you're concerned about, but will take a look at the results later.  @henryso, if you could elaborate on what you're worried about, I'll have a better idea of what to look for.

henryso:
In rpspringuel/gregorio#5, I moved `\GreLastOfScore` to its position prior to this pull request.  That was either part of the fix for #1110 itself or the issue you identified in #1116.

rpspringuel:
Well, the problem I identified in #1116 has to do with discretionaries mid-score, so this won't effect that (as evidenced by the fact that the clef-change test doesn't change).  As for #1110, there is a test for that now and it looks good too, so I don't see any problems.  All tests pass and the new test result looks good.

henryso:
Ok, then I guess this is good to merge.  Thanks!

PR #1217, opened 2016-09-07
Made install-gtex.sh generate uninstall-gtex.sh.

henryso:
This has been discussed in the past.  This basically generates an install manifest.  I'm not sure how best to document the use of this generated script on the website, which obviously should not happen until we release 5.0.


eroux:
good idea, thanks a lot!

PR #1218, opened 2016-09-07
Obsoleted all deprecations.

henryso:
The only tests that break are the backwards-compatible tests, which have been adjusted.


henryso:
Any objections to merging this?

.gitignore
```diff
@@ -105,6 +105,7 @@ auto/
 
 #gregorio specific
 doc/_minted-*
+doc/*.pyg
```
rpspringuel: This is now redundant with a modification that was made as part of 4.2 (where all `*.pyg` files are ignored, not just the ones in `doc`).

henryso: Done.  Please merge if this is satisfactory.

PR #1222, opened 2016-09-08
Prevented a rising note after an oriscus flexus from generating a porrectus.

### LINKED ISSUE #1220: glyph determination improvement when invalid oriscus
The following MWE:

```tex
\usepackage{libertine}
\begin{document}
\gabcsnippet{(c4) A(g)men.(hogh.) (::) A(g)men.(hog!/h.) (::)}
\end{document}
```

gives

bug

while the first should be equal to the second according to the people working on the Italian Antiphonale project. I kind of agree: when a user typesets an o, it probably shouldn't be ignored. But this complexifies the glyph determination process I'm affraid... @henryso what do you think? It's not really top priority, as it's easy to work around.

henryso:
Fixes #1220.

The tests have been updated to reflect and exercise this change.


eroux:

PR #1227, opened 2016-09-18
Doc fix

rpspringuel:
I still need to document `\grenabccharno`, but I need to know what it does first.  That, of course, would be easier if someone who knows off hand would chime in.

Barring that, I'll try to figure it out from a detailed code inspection when I next get a free block of time (I have some homework to do for now).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,3 +123,4 @@ res.txt
 *.rc
 gregorio-*
 ctan
+*.pyg
--- a/.gregorio-version
+++ b/.gregorio-version
@@ -1,4 +1,4 @@
-4.2.0-rc3
+4.2.0
 
 *** Do not modify this file. ***
 Use VersionManager.py to change the version.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,65 +2,31 @@
 All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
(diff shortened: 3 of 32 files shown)


henryso:
`\grenabccharno` has been in from the beginning of gregoriotex-nabc.tex, but it was defined by `\newcommand`, so it wasn't caught before.  The best person to ask would be @jakubjelinek, but it simply calls (in Lua) `gregoriotex.parse_nabc`.  Looking at this function, it prints the nabc glyphs for the nabc string passed as `#1`. `#2` is the font name used as an index into nabc glyph tables, and `#3` is a scaling factor that is currently hard-coded to 1 (presumably for some future configurability?)

rpspringuel:
Given that description, I think this function needs to be moved out of the user namespace (where it currently resides) and into the internal namespace (i.e. be renamed to `\gre@nabccharno`).  Are there objections to me making that change and documenting this function as an internal function?

henryso:
I have no objection, but I'm also not the author of said macro.

rpspringuel:
Pending approval from @jakubjelinek on my renaming his macro, I think this is ready for merge and then 4.2 can be released.

rpspringuel:
I just added the last two commits necessary for release: updating the version number and consolidating the CHANGELOG entries from the betas and release candidates.  I don't have time to do anything more tonight, but assuming no objections, then my plan for tomorrow morning is to merge this PR, merge the resulting 4.2 branch into master, and build the release files.

henryso:
No objections from me.

PR #1239, opened 2016-10-02
Shortened the Dominican plica left stem and corrected plica shape errors.

### LINKED ISSUE #1238: Changes to Dominican plica
See https://mail.gna.org/public/gregorio-users/2016-09/msg00002.html

The length of the left stem of a plica doesn't really correspond to the length of a virga, so the rules should be different.  Also the "openqueue" variety on the ascending figure has the stem going in the wrong direction.

henryso:
Fixes #1238

I added a test specifically for Dominican plicae.


eroux:
Ok for me, thanks a lot!

PR #1243, opened 2016-10-17
Passed height arguments to \GreNABCNeumes.

### LINKED ISSUE #1181: passing height arguments to \GreNABCNeumes
I think that in order to allow more interesting vertical placement of nabc neumes, it would be necessary to have two more arguments to `\GreNABCNeumes` which would be the highest and lowest height of the notes (or signs) of the square notes it corresponds to. This would allow two things:

- if, on a line, only one note is at a very high pitch, not all nabc neumes need to be raised, only those corresponding to this note
- in some layouts, the nabc notes are always just above the notes, even if it's inside the staff lines

Once this is done we can discuss more in details about the kind of layout and TeX interfaces we want, but this is a first step we'll have to make anyway...

henryso:
Fixes #1181.

The `#3` and `#4` arguments to `\GreNABCNeumes` are presently ignored.  Tests don't change.  I didn't add a CHANGELOG entry since there is no user-visible effect to this change.

### DIFF
--- a/doc/Command_Index_gregorio.tex
+++ b/doc/Command_Index_gregorio.tex
@@ -1095,12 +1095,14 @@ Macro used to prevent a line break from occurring at a given position.
 \macroname{\textbackslash GreScoreId}{}{gregoriotex-main.tex}
 A Lua\TeX\ attribute which designates a unique identifier for each score.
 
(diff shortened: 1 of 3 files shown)


eroux:
Looks good, thanks a lot!

Issue #1029, opened 2016-03-12
gabc <clear>tag


jakubjelinek:
I've noticed that
(c3)
<sp>V/</sp>. R{e}(h)ges(h) Thar(h)sis(h) et(h) ín(hi)su(h)læ(h.) (,)
mú(h')ne(f)ra(e) óf(f_0h)fe(h)rent.(hiH'Ghih.ghG'FE'fggf.) <sp>R/</sp>.(::)
Re(h)ges(h) A(h)ra(h)bum(h) et(h) Sa(hi)ba(h'_) (,)
do(h)na(f') ad(e)dú(f_0h)cent.(hiH'Ghih.ghG'FE'fggf.) (::)

emits R/ too much to the left of the :: bar, haven't checked yet if it is old or new algorithm, but 4.0 wasn't doing this.  Is there an option or some gabc syntax that would tell that I want it only at the bar or later, but not earlier? I've been using R/ etc. on the (::) rather than on the next syllable, which also has ugly issues (if the bar ends up at the end of a line, R/ etc. is emitted there instead of the next line), so perhaps being able to use (::)
<sp>R/</sp>. R{e}(h)...
instead would be nice, but having some option/syntax to say that I don't want it in the middle of previous melisma would be appreciated.  Though it is true Solesmes books put it happily even slightly before the bar.

eroux:
I think you want to tune `maxbaroffsettextX@Y`, see `gsp-default.tex`.

jakubjelinek:
That applies to all text though, right?  I think I'd like different handling for selected special characters, or a way to force it through some gabc markup.  It is fine if normal text is moved to the left, I'd just find it cleaner if verses or subverses couldn't overlap each other.

eroux:
Ok, understood... Well, I don't really know how to do that... perhaps a markup is needed in gabc, I don't really know what it should be... ?

henryso:
How about a new notation in gabc text, let's say `<clear>`, which tells GregorioTeX that the given syllable may not overlap any notes from any previous syllables?  Not that I know how to implement this on the TeX side, but does that sound feasible?

rpspringuel:
One could use the verbatim tags (`<v></v>`) to insert `\hspace` or `\kern` commands before or after the special character, but there would probably be a bit of playing around to get it right.

rpspringuel:
@henryso, I can see two ways of implementing that sort of thing on the TeX side:

[...]

Issue #1007, opened 2016-03-04
counting clefs in additional bottom/top spaces


eroux:
Working on #883, I realize it never occured to me that clefs on the first and fourth line should have the same effect on additional bottom/top space as `c` and `k`. Not an emergency though...

eroux:
Oh, virgula and divisio minima should be counted like that too, this would allow much more precise settings

henryso:
Virgula and divisio minima are *on* the top line, so shouldn't they count as a punctum on the top line, on `j` rather than `k` (for a 4-line staff)?  If that's the case, they shouldn't really change the additional top space, correct?

eroux:
absolutely yes

henryso:
@eroux Can you give me some samples where a clef on line 1 or 4 should alter something, so that I can see if what I am doing is having any real effect?

eroux:
I think you can take an example similar to the one in https://github.com/gregorio-project/gregorio/issues/960

henryso:
Thanks.  That opens one question (in my head).  Does the clef at the beginning of the line affect these spaces or only clefs within a line?

eroux:
Hmmm, I don't think a clef at the beginning of line should be treated differently...

PR #1247, opened 2016-11-02
Added protrusion handling.

### LINKED ISSUE #931: protrusion

henryso:
Fixes #931.

Note: this is built upon changes in the fix-1029 branch, and so merging it will merge that branch as well.  It also means the changes are merged in the pull request.  You can focus on 87f406c to see the changes specific to this feature.

Tests expectations have been updated and a new test is added for the new feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,7 @@ fonts/*.ttf
 *.bz2
 *.zip
 *-[ly].[ch]
+*-y.output
 windows/Ouput
 *.dirstamp
 config_.h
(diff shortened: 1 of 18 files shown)


eroux:
Looks very nice! The test with a protrusion factor of 0.5 is really what I was looking for!

eroux:
(so if there's o objection, I think we can merge)

henryso:
Although this will pull in the (thus far do-nothing) changes for #1029, since there have been no objections, I will merge this tonight so I can pull in my work for #1007.

PR #1249, opened 2016-11-08
Made high and low clefs influence the additional space calculation.

### LINKED ISSUE #1007: counting clefs in additional bottom/top spaces

henryso:
Fixes #1007.

Nothing has been done for the divisio minima or the virgula since those bar shapes, being *on* the top line, do not affect the space calculation.

I added a test for the specific case and updated the expectations (mostly gabc-to-gtex) which change appropriately.


eroux:

PR #1251, opened 2016-11-08
Added minimalinitialwidth dimension.

### LINKED ISSUE #1213: minimal width of initial
A request from Solesmes: it would be useful to be able to set a minimal width for an initial, so that there is no big difference between an I and an M as intials, and make a more coherent output.

henryso:
Fixes #1213.

Added a test for this new feature.  No other tests change.

### DIFF
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -444,11 +444,11 @@
     \gre@debugmsg{ifdim}{ manualinitialwidth = 0pt}%
     \ifdim\gre@space@dimen@manualinitialwidth=0 pt\relax%
       \global\gre@dimen@initialwidth=\wd\gre@box@initial %
-      \ifdim\gre@dimen@initialwidth<\gre@space@dimen@minimalinitialwidth\relax%
-        \gre@style@initial%
-        \global\gre@dimen@initialwidth=\gre@space@dimen@minimalinitialwidth\relax%
-        \endgre@style@initial%
-      \fi%
+      \gre@style@initial%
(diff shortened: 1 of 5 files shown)


tex/gregoriotex-main.tex
```diff
@@ -475,6 +480,11 @@
     \gre@debugmsg{ifdim}{ manualinitialwidth = 0pt}%
     \ifdim\gre@space@dimen@manualinitialwidth=0 pt\relax%
       \global\gre@dimen@initialwidth=\wd\gre@box@initial %
+      \ifdim\gre@dimen@initialwidth<\gre@space@dimen@minimalinitialwidth\relax%
+        \gre@style@initial%
```
eroux: the `style` lines look a bit weird, what happens if you remove them?

tex/gregoriotex-main.tex
```diff
@@ -444,6 +444,11 @@
...
```
henryso: I copied that from the manualinitialwidth code immediately below.  I'm pretty sure it's there to handle relative dimensions like em since space@dimen macros are text.  Therefore, I don't think the style lines should be removed.

rpspringuel:
That is their purpose, but I think the need to be outside the conditional so that the comparison picks up the right font settings too. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

henryso:
Thanks for catching that.  The review process is working!  Fixed.

eroux:
ok, I think we can merge

PR #1252, opened 2016-11-09
Made the generated uninstall-gtex.sh TDS-zip-friendly.

### LINKED ISSUE #1223: Install uninstall-gtex.sh into texmf tree?
Now that I added the code to generate an uninstall manifest (see #1217), should this script be installed into (and therefore also delete itself from) the texmf tree into which it installed GregorioTeX?

This would make the TDS installation different from the command-line installation, but it would seem to be a good place to install the manifest for easier uninstall of GregorioTeX.

If we do this, I presume it would be under `scripts/gregoriotex`?

henryso:
Fixes #1223.

No tests change, nor should any test be affected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,5 +125,4 @@ res.txt
 gregorio-*
 ctan
 *.pyg
-uninstall-gtex.sh
 .agignore
--- a/install-gtex.sh
+++ b/install-gtex.sh
@@ -116,7 +116,6 @@ case "$arg" in
     tds)
         TDS_ZIP="${NAME}.tds.zip"
         TEXMFROOT=./tmp-texmf
-        GENERATE_UNINSTALL=false
         ;;
     var:*)
         TEXMFROOT=`${KPSEWHICH} -expand-path "\$${arg#var:}"`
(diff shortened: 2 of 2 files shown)


rpspringuel:
The changes look good.  I'll see if I can't make some time to test them tomorrow.

rpspringuel:
`uninstall-gtex.sh` is not executable.  Can you set its executable bit when you create it?

The `--no-uninstall` flag is not working for me.

henryso:
I'm not sure `uninstall-gtex.sh` should be executable.  Wouldn't it be better to instruct users to run `bash uninstall-gtex.sh` rather than accidentally running `uninstall-gtex.sh` because it's executable?  The executable bit in a zip file is also not completely portable (though it works with pretty much everything except the original `pkzip`).

`--no-uninstall` is a vestige; it really depends on setting GENERATE_UNINSTALL (i.e., `GENERATE_UNINSTALL=false ./install-gtex.sh`).  At this point we have two options.  I can either make `--no-uninstall` work, which would require more sophisticated argument parsing, or take that out of the usage and leave it to those reading the comments at the top of file to understand how it works.  I favor the latter.

rpspringuel:
I've no problem with the instructing the users to use `bash uninstall-gtex.sh` but in that case isn't the shebang redundant?  It's my understanding that the shebang is there to indicate which program to use to run an executable and thus would not be necessary in a script which is not meant to be an executable.

Removing `--no-uninstall`  from the usage statements sounds fine.

henryso:
Yes, your understanding of the shebang is correct; it is not necessary, but it hints to some editors (like mine) that the file is a bash script.  However, I can accomplish that other ways, so I will remove the line and replace it with a comment.

rpspringuel:
Okay, everything checks out.

Issue #1255, opened 2016-11-23
A forced hyphen in a syllable affects the syllable before it.


henryso:
In a gabc sequence such as `aaa(a)bbb-(b)ccc(c)`, the "aaa" portion gets a forced hyphen even though it shouldn't.

PR #1256, opened 2016-11-23
Fixes for protruding characters and spurious hyphenation.

### LINKED ISSUE #1254: Syllables ending with a hyphen disrupt a translation that follows.
A gabc sequence like `test-[test](def)` will:
1. truncate the translation to its first character,
2. append the `\GreForceHyphen` to the truncated translation, and
3. cause a memory leak (because the truncated characters are not freed)

Because the code is similar, a similar bug also occurs with auto-protruding characters (`,`, `.`, `;`, and `:`).

henryso:
Fixes #1254 and #1255, the latter of which was discovered in testing the prior, which is why these are in the same branch.

I altered the translation test to include these cases, and several expectations needed to be updated due to the spurious hyphenation fix.


eroux:
ok for me, thanks a lot!

PR #1258, opened 2016-11-28
Fix 1029

rpspringuel:
A somewhat working attempt to make things line up.  There are some quirks still that I'm not quite sure what to do with.  They're small enough, however, that could probably be neglected for now.  Rather than tell you what they are, I'd like to see if others can spot them as if you can, that clearly indicates that they need to be fixed.

### DIFF
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -1190,6 +1190,11 @@
     \gre@dimen@temp@three=\gre@space@dimen@maxbaroffsettextleft@eol\relax%
     \gre@dimen@temp@two=\gre@space@dimen@maxbaroffsettextright@eol\relax%
   \fi%
+  % If the bar syllable is to be cleared from overlapping the previous syllable then we force the text and bar to be exactly aligned.
+  \ifgre@textcleared%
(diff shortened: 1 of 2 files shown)


henryso:
I hope you don't mean that comment towards me, because I don't really grok the spacing algorithms and all the underlying spaces and how they are composed to make the system produce what it does.

rpspringuel:
Actually, I don't expect anyone to spot the problems in the code (though if anyone can, that would be nice).  Rather I'm looking for people to examine the test results and visually decide if they are "right" or not.  I can spot the "problems" because I've been looking at the results with the barspacing and notespacing markers, but they're fairly small and so I'm not sure that someone approaching an unmarked score with fresh eyes will even be able to spot them.  If no one sees them, then I think it's reasonable to accept this as is and wait for some use case to crop up in which those small problems become unacceptably large before spending more time trying to quash them.

henryso:
On that front, I personally think that we should look at the output, and if it looks reasonable (i.e., not bad or good enough) then it's fine.  Given my position on that, the output looks fine.  I cannot definitively or even subjectively say that the difference is "worse" or "better" than before.

eroux:
The code looks simple enough, I think this can be merged, thanks a lot!

PR #1260, opened 2016-12-06
Execute defaults as root

rpspringuel:
The TeXShop plist to which the gabc extension needs to be added is not owned by the user and thus elevated privileges are needed in order to modify it.

### DIFF
--- a/contrib/TeXShop/auto-configure.command
+++ b/contrib/TeXShop/auto-configure.command
@@ -34,7 +34,7 @@ fi
 TeXShopDir=`osascript -e 'POSIX path of (path to app "TeXShop")'`
 
 echo "Adding gabc to list of valid extensions in TeXShop"
-defaults write "$TeXShopDir/Contents/Info.plist" CFBundleDocumentTypes -array-add '<dict>
+sudo defaults write "$TeXShopDir/Contents/Info.plist" CFBundleDocumentTypes -array-add '<dict>
 <key>CFBundleTypeExtensions</key>
 <array>
 <string>gabc</string>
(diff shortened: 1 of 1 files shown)


rpspringuel:
And a couple of other changes which help.

I realize no one can test this (since it's Mac specific) so please just check to make sure my bash scripting is up to standards.

contrib/TeXShop/auto-configure.command
```diff
@@ -61,13 +61,25 @@ defaults write "$TeXShopDir/Contents/Info.plist" CFBundleDocumentTypes -array-ad
 
 echo "Adding Gregorio file extensions to appropriate preference lists"
 #enable syntax coloring and the Typeset button for gabc files
-defaults write TeXShop OtherTeXExtensions -array-add "gabc"
-defaults write TeXShop OtherTeXExtensions -array-add "gtex"
+TeXExtensions=$(defaults read TeXShop OtherTeXExtensions 2>/dev/null)
+if [[ ! $TeXExtensions == *"gabc"* ]]; then
```
henryso: Why are you quoting "gabc" on this line (and similarly on the if statements below).

rpspringuel: Mostly because the [model](http://stackoverflow.com/questions/229551/string-contains-in-bash) I was looking at for this sort of test did so as well.  I still don't understand 100% when quoting is necessary in a bash script and when it is not.

henryso: Here's a link on [stackoverflow](http://unix.stackexchange.com/questions/68694/when-is-double-quoting-necessary) with a pretty good explanation.

eroux:
I think this can be merged?

henryso:
Provided @rpspringuel confirms it works properly, yes.

rpspringuel:
Yes, this works.

PR #1262, opened 2016-12-08
fix #1259

### LINKED ISSUE #1259: add script to check score against syllabation rules
It would be useful to have a script (in `contrib/`) to check a score against syllabation rules. I have one ready, I'll make a PR soon.

eroux:
This adds the script to check the syllabation of scores. It has proven extremely useful in the italian Antiphonale Monasticum projects and allowed to spot many errors, so I advise everyone to use it!

It's quick and dirty and could certainly be improved (some gabc constructions can confuse it, but it's extremely rare), feel free to improve it!


jakubjelinek:
Hardcoding /home/eroux/gregorio/hyphen-la/patterns/ in the script is not going to be very usable for others.

eroux:
oops, sorry, fixed

rpspringuel:
Wouldn't it make more sense to have this tool in the hyphen-la?  I realize the script needs some specific knowledge of gabc in order to be able to determine what can be ignored when checking the syllabification, but given that it's unusable without the hyphenation patterns in hyphen-la, I feel like it would be more appropriately placed in that repository.

eroux:
I tend to disagree... this tools checks a gabc score against some hyphenation patterns, whether they are Latin or not, come from hyphen-la or not... so it's about checking a gabc score against some patterns, not checking something against hyphen-la... do you see what I mean?

rpspringuel:
Okay, I see.  I guess what confused me is that the default hyphenation pattern file name is specifically `hyph_la_liturgical.dic` which comes from hyphen-la.  I wonder, then, if there should be a default value for the pattern file.  I understand it's convenient when one uses a particular pattern file a lot, but might it not be better to have no explicit default and allow users to create wrapper scripts if they want that functionality?

eroux:
Well, `checkAllGabc.sh` *is* a wrapper script around `checkSyllabation.py`. Quoting line 27 of the README:

```
by default it will consider that the `.dic` file is `hyph_la_liturgical.dic` in the current directory. If it's not the case, please change the `PATPATH` variable in the script.
```

but if you don't want any default, I can remove it... what do you think?

eroux:
Any update on this? Also, Solesmes would like to use this but under Windows, @rpspringuel do you think it's feasible to convert the `.sh` script to `.bat`? I have no experience with it and I cannot test it...

henryso:
Conversion to .bat doesn't look infeasible for the wrapper script, but python would negate the need to have multiple scripts.  What do you think?

eroux:
that's sensible yes... I'll make the conversion later (unless someone has time to do it!)

eroux:
done, what do you think about the current state?

eroux:
that makes sense yes, done

henryso:
This looks OK to me.

PR #1264, opened 2016-12-14
Fix 1257

### LINKED ISSUE #1257: show/hide nabc
It seems it would be practical to have a high level command to show/hide nabc, something like 

```tex
\setvoicevisibility{1}{visible} % gabc
\setvoicevisibility{2}{hidden} % nabc
```

It would be useful for the Italian Antiphonale project. I think this should be quite easy...

eroux:
fixes #1257 and adjusts checkSyllable.py to some features found in Jakub's scores


tex/gregoriotex-nabc.tex
```diff
@@ -50,8 +50,25 @@
   \endgre@style@nabc%
 }}
 
+\newif\ifgre@nabcvoice@i@visible
```
henryso: Did this slip in by accident?

If not, then these two should be documented.

eroux:
sorry, I completely forgot about that! done

henryso:
Looks OK to me.

PR #1266, opened 2016-12-18
Updated spaces from the Italian Antiphonale Project.

### LINKED ISSUE #1182: changing default spaces

henryso:
Fixes #1182.

@eroux Please double check that I pulled in all the changes.  I didn't set change the eolhyphen or custosalteration settings, and I left the emergency stretch set as `\the\emergencystretch`.  You'll also see that I changed various `interwordspacetext` values from the old ex-dimensioned values (which seemed weird given ex is a relative vertical height) to the cm-dimensioned values for similar spaces from the Italian Antiphonale Project.

I added a `CHANGELOG.md` entry, but no `UPGRADE.md` because I can't see a way that a user can easily "opt out" of these changes.

See the corresponding test pull request for notes on the tests.


eroux:
Thanks a lot for that! I won't have time for a review today, but I'll review that carefully in the next few days. There will also be adjustments in the inter line spacings, but they don't impact the test in a meaningful way so this can be done separately, it will be straightforward. Thanks again!

eroux:
I think it's perfect, you did an excellent job at commenting the tests, thanks a lot!

PR #1267, opened 2016-12-21
Added excludes and made fixes from coverage testing for 5.0 release.

henryso:
See corresponding tests in gregorio-project/gregorio-test#284.

### DIFF
--- a/src/gabc/gabc-write.c
+++ b/src/gabc/gabc-write.c
@@ -231,7 +231,7 @@ static void gabc_print_char(FILE *f, const grewchar to_print)
 {
     if (write_state == GABC_AT_PROTRUSION_FACTOR) {
         write_state = GABC_IN_PROTRUSION_FACTOR;
-        if (to_print == ':') {
+        if (to_print == 'd') {
             fprintf(f, "<pr");
         } else {
             fprintf(f, "<pr:");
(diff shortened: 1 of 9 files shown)


eroux:
Thanks a lot! Looks good to me, although I admit there are a few things I didn't understand, but it's not blocking, it's ok for me to merge

src/gabc/gabc-write.c
```diff
@@ -231,7 +231,7 @@ static void gabc_print_char(FILE *f, const grewchar to_print)
 {
     if (write_state == GABC_AT_PROTRUSION_FACTOR) {
         write_state = GABC_IN_PROTRUSION_FACTOR;
-        if (to_print == ':') {
+        if (to_print == 'd') {
```
eroux: this change is the most mysterious to me...

henryso: This is a bug in the gabc->gabc translation that was made apparent by the coverage testing.  The parser uses `d` for the argument in the internal structure when `<pr>` (no argument) is used.

eroux: I admit I don't get it... can you point me to the line in the source that sets something to 'd'? (this sounds like an uninitialized variable?)

henryso: See https://github.com/gregorio-project/gregorio/blob/develop/src/gabc/gabc-score-determination.y#L957

eroux: Oh, ok, thanks!

PR #1273, opened 2017-01-06
Added [nocustos] to suppress an end-of-line custos at that point.

### LINKED ISSUE #1271: no custo after bar
In most cases, there should be no custos before the first `<sp>V/</sp>` sign (I'm still waiting for some input on this from Solesmes), but as there is no easy way to catch this in gregorio (some will use `<v>\myVBar</v>` for instance), there should be a way to prevent a custos if the line breaks after a bar. Currently it's only possible by adding a forced end of line with no custo (`z-`), it would be practical if there was a way to express that without forcing the end of line, so that the gabc file remains correct at all widths. Maybe something like `(::0)`? What do you think?

henryso:
Fixes #1271.

None of the current tests fail, and I added a test for this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - More general support for protrusions.  The `<pr>` tag in gabc indicates where a protrusion should begin.  There is an optional argument `<pr:.5>` that allows the protrusion factor to be specified (in this example, `.5`).  Additionally, the comma, semicolon, colon, and period are automatically protruded, with configurable protrusion factors.  See GregorioRef and [#931](https://github.com/gregorio-project/gregorio/issues/931) for more information.
 - `minimalinitialwidth` space, controlling the minimum amount of space an initial should take require; ignored when `manualinitialwidth` is set to something non-zero (see [#1213](https://github.com/gregorio-project/gregorio/issues/1213)).
 - new scripts in `contrib/` to check the syllabation of a gabc score against hyphenation rules, such as those on [hyphen-la](https://github.com/gregorio-project/hyphen-la)
-- the visibility of a particular nabc voice can now be set by `\gresetnabc{n}{(in)visible}` (where `n` is the number of the nabc voice), see (see [#1257](https://github.com/gregorio-project/gregorio/issues/1257)).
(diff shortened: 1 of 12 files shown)


eroux:
I think it misses an entry into CHANGELOG, and maybe also in the gabc part of the documentation... 

The `[nocustos]` thing can be applied to anything right? That's quite a good idea!

also, currently the test only tests `[nocustos]` on bars and breaks with custos on regular notes, maybe it could be mixed?

Apart from that, good job, thanks a lot!

henryso:
It's unclear to me what you mean by "mixed".  Would you please elaborate?

henryso:
The gabc part of the documentation does not (unfortunately) currently describe the actual syntax of gabc and rather covers major features that have been added (relatively) recently.  I don't think there's a good place in the current documentation to add something about this feature.  I think it's OK to leave it out until we do finally document the syntax.

eroux:
Well, I think it's necessary to document it somewhere, and enven though the gabc part of the pdf is not really the best place, at least it's somewhere. When someone will compile a gabc documentation, he'll take this as a reference (among other things), and so he'll document `[nocustos]`, so I think it's helpful to put it there, even though it won't be very useful right now... What do you think?

About mixed, something like

```
foo... (::[nocustos]) line break
bar... (::) line break with custos
foo... (a[nocustos]b) line break between a and b
bar... (ab) same
foo... (a[nocustos]) (b)
bar... (a) (b)
```

henryso:
In trying to get `(a[nocustos]b)` to work, I have noticed that GregorioTeX is very willing to push the `b` past the end of the staff (and this happens even if not using `[nocustos]` at all):

image

I'm not sure what's going on here or how to fix it.

eroux:
well, without the actual gabc code it's difficult to debug, but you can leave it for another issue if there's something wrong, we can merge this one before...

henryso:
I added these "mixed" tests, and you can see the weird pushing past the margin that I mentioned above.  Do you have any ideas how I might be able to fix it?

eroux:
Hmmm I don't know, maybe the very long texts are too twisted for Gregorio to handle... maybe the most simple solution is to just remove them....

henryso:
I adjusted the test to look nice (and also test the cases you wanted tested).  However, if I touch it in certain ways, like removing a `g` from the ...`f f f f[nocustos] g g g`... sequence, it will again push past the end of the staff.

I still have to work on the documentation.

henryso:
Documentation is in and (I think) the tests are complete.  Please review.

eroux:
Looks perfect, thanks a lot!

PR #1279, opened 2017-01-12
Added musica ficta signs as rare signs r6, r7, and r8.

### LINKED ISSUE #1278: Musica Ficta
It would be useful for producing scholarly editions of chant that flat signs, natural signs, and sharps display above the staff as "musica ficta."  This is common in editions of medieval and renaissance music to denote places where the original manuscript source did not indicate a flatted pitch (or less often in chant, a natural sign or sharp), that nevertheless was intended to be sung as a flat, usually to avoid a tritone.  Using the polyphony feature, I have had success with simple neumes.  For more complicated neumes, the polyphony feature does not work as well: E.g., (Ky(dss)ri({gv}kxFEfv)e(hiHFED/eV/@ed)

screen shot 2017-01-09 at 11 18 54 pm

Notice the undesired space between the virga with ficta mark and the descending puncta.  In some cases, the ficta sign needs to be placed over one of a cluster of descending puncta, for example, where it is virtually impossible to properly align the ficta sign with the appropriate note using the polyphony feature:

screen shot 2017-01-10 at 3 07 53 pm

In chant, the effected pitch is usually diatonic "b" (as below c-clef), which could potentially appear above the fourth line of the chant staff depending on clef placement.  It is not necessary that the ficta sign appear on the "k" line so much as it should hover above the staff directly in line with the effected neume, like the "rare" sign does.  In some editions of treble clef music, the ficta sign appears in a smaller font than the flat on the clef, although this may or may not be desirable for chant.  I've included an example for the purpose of comparison:

screen shot 2017-01-10 at 1 58 33 pm

henryso:
Fixes #1278.

All current tests pass and I added one for the new figures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - new scripts in `contrib/` to check the syllabation of a gabc score against hyphenation rules, such as those on [hyphen-la](https://github.com/gregorio-project/hyphen-la)
 - the visibility of a particular nabc voice can now be set by `\gresetnabc{n}{(in)visible}`, where `n` is the number of the nabc voice (see [#1257](https://github.com/gregorio-project/gregorio/issues/1257)).
 - `[nocustos]` may be used in gabc to prevent a custos should the line end at that point (see [#1271](https://github.com/gregorio-project/gregorio/issues/1271)).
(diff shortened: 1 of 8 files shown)


eroux:
Looks perfect, thanks a lot!

henryso:
I have confirmed that this change passes the memory leak and code coverage tests.

Issue #1125, opened 2016-05-26
Finer control on notes-text spacing


eroux:
In some cases, a short `spacelinestext` should make `d` pitch considered as a low note adding space between notes and text.

See the following example

``` tex
\documentclass[11pt]{article} %%
\usepackage{fontspec} %%
\usepackage{gregoriotex} %%
\begin{document} %%

% espace au-dessus des lignes
\grechangedim{spaceabovelines}{0.429 cm plus 0.286 cm minus 0.0715 cm}{scalable}%
%the space between the lines and the bottom of the text
\grechangedim{spacelinestext}{0.32 cm}{scalable}%

\gabcsnippet{%
(c4)
Mys(e)té(gh)ri(h)um(hg) fí(h!iwjh)de(gh)i.(h) Vel:(::) Mys(e)té(gh)ri(h)um(h) fí(i)de(h)i.(h) (::Z)
<sp>R/</sp> Mor<alt>Omnes:</alt>(e)tem(g) tu(gh)am(h) an(h)nun(h)ti(h)á(h)mus,(g) Dó(h!iwj)mi(h)ne,(h) (;) et(e) tu(gh)am(h) re(h)sur(h)rec(h)ti(h)ó(h)nem(h) con(i)fi(h)té(gh)mur,(g) (,) do(hf)nec(gg) vé(e)ni(de)as.(e) (::) (g+)
}%
\end{document}
```

in which the `i` merges within the podatus. Of course this example is a bit extreme, but I think having a system similar to what's happening at the top of the staff with (`additionaltopspacealtthreshold`) would fix this issue...

henryso:
There is a request to try to get this specifically into 5.0, so if everyone else is busy, maybe someone can give me pointers to "what's happening at the top of the staff", so I can try to look into this?

eroux:
[...]

PR #1283, opened 2017-01-19
Fixed height counting of a vertical episema on a virga / stemmed oriscus.

henryso:
Related to #1125.

I found a bug where a vertical episema on a virga lowers the low height by one, which is incorrect because the vertical episema is at the same height as the queue.

I added a test for this bug fix.


eroux:
Thanks!

PR #1287, opened 2017-01-30
Initial checkin of the grelaon font.

jakubjelinek:
AFAIK we don't have yet permission from the Laon city library, so this includes the font with the background images cleared, sfd with those images available upon request (or if we get permission, it can be checked in).
I still need to finish up the table from Gregorian Semiology for the Laon neumes and try first scores with laon neumes.


eroux:
This is really a great work, thanks a lot!

I've just sent another mail to Laon's library, we'll see...

A few remarks:
- there should be an entry in the `CHANGELOG.md` file
- I think the doc would be clearer if the beginning was more structures, with titles, subtitles, etc. here it's quite a lot of text and the structure is not obvious...

eroux:
I just got an email from the Laon library, they're ok as long as we mention the name `LAON_BM_ms239` specifically written this way and we give a pointer to the website (they didn't say which one but i guess http://manuscrit.ville-laon.fr is ok). Can you add these two informations in the comment of the sfd? And also add "Bibliothèque municipale de Laon" in the `Thanks` section of `CONTRIBUTORS.md`? (with a link pointing to the website)

jakubjelinek:
Great, thanks.  I'll do those changes and also check in the font with the background images soon, though don't know what is the current schedule, can it wait say till Friday?

henryso:
We are just trying to release the beta at this point.  Whether or not it gets into the beta doesn't matter too much.  It should still make 5.0 (which has two big issues to fix).

eroux:
Thanks a lot, that's a great step forward!

PR #1291, opened 2017-02-02
Make autocompile the default

rpspringuel:
I think it’s time to complete the transition to the autocompile system by turning it on by default.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,7 +4,7 @@ This file contains instructions to upgrade to a new release of Gregorio.
 
 ## 5.0
 
-### Auto-compiliation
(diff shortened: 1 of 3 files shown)


henryso:
If you're going to do that, you should also put a note in UPGRADE.md.

eroux:
I agree with the change, and also with Henry

UPGRADE.md
```diff
@@ -4,6 +4,10 @@ This file contains instructions to upgrade to a new release of Gregorio.
 
 ## 5.0
 
+### Auto-compiliation
```
henryso: I think this should be "Auto-compilation" (typo).

PR #1294, opened 2017-02-04
fix #1293

### LINKED ISSUE #1293: Support for third party fonts with no hollow version
Related to #844 : I'm currently adapting the Solesmes font for Gregorio 5, it's quite straightforward but I'm not planning to create a hollow font. The compilation fails when it looks for `mechtildis-hollow`, which doesn't exist... I can't find the way to do that simply nor any indication in the doc... How can I do that? I don't think we need a high level function for that, third party fonts are quite rare...

eroux:
This fixes #1293. As only Solesmes can use this macro, I don't think it's worth mentioning in the CHANGELOG, nor do I see any particular need (or way) to test...

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -453,6 +453,14 @@ This function is the same as above, with a third argument to scale the font. The
 
 The two first arguments are the same as \texttt{\textbackslash gresetgregoriofont}. The third argument is an integer representing the scaling factor, where the one used by \texttt{\textbackslash gresetgregoriofont} is 100000.
 
+\macroname{\textbackslash greloadholehollowfonts}{\{\#1\}}{gregoriotex-main.tex}
+
+If set to false, will not load the \texttt{hollow} and \texttt{hole} variants of the next font to load. Use it before loading third party fonts not having these variants (rare case).
+
+\begin{argtable}
+  \#1 & string & \texttt{true} or \texttt{false}.\\
+\end{argtable}
+
 \subsubsection{Glyph Alteration}
 In addition to the normal glyphs loaded by the choice of font, Gregorio\TeX\ also supports several methods for fine tuning the choice of glyphs.  Using the below functions, you can choose from alternative glyphs which are already built into Gregorio\TeX\ or import custom glyphs you have designed yourself.
 
--- a/doc/Command_Index_internal.tex
+++ b/doc/Command_Index_internal.tex
@@ -1672,6 +1672,9 @@ Boolean which specifies whether ledger line heuristics will be used or not.
 \macroname{\textbackslash ifgre@usestylefont}{}{gregoriotex-main.tex}
 Boolean which specifies whether the style font should be loaded or not.
 
+\macroname{\textbackslash ifgre@loadholehollowfonts}{}{gregoriotex-main.tex}
+Boolean which specifies whether the hole and hollow fonts should be loaded or not.
+
 \macroname{\textbackslash ifgre@keeprightbox}{}{gregoriotex-signs.tex}
 Boolean which specifies if we have to keep the localrightbox until the end.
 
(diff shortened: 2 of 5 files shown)


henryso:
This should still be documented in GregorioRef.

A couple of things to consider:
1. check_font_version will check all the fonts regardless of this flag.  This is OK since it'll just be rechecking greciliae, but maybe it should be changed for consistency.
2. The GreCavum macro will use greciliae (or the previous font) for hollow figures.  This is probably OK, but should either be documented or changed to emit some kind of warning when the flag is set.

eroux:
Hmm, ok, I should modify the GreCavul macro too then... I'll give it a try

eroux:
Well, I've managed to find that I should use something like

```tex
\grechangecavumglyph{Punctum}{mechtildis}{PunctumCavum}[mechtildis][PunctumCavumHole]%
\grechangecavumglyph{LineaPunctum}{mechtildis}{LineaPunctumCavum}[mechtildis][LineaPunctumCavumHole]%
```

but now the problem is that the `PunctumCavum` and `PunctumCavumHole` (and `Linea` variants) are not in the .ttf produced by `squarize.py`... I tried to add it, but I'm really not confortable doing so, as there's a lot of code handling glyphs which name ends with `Cavum` and I really don't know this code... @henryso would you have time to produce a small patch to copy these glyphs in the .ttf if they exist in the .sfd?

Thank you!

henryso:
I'll try to have something you can use by your morning.  I'm currently thinking of adding something to the json configuration for additional glyphs to copy.

eroux:
oh ok thanks! You don't have to make something too generic though, third party fonts are really rare...

henryso:
I made the requested change.  Add an `"extra glyphs"` key to the json configuration, like:

```
"extra glyphs": [ "GlyphOne", "GlyphTwo" ]
```

eroux:
Thanks Henry, the glyphs are now correctly copied, but the glyphs don't appear (although there is no error). The commands I tried are

```tex
\grechangecavumglyph{Punctum}{mechtildis}{PunctumCavum}[mechtildis][PunctumCavumHole]%
\grechangecavumglyph{LineaPunctum}{mechtildis}{LineaPunctumCavum}[mechtildis][LineaPunctumCavumHole]%
```

but nothing appears in the output... it can be seen in the bitbucket project for mechtildis (I think you can access it), I admit I'm a bitt puzzled on how to fix this...

henryso:
I suspect a problem with the font scale.  If I use greciliae Flat/FlatHole, the glyph appears.

henryso:
@eroux Give this a try.  Use:

```
\grechangecavumglyph{Punctum}{mechtildis@17307}{PunctumCavum}[mechtildis@17307][PunctumCavumHole]%
\grechangecavumglyph{LineaPunctum}{mechtildis@17307}{LineaPunctumCavum}[mechtildis@17307][LineaPunctumCavumHole]%
```

eroux:
Thanks! Still doesn't work though, this time with the error:

```
! Number too big.
l.1 ...t@variant@A = {name:greciliae} at 17000000000
                                                  0 sp\relax
l.1771 \gresetgregoriofont{greciliae}
```

henryso:
Can you send me the tex file?

eroux:
it's the test file of the mechtildis repository on bitbucket, you should have access to it, tell me if you don't

henryso:
Taking the latest version of the test file, I get unstable output (the gaux file keeps being written), but I don't get "Number too big".  This instability is a problem, but I don't know how to proceed with your problem since it's not happening on my system.

eroux:
Ok, thanks or inspecting, I'll come back to it when I have time (probably this week-end)

eroux:
hmpf, sorry, I had a problem in the way I was testing (mixing version), everything works, thanks a lot!

PR #1299, opened 2017-02-09
Small grelaon progress.

jakubjelinek:
I've typesetted so far 2 scores with grelaon neumes, one is included on the title page of the nabc documentation.
Still, only one nabc line is supported, would appreciate help with getting that extended.
Right now, the `\gresetnabc{fontname}{size}` affects all the neumes, there is no (easy) way to override the neume color and no way to say to typeset neumes below score instead of above it.  Would be nice if we had set `\gresetnabcline{nabcnumlines}{nabcline}{fontname}{size}{color}{placement}`
where one could express: if the score has nabc-lines: 2 (first argument), then for 1st nabc-line use font, size and color. So I could say:
\gresetnabcline{1}{1}{gregall}{8}{gregoriocolor}{above}
\gresetnabcline{2}{1}{gregall}{8}{gregoriocolor}{below}
\gresetnabcline{2}{2}{grelaon}{8}{black}{above}
to mean, for nabc-lines: 1; scores, use red St. Gall neumes above score, for nabc-lines: 2; place 1st St. Gall neumes below score and 2nd Laon neumes black above lines.  There could be off placement that would hide the particular neumes.  Thoughts on this?

### DIFF
--- a/doc/GregorioNabcRef.tex
+++ b/doc/GregorioNabcRef.tex
@@ -59,6 +59,26 @@
       \end{list}%
     \endgroup%
 
+    \vspace{1cm}
+    \begingroup%
+      \greannotation{\scriptsize{CO}}
+      \greannotation{\scriptsize{II}}
+      \gresetnabcfont{grelaon}{8}
+      \color{black!60}
+      \setmainfont[SmallCapsFont=AlegreyaSC]{Alegreya}%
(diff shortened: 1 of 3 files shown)


eroux:
I'm ok to merge this PR. I agree that something like what you describe should be coded, I'm just not available to do it...

PR #1300, opened 2017-02-11
Fix various significant letter typesetting issues:

jakubjelinek:
1) grelaon uses position 5 for significant letters (inside neume), which must be
   always present in the font, use highest priority for those
2) typesetting has been wrong when using toSlsa6lti8, the tironian letter which
   was supposed to be below the torculus actually ended up below the augete
3) the order of letters when at the same position matters, we would typeset
   tslst2lsc2 as tslsc2lst2 because only the latter is present in the font
4) grelaon font now has also two glyphs with 3 significant letters


eroux:
ok for me!

PR #1301, opened 2017-02-11
Finish up the Laon neume table from Gregorian Semiology (except

jakubjelinek:
for glyphs I'll still need to find).


eroux:
Thanks a lo!

PR #1306, opened 2017-02-19
Dealing with the supplemental fonts

rpspringuel:
I haven't done documentation yet and I still need to do some more testing (specifically on Windows), but I think this has all the needed functionality.  Seeing as I need to call it quits for the day, please look it over and let me know if I missed anything and I'll take that up when I resume work on this (probably Tuesday, since I have an extra day off this week).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - `initial-style` gabc header, supplanted by the `\gresetinitiallines` TeX command.
 - `biginitial` style, consolidated into the `initial` style.
 - `\grescorereference`
-- The Gregorio and Grana Padano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with their Dominican "-op" variants) in `supp_fonts-$GREGORIOVERSION.zip`.  See [UPGRADE.md](UPGRADE.md) and [#844](https://github.com/gregorio-project/gregorio/issues/844) for details.
+- The Gregorio and Grana Padano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with their Dominican "-op" variants) in `supp_fonts-5_0_0.zip` (the version number will change to match future releases).  See [UPGRADE.md](UPGRADE.md) and [#844](https://github.com/gregorio-project/gregorio/issues/844) for details.
 
 
 ## [4.2.1][Unreleased]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - `initial-style` gabc header, supplanted by the `\gresetinitiallines` TeX command.
 - `biginitial` style, consolidated into the `initial` style.
 - `\grescorereference`
-- The gregorio and granapadano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with optional Dominican "-op" variants).  ***If there are other logistics, put them here or delete this statement before release.***  See [UPGRADE.md](UPGRADE.md) and [#844](https://github.com/gregorio-project/gregorio/issues/844) for details.
+- The Gregorio and Grana Padano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with their Dominican "-op" variants) in `supp_fonts-$GREGORIOVERSION.zip`.  See [UPGRADE.md](UPGRADE.md) and [#844](https://github.com/gregorio-project/gregorio/issues/844) for details.
 
 
 ## [4.2.1][Unreleased]
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,11 +8,11 @@ This file contains instructions to upgrade to a new release of Gregorio.
 
 The `autocompile` option is now the default option for the behavior of `\gregorioscore`.  If you want to go back to the old, manual behavior, then you should pass the `nevercompile` option when loading `gregoriotex`.
 
-### Gregorio and Grana Padano fonts
+### Additional fonts
 
-The Gregorio and Grana Padano fonts, as well as their "-op" Dominican variants, will now be available as a separate download.  If you were using these fonts, either download `supp_fonts-5_0_0.zip` and install them or switch to Greciliae, which continues to be distributed (with its "-op" Dominican variants) along with Gregorio by default.
+With the addition of more cavum shapes, the number of font options which come with Gregorio (the package) by default has been reduced in order to control the download size.  From now on only Greciliae will be installed by default.  The Gregorio and Grana Padano fonts, as well as their "-op" Dominican variants, will now be available as a separate download.  If you were using these fonts, either download `supp_fonts-5_0_0.zip` and install them or switch to Greciliae.
 
-To install the new versions of Gregorio and Grana Padano, unzip `supp_fonts-5_0_0.zip`, navigate to the uncompressed folder in Terminal (or Command Prompt on Windows) and execute `texlua install_supp_fonts.lua` after you have installed the main Gregorio distribution.  This script will look for the location of Greciliae and copy the other fonts to that location.  If you wish to control the installation of the fonts manually, the script takes an optional argument with one of the following values:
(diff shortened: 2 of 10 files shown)


fonts/Makefile.am
```diff
@@ -27,13 +27,15 @@ TTF_GREGORIO_OP = gregorio-op.ttf gregorio-op-hollow.ttf gregorio-op-hole.ttf
 TTF_GRANAPADANO = granapadano.ttf granapadano-hollow.ttf granapadano-hole.ttf
 TTF_GRANAPADANO_OP = granapadano-op.ttf granapadano-op-hollow.ttf granapadano-op-hole.ttf
 TTF_EXTRA = greextra.ttf gregall.ttf gresgmodern.ttf grelaon.ttf
-TTF_ALL = $(TTF_GRECILIAE) $(TTF_GRECILIAE_OP) $(TTF_GREGORIO) \
-		  $(TTF_GREGORIO_OP) $(TTF_GRANAPADANO) $(TTF_GRANAPADANO_OP)
-SOURCEFILES = gregorio-base.sfd granapadano-base.sfd greciliae-base.sfd \
-			  greextra.sfd gregall.sfd gresgmodern.sfd grelaon.sfd
+TTF_ALL = $(TTF_GRECILIAE) $(TTF_GRECILIAE_OP) $(TTF_GREGORIO) $(TTF_GREGORIO_OP) $(TTF_GRANAPADANO) $(TTF_GRANAPADANO_OP)
```
henryso: I don't like these really long lines, but they work, so I'm not going to argue too hard.

henryso:
Make sure you test this with `make distcheck` and `make distcleancheck` if you haven't done that already.  Sometimes autotools doesn't like stuff not being in `EXTRA_DIST`.

fonts/Makefile.am
rpspringuel: Ascetically I don't like them either, but I had some trouble with tabs vs. spaces in my editor and forcing them to one line was the simplest way to get around the issue.  I'll see if I can't solve that issue and get rid of the long lines.

rpspringuel:
Can you elaborate on the proper testing procedure here?  I've not done those tests before.

henryso:
Just run those commands: `make distcheck` and `make distcleancheck`.  The commands will fail if there's a problem.

henryso:
I should also note that you should do it from the gregorio root (not the `fonts` directory).

rpspringuel:
`make distcheck` passes.

`make distcleancheck` returns `ERROR: distcleancheck can only run from a VPATH build`.  After looking that up online, it seems that `distcleancheck` needs to be run from a build happening in a side-along directory.  Once I do that, it too passes.

henryso:
Great.  Other than documentation, I think this is fine.

rpspringuel:
I just pushed the documentation.  Please look it over and make sure it's comprehensible.

CHANGELOG.md
```diff
@@ -49,7 +49,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
 - `initial-style` gabc header, supplanted by the `\gresetinitiallines` TeX command.
 - `biginitial` style, consolidated into the `initial` style.
 - `\grescorereference`
-- The gregorio and granapadano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with optional Dominican "-op" variants).  ***If there are other logistics, put them here or delete this statement before release.***  See [UPGRADE.md](UPGRADE.md) and [#844](https://github.com/gregorio-project/gregorio/issues/844) for details.
+- The Gregorio and Grana Padano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with their Dominican "-op" variants) in `supp_fonts-$GREGORIOVERSION.zip`.  See [UPGRADE.md](UPGRADE.md) and [#844](https://github.com/gregorio-project/gregorio/issues/844) for details.
```
henryso: Are we actually going to substitute the result here?  I don't particularly think we should, and if we are **not**, then this should be explained a bit less technically.

UPGRADE.md
```diff
@@ -10,9 +10,19 @@ The `autocompile` option is now the default option for the behavior of `\gregori
 
 ### Gregorio and Grana Padano fonts
 
-The gregorio and granapadano fonts, as well as their "-op" Dominican variants, will now be available as separate downloads.  If you were using these fonts, either download and install them or switch to greciliae, which continues to be distributed along with Gregorio by default.
+The Gregorio and Grana Padano fonts, as well as their "-op" Dominican variants, will now be available as a separate download.  If you were using these fonts, either download `supp_fonts-5_0_0.zip` and install them or switch to Greciliae, which continues to be distributed (with its "-op" Dominican variants) along with Gregorio by default.
```
henryso: The fact that Gregorio is the name of the font and the name of this application is confusing in this statement.  That's the reason I originally used the font filenames (lowercase) rather than the title-case name of the font.

UPGRADE.md
```diff
@@ -10,9 +10,19 @@ The `autocompile` option is now the default option for the behavior of `\gregori
...
+
+ * `auto` (optional): the same folder as Greciliae
+ * `system`: the appropriate font folder in `$TEXMFLOCAL`
+ * `user`: the appropriate font folder in `$TEXMFHOME`
+ * `<dir>`: the name of an alternate texmf root directory you want to use
+ 
+Additionally, if you are building the fonts yourself from a git clone, `install-gtex.sh` and `install_supp_fonts.lua` will install all the fonts you have built, not just the ones they are normally distributed with.
+
+**Note:** All the above methods assume you only need to access the fonts from within a TeX document (and thus are designed to put them into the correct folder in a texmf tree).  If you want to use the fonts in other programs, then you will need to consult the documentation appropriate to your platform and/or the program and manually move, copy, or link the fonts to the necessary location.
```
henryso: These instructions work for me, but I think we need to have someone less close to development read them over.

henryso:
Other than the above points, at least for me, I think it's fine.  We should have someone less technical review the instructions.  Perhaps that will happen automatically, but I don't think many people use these alternate fonts, so we might want to ask the list to specifically review these instructions.

UPGRADE.md
rpspringuel: I didn't pick up on that.  Before switching it back, however, I had the following thought:

@eroux, is there some reason why Gregorio isn't the default font?  My understanding of its development was that it was the original font for the package and it was only later that the Caeciliae and Parmesan fonts were adapted to become Greciliae and Grana Padano.  Why was Greciliae made the default?

If my understanding of the history is correct, then logically it might make more sense to make Gregorio the default font (yes, I realize this will change lots of things) and then distribute Greciliae and Grana Padano as the supplemental fonts.  This might ease confusion as then "Gregorio" would always be referring to the defaults (though the font vs. program ambiguity would remain).

Another/additional option might be to rename the Gregorio font (like we did Parmesan -> Grana Padano) in order to eliminate the ambiguity.

CHANGELOG.md
rpspringuel: I'm not sure what you mean.  Are you saying that mentioning the name of the zip file isn't necessary here?

UPGRADE.md
henryso: Gregorio was the original font that came with Gregorio.  Greciliae was adapted from the Caeciliae font and pretty much everyone agrees it looks better than Gregorio.  History being what it is, I think we should just leave these names alone.  The Parmesan rename was necessary because of a collision with the "real" Parmesan font from LilyPond.

CHANGELOG.md
henryso: No, I think an explanation of what `$GREGORIOVERSION` means is necessary here, and possibly in README.md as well.

rpspringuel: What about something like the following then:
```
The Gregorio and Grana Padano fonts are no longer distributed with the Gregorio distribution archive.  They will now be available for download separately (along with their Dominican "-op" variants) in `supp_fonts-5_0_0.zip` (the version number will change to match future releases).
```

henryso: Perfect.

henryso:
As far as I'm concerned, this is done, though we should still get someone further away from the project to read through the it.

rpspringuel:
I just pushed a revision to UPGRADE.md which hopefully makes the distinction between the font and the package clearer.

Unless there's something I should hold out for, once we merge this I'll generate a new beta and make sure to ask people to test the supplemental font installation as part of the announcement.

henryso:
I'm OK with the merge.

PR #1312, opened 2017-02-20
Added drawing of pes quadratum initio debilis glyphs.

### LINKED ISSUE #1310: initio debilis with virga
In some rare cases initio debilis should be linked with a following virga, like this example:
iubilemus

It seems that none of the following tests make it possible.

```tex
% !TEX TS-program = LuaLaTex+se
\documentclass{article} %%
\usepackage[autocompile]{gregoriotex} %%
\begin{document} %%
\gabcsnippet{(c4) Iu(f)bi(f)lé(-fgv)mus(f)}
\gabcsnippet{(c4) Iu(f)bi(f)lé(@-fgv)mus(f)}
\gabcsnippet{(c4) Iu(f)bi(f)lé(-f@gv)mus(f)}
\end{document}
```

henryso:
Fixes #1310.

I'm labeling this a bug because gregorio does parse `-fqg` into a code point that did not exist before this change.

Additional changes will be necessary if `-fgv` should generate these figures in addition to `-fqg`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+### Fixed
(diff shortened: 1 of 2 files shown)


eroux:
Ok for me, thanks a lot!

PR #1318, opened 2017-02-25
Fonts

rpspringuel:
I realized that the supplemental fonts were ending up in a few places they shouldn't in the releases.  These modifications should remove them.

**Note:** Two of the modifications make the assumption that the gregorio and granapadano fonts have not been built and thus we only need concern ourselves with their source files (`gregorio-base.sfd` and `granapadano-base.sfd`).  I could code a check for the ttf files and remove them if they are there but haven't done so at this point because I don't think it's entirely necessary.  Does anyone think otherwise?


henryso:
If you don't think it's necessary, then it's better to leave it out.

rpspringuel:
I don't think it's necessary because the same effect can be had simply by doing the builds for a release in the right order and from a clean repository.  I've already altered my release building script (which automates most of the steps to creating a full set of release files) to reflect that order and the first step in that script has always been `make maintainer-clean`, so this is a problem I'm not going to run into again.

I guess the better question, then, is "Is this sufficient?"  It certainly works so long as I'm the only one running the release build cycle.  Do we want to count on that?

henryso:
I think it's sufficient as long as the release build order is documented, perhaps in the wiki.

rpspringuel:
I've revised the wiki pages.

henryso:
Looks good.

PR #1324, opened 2017-03-02
Decouple commentary from initial

rpspringuel:
This change makes it so that the commentary box can appear in scores which don’t have an initial.


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
+### Changed
+- The printing of the commentary box is no longer tied to the printing of the inital.  As a result it is now possible to have a score which has a commentary but not an initial.  See [this thread on the user list](http://www.mail-archive.com/gregorio-users@gna.org/msg03911.html).
```
henryso: Not sure about this personally, but would it be better to list this as "fixed" rather than "changed"?

rpspringuel: I went with `Changed` because there was nothing wrong with the commentary printing code per se.  However, I suppose from the perspective of intent (was the commentary intended to be an initial only thing?) one could consider this a fix in that the code is moved from the "wrong" place to the "right" one.

I'm willing to go either way.

eroux: From my point of view it's more of a fix, but I'm not really strong minded about it

henryso: Let's make it a fix then.  I'm pretty sure it wasn't intended that commentary only belongs to scores with initials.

rpspringuel: Done.

Issue #1320, opened 2017-02-27
Problem of elision


fradec:
the following code
```tex
% !TEX TS-program = LuaLaTex+se
\documentclass{article} %%
\usepackage[autocompile]{gregoriotex} %%
\begin{document} %%
\gabcsnippet{(c4) sal(d)vúm(d)qu<e>e</e> ab(d) hos(dh)tis(h) ím(h)pe(g)tu(hvGFgwh/GFED.) (;)}
\end{document}
```
gives this
![screenshot](https://cloud.githubusercontent.com/assets/24571790/23365672/298027a4-fd04-11e6-93d8-17f6b38433d8.png)
but should give this
![screenshot2](https://cloud.githubusercontent.com/assets/24571790/23365693/3f5bf3e6-fd04-11e6-8eff-e4f5ef93aed9.png)
I known I can obtain this with `qu<e>e</e> {a}b(d)` but it would be good that the elision of a vowel preceded of 'qu' that can be correctly recognized.

rpspringuel:
I believe the intended usage of the elision tags is to encapsulate the whole elided syllable, not just the vowel in it.  `<e>que</e> ab` does get the neume position right, but italicizes (i.e. applies the elision format to) `que` not just `e`.  Is it common for people to want just the vowel in the elided syllable formatted?

eroux:
The initial elision feature request actually comes from @fradec, so I think he knows how it should be used

fradec:
The elisions concern only the vowels since the consonant must be pronounced, vocalized with the vowel of the following word, so it's not good to insert the entire syllable in the tags. See for example this hymn from Liber Hymnarius p.38
![screenshot3](https://cloud.githubusercontent.com/assets/24571790/23367556/eef64a26-fd0a-11e6-9c26-1b24eae93ef6.png)

henryso:
Question: in general, putting aside elision into account, for Latin, if you have "que", should this be equivalent to `q{ue}` or `qu{e}`?

rpspringuel:
From [here](http://ewtn.com/expert/answers/ecclesiastical_latin.htm):

> U preceded by Q or NG and followed by another vowel as in words like qui and sanguis, keeps its normal sound and is uttered as one syllable with the vowel which follows : qui, quae, quod, quam, sanguis. But notice that cui forms two syllables, and is pronounced as koo-ee. In certain hymns, on account of the metre, this word can be treated as one syllable.

[...]

PR #1325, opened 2017-03-03
Moved the change log entry for #1320 to the correct section.

henryso:


henryso:
Merging this trivial fix.

PR #1327, opened 2017-03-05
Switched Lua debug messages to use debugmessage.

### LINKED ISSUE #1326: Font mapping messages
Is it possible to assign the font mapping messages to a specific debug flag so that they don't appear in all log files?  There are a lot of them and wading through them is a chore when taking a detailed look at a log file.

henryso:
Fixes #1326.

No tests change, as expected.

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -38,7 +38,7 @@ The optional arguments are:
 \begin{tabular}{lp{10cm plus .5cm}}
   Argument & Description \\
   \hline
-  \texttt{debug} & Debug messages will be printed to the output log.  Can also be specified as \verb:debug={<types>}:, in which case only messages of the types listed will be printed to the output log.\\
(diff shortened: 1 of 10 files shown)


rpspringuel:
I'll take a look later today, but one thing that occurs to me right off the bat is that we've been really sloppy about debug message keywords thus far, simply adding whatever we felt we needed as we worked on a specific problem.  I think it might be a good idea to start documenting the debug message keywords, i.e. a list of what words have been used and a short description of what sort of messages they correspond to (much like what @henryso just did in #1326 when directing us to this PR).  Perhaps a section in `Command_Index_internal`?

rpspringuel:
I've checked the functionality and am happy with it.  The remaining issue is the documentation one I mentioned above.  @henryso, if we decide to start documenting the debug keywords, I think it's sufficient for this PR to document the three added here.  We can pick up the others in a separate PR (I'll modify `doc_check.sh` to flag undocumented debug keywords in order to generate that list).

henryso:
OK, I added the documentation you wanted (and cleaned up some problems I found in the documentation).

rpspringuel:
You've actually gone whole hog there and documented every keyword so far.  Thanks, thats actually really helpful.

Issue #1331, opened 2017-03-13
Weird behavior in nabc rendering


rpspringuel:
Debug messages are, by design, not supposed to affect the typesetting output.  However, in playing around with a new debugging feature, I've found a message (or two) which seem to be changing the way an nabc glyph is being rendered.

Take a look at the tex-output/syllable-rewriting test in the test repository.  Comment out or delete the `syllablerewriting` debug messages in `\gre@emit@syllabletext` (either, or both) in gregoriotex-syllable.tex and you'll find that the test run against this modified file fails because of a small tick mark in the nabc glyph over the syllable "spi" in the first line (all three times) is moved to the left (I advise upping the DPI to 300 in order to make the changing pixels more numerous and opaque, but the algorithm detects the change at my usual 150dpi, the pixels just happen to be fairly transparent, and thus hard to spot).  Normally I'd associate this kind of shift with a spurious space or missing `%`, but everything seems fine here in that regard.

Not knowing anything about the nabc, I can't tell which position of the tick mark is right.

Can others please take a look at this test and see if this behavior is reproducible on their setup.

PR #1332, opened 2017-03-13
Trace

rpspringuel:
I started this while working on 1285 and found it so useful that I went whole hog and formalized it and applied it to all the internal functions.  It could be applied to Gregorio or user name space functions as well, but I'm less sure that there's a gain from that.

Nominally at least, tests aren't supposed to change as a result of this PR, but #1331 was found as a result of applying this, so that needs to be resolved before this is merged.


henryso:
Why aren't all macros (i.e., user and gregorio macros) traced as well?

rpspringuel:
I only traced the internal macros because the calling of them tends to be more hidden than the calling of user or Gregorio namespace ones, which actually appear in source files (or gtex ones).  I could add tracing to all macros if that would be more useful.

henryso:
I guess some finesse is needed.  It's probably important to trace user and Gregorio macros that are called from internal macros.

rpspringuel:
That is probably a good point, but I don't know how to design a mechanism to selectively track such macros based on who is calling them.

henryso:
Well, I just meant being selective in which user and gregorio macros are traced.

rpspringuel:
That would be substantially easier to pull off.  Do you have a list of macros which you think should be traced?

henryso:
Based on recent experience, `\GreSetLinesClef`.  Maybe we should just add trace to user and gregorio macros as we encounter the need to do so.  Alternately, I guess we can look for gregorio macros called from other GregorioTeX macros?

rpspringuel:
Okay, I've added tracing to all of the possible functions in gregoriotex-signs.tex.  Take a look at what that does, and let me know if I should do the same for the other files.

henryso:
I think this is good enough.  More can be added as we encounter the need to add it.

PR #1344, opened 2017-04-02
Fix 1224

eschwab:
The tests pass for me, but I still have problems running the tests. This should be ready to go if acceptable.


rpspringuel:
I'll check the existing tests, but I'm not sure that we have any mechanism for actually testing the feature itself in the test repository.

rpspringuel:
The syllable-rewriting test fails for me, but it looks to be a false positive.  Unless some one wants to devise a method for testing this feature and add it to the test repository, I think this is good to go.

henryso:
The code looks OK to me.

PR #1349, opened 2017-04-14
TeXShop engine modifications

rpspringuel:
TeXShop 3.77 has a new feature (by my request) that allows one to add an additional parameter to TeX files which can then be exploited by the engine files.  For these engines I’ve used the parameter to add a folder (specified by the parameter) to the input paths.
I’ve also added a Plain TeX version of the engine.


henryso:
I cannot opine on this, so if it works for you, let's merge it.

rpspringuel:
Note to self: test changes before committing them.

PR #1350, opened 2017-04-15
Added --enable-texlive flag and support for a suffix-less executable.

### LINKED ISSUE #1292: name of the executable for TeXLive
Following a private discussion, it seems we have to provide an executable named `gregorio` for TeXLive (no version number appended). I see two ways to do that:

- first is to alter the build process and have a texlive option to build a binary without the file version appended (and also modify the .tex files accordingly)
- the other is to just always remove the version number and have a check on `gregorio --version` before the first score compilation

What do you think?

henryso:
Fixes #1292.

Because we laid so much groundwork for this, it took less time to implement than I thought, but I need help testing it in anything other than Linux.

I'm also not sure how to document this.  This is not a user-visible change, so it doesn't belong in CHANGELOG.md, and no entry is needed in UPGRADE.md.  As mentioned in #1292, it basically allows the build files to be set up for TeX Live.

Note: I changed @rpspringuel's code a bit because (1) redirection to `/dev/null` will likely not work under Windows, (2) to make executable check lazy, so it's only done (once per run) if needed, and (3) to try to find the suffixed executable before the suffix-less executable.

Please review and give suggestions, or merge if this is good enough.

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -37,7 +37,7 @@ local err, warn, info, log = luatexbase.provides_module({
     license            = "GPLv3+",
 })
 
-local gregorio_exe = 'gregorio-5_0_0-rc1' -- FILENAME_VERSION
+local real_gregorio_exe = nil
 
 gregoriotex.module = { err = err, warn = warn, info = info, log = log }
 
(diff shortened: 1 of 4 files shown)


tex/gregoriotex.lua
```diff
@@ -120,6 +120,37 @@ local translation_mark = 1
 local abovelinestext_mark = 2
 log("marker whatsit id is %d", marker_whatsit_id)
 
+local function gregorio_exe()
```
rpspringuel: Since you've made this a function, doesn't this mean that this code block will be executed each time we need the program name?  Since that will always be the same for a given compilation, wouldn't it be more efficient to run the check code once when the package is loaded, defining a variable with the executable name in the process and then simply accessing that variable when we need it?

rpspringuel:
For documentation, perhaps an edit to the README.md?  Or maybe adding a README-TEXLIVE.md which explains how to build the TeXLive version?

tex/gregoriotex.lua
rpspringuel: Wait, rereading your comment above, that is what you were trying to do, so clearly I've misunderstood something.  Allow me to examine the code a bit further.

tex/gregoriotex.lua
```diff
@@ -120,6 +120,37 @@ local translation_mark = 1
...
+    exe_version = io.popen(real_gregorio_exe..' --version', 'r')
+    exe_version = exe_version:read("*line")
+    if not exe_version then
+      -- look for suffix-less executable
+      real_gregorio_exe = 'gregorio'
+      exe_version = io.popen(real_gregorio_exe..' --version', 'r')
+      exe_version = exe_version:read("*line")
+      if not exe_version or string.match(exe_version,"%d+%.%d+%.")
+          ~= string.match(internalversion,"%d+%.%d+%.") then
+        real_gregorio_exe = nil
```
rpspringuel: Okay, I think I've figured the sequence now, and this line throws a flag for me.  The basic process is "if I don't know what the name of the executable is (i.e. `real_gregorio_exe` is `nil`) then figure out the name and proceed".  `real_gregorio_exe` is initialized to nil, so this block will run the first time.  After that, we know the name and simply proceed.  Thus things work fine when we can find the executable.  However, if we can't find the executable, then we set `real_gregorio_exe` back to `nil` which means that we're going to try again when we need it again (which of course will also fail).  As a result, we're going to try to find the executable (and fail) each time we need it.  Once we've failed once, we don't really need to keep checking for the executable.

rpspringuel:
On the testing front, I can build using the `--enable-texlive` flag on my Mac.  I don't have time to do it tonight, but I'll work up some dummy installers using that flag tomorrow and test the operation of the program on Windows and Mac.

tex/gregoriotex.lua
henryso: Calling `err` will generate a TeX error, interrupting the run, so we would not check again unless the user (in interactive mode) instructed the engine to continue, in which case we may want to try again just in case the user has corrected something.  At least that was my thinking, but I can change it.

henryso:
I will add the documentation of this to the README.md.

As for the testing, what's needed is more on the "does the Lua code work correctly" front, especially under Windows.  Mac and Linux are close enough that it'll probably work but of course it should be tested explicitly.

eroux:
It looks good enough to me to be merged, if some glitches appear, we can release a new version with only TeX changes, no problem

PR #1353, opened 2017-04-16
Lua-based fixes from Akira Kakuto.

henryso:
With thanks to @a-kakuto.

We should probably wait a few days for "hot" bugs to appear and then release 5.0.1 with this fix so that 5.0.1 makes it into TeX Live proper.


henryso:
Closing this to re-pull it under a 5.0.1 branch.

PR #1354, opened 2017-04-16
Lua-based fixes from Akira Kakuto.

henryso:
With thanks to @a-kakuto.

We should probably wait a few days for "hot" bugs to appear and then release 5.0.1 with this fix so that 5.0.1 makes it into TeX Live proper.


rpspringuel:
Huh, I thought I'd tested for this sort of thing.

Also, I think it makes more sense to continue to use the `release-5.0` branch for this kind of bugfix.  Otherwise we'll end up creating a whole lot of branches.

henryso:
Ok, I'll rename the branch (there is no release-5.0 branch).

PR #1357, opened 2017-04-18
Fix some typos

avbop:
Just a few typos I noticed in some of the READMEs.


rpspringuel:
These changes are acceptable to me.  Thanks for the proofreading help.

henryso:
Should we merge this directly into master?

rpspringuel:
Given that it's purely in the documentation, I don't have a problem with it per se, but I think our policies would indicate that this should be merged into release-5.0 (and develop) and only pushed to master when we make another release.

henryso:
Ok, I will re-pull it against release-5.0.

PR #1363, opened 2017-05-22
Changed from os.popen to os.execute due to TL2017 strangeness.

### LINKED ISSUE #1362: TL 2017 pretest not finding executable
I've just run the first pass of the test repository against the TL2017 pretest and it seems that the TeX/Lua code can't find the executable.  All the dump and gtex tests pass, so I know the problem isn't with the executable itself.  I'll try to make some time to work on this further this weekend, but if some one else has the time and inclination to look at this first, please feel free.

henryso:
Fixes #1362 

This seems to work in my runs under TL2017.


rpspringuel:
Making these changes to the TL2017 pretest, I was able to successfully run through the tests.  There were some tests with differences, but that's probably not a result of this change.

rpspringuel:
I've also confirmed that this works with TL2016 on my main machine.

PR #1368, opened 2017-06-19
Opened auto-compiled gabc files in TeX so opening is recorded correctly.

### LINKED ISSUE #1367: TL2017: latexmk no longer detects autocompiled gabc files as dependencies.
Prior to TL2017, we opened autocompiled gabc files in Lua so that latexmk could detect that the autocompiled gabc files were dependencies of the final output.  The gabc would be recorded in the `.fls` file, and changing a gabc file would cause latexmk to rebuild the pdf.  This no longer works in TL2017.  This technique does not record the gabc in the `.fls` file.  I tried using an `\openin` technique to do it in TeX, but, though the file is then recorded in the `.fls` file, it is still not detected as a dependency of the output.

henryso:
Fixes #1367.

Please review and merge if satisfactory.  Tests are not affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][unreleased]
-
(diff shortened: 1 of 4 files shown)


rpspringuel:
I don't usually use latexmk and haven't yet installed TL2017 on my main machine.  @eroux are you capable of testing this quickly?  If not I'll set aside some time for getting this onto my VM with TL2017 for testing.

eroux:
I can't really test it, but I trust it works and it cannot harm, you can merge

PR #1383, opened 2017-11-19
Prevented the change of \gre@diment@lastglyphwidth by a custos alteration

### LINKED ISSUE #1379: Misplaced horizontal episema
When I try to write this:

`præ(j_h)`

I see this:

screenshot from 2016-11-18 23-13-24

In a previous (>1 year old but I'm not sure exactly which) version of Gregorio the horizontal episema was on the first note. As I understand gregorio's syntax, it should be on the first note, shouldn't it? I can't figure out how to place it on the first note now. (The example is from the gradual of the Immaculate Conception, for the record.)

I am using gregorio 5.0.1 (from the Arch AUR).

henryso:
Fixes #1379.

I added a test for this change (from the MWE).  I also accepted a minor difference in another test (due, I think, to slight typesetting changes in fonts and/or TeX Live 2017).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ## [Unreleased][unreleased]
 - Fixed issue with Tex Live 2017 latexmk not detecting auto-compiled gabc files as dependencies (see [#1367](https://github.com/gregorio-project/gregorio/issues/1367)).
 - Added note about loading microtype after gregoriotex (see [#1364](https://github.com/gregorio-project/gregorio/issues/1364)).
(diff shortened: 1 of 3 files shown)


rpspringuel:
I got some 93 test failures, but all seem to be minor font/typesetting differences.  I think this is the first time I've run the test suite since installing TeXLive 2017, so that's probably to be expected.

In any event, this does seem to fix the particular problem being addressed.

henryso:
I am using TeX Live 2017, though it's been about a month since I last updated, and I unfortunately don't get those 93 errors.  I wish there were a good way to allow for tolerances like these.

rpspringuel:
We've covered that ground in the past and as I recall it all comes down to the DPI setting for the comparison.  Lowering the DPI allows one to avoid seeing all the test failures, but can also fail to detect changes to a punctum mora or episema.  I use a default DPI of 150.  Are you using something lower than that?

henryso:
I'm using 200 dpi.

henryso:
I tried it at 150 dpi and I still don't get any differences.

PR #1387, opened 2017-11-23
Added support for dotted barlines.

### LINKED ISSUE #1360: Dotted Barline
A user has requested the ability to make the dotted bar line as shown in this picture:

I've seen them elsewhere in a slightly different style (bigger and fewer dots), but can't remember where off the top of my head.  If I find it, I'll post an image.

To make this work we need the following:

- [ ] gabc syntax for the element
- [ ] lexer modifications to recognize the new syntax
- [ ] new glyph in the fonts
- [ ] modification to `\gre@writebar` for the new bar and `\GreDottedBar` alias

On the syntax issue, `:` would be the most obvious choice, but is already used for divisio maior.  We could swap in `|` for that, but such a change to the gabc syntax is probably way too invasive for those who have been using Gregorio for a while.  On the other hand I think that using `|` for the dotted bar and `:` for the divisio maior would be counter-intuitive and thus is probably not a good idea for newbies.  I'm not sure if there is any other intuitive character that could be added.

We could make it a post-fix solution, say `:0`, but that raises the question as to whether we need dotted versions for all the current bar types and not just the divisio maior.

henryso:
Fixes #1360.


rpspringuel:
Did you add just the one style or is the other style somewhere I'm not seeing?

henryso:
I only added one style that imitates the one in the image.  I don't have examples of the other style to try to imitate.

rpspringuel:
I posted a pdf with a different style.  It uses much larger dots (really more like dashes).

rpspringuel:
I just reupload the image as something that will display directly (you have to download a pdf in order to view it, so it's kind of easy to miss).

I'm running through the tests and have been able to clear at least some (if not all, I'll find out when the tests finish) of the spurious failures by clearing my image cache.

henryso:
Looking at the sample, I think that's more of a typographic difference rather than a variant/style within the typographic style.  I think it's best to just have that be a font thing and don't see a need to put another set of glyphs in the current font.  I could put this other style in the gregorio font, if you think that OK.

However this does bring up an issue of the difficulty in redefining bar glyphs.  The code currently uses font substitution to select bar glyphs based on number of staff lines.  This makes it impossible to change the bar glyphs via glyph substitution.  The solution is not as simple as an additional level of indirection because sized bars are defined as variant glyphs.  I think the solution will have to be an additional level of indirection plus a renaming of bar glyphs, but I need to think about this a bit more.

rpspringuel:
> Looking at the sample, I think that's more of a typographic difference rather than a variant/style within the typographic style. I think it's best to just have that be a font thing and don't see a need to put another set of glyphs in the current font. I could put this other style in the gregorio font, if you think that OK.

That sounds good.

> However this does bring up an issue of the difficulty in redefining bar glyphs. The code currently uses font substitution to select bar glyphs based on number of staff lines. This makes it impossible to change the bar glyphs via glyph substitution. The solution is not as simple as an additional level of indirection because sized bars are defined as variant glyphs. I think the solution will have to be an additional level of indirection plus a renaming of bar glyphs, but I need to think about this a bit more.

Would it be possible to use a sort of offset system?  I.e. set things up so that the 2, 3, 4, and 5 line variants of the bar lines always have the same relative location to each other in the code points.  We would then add a particular offset which corresponds to the number of lines to a base code point which corresponds to the kind of bar line.  Changing to a variant bar line, therefore, would be a matter of changing just the base code point (which should be fairly similar to the way our variant glyph system works for other glyphs), while changing the number of lines would be changing the offset.

henryso:
I think it's more straightforward to bring in all the sizes of bars into \GreCP variables (e.g., \GreCPVirgulaTwo, \GreCPVirgulaThree, \GreCPVirgulaFour, \GreCPVirgulaFive, etc.) and have the code that switches number of lines set the variables to be used for the given score (e.g., set \GreCPVirgula to \GreCPVirgulaFour for a 4-line staff).  Then the user could map these glyphs directly, even with asterisks if the names in the alternate font are the same.  This is my current thought, but I have dived into the code to see how feasible this implementation is.

henryso:
OK, this should now allow what you need.  To use, add:
```
\grechangeglyph{DivisioMaiorDotted*}{gregorio}{}
\gresetlinesbehinddottedbar{visible}
```
to your preamble.

rpspringuel:
The changes look good in the abstract.  I'll check them in practice when I am back at my computer.

Perhaps one of the tests should be modified to test the variant dotted bar?

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

henryso:
Added.

rpspringuel:
Looks good.

PR #1391, opened 2017-11-27
Gabc Documenation

henryso:
I took a stab at adding GABC documentation to GregorioRef ~~since no one else seemed to want to start it~~.  I've pushed this directly to the main repository in its own branch so we can all collaborate on getting this section into better shape.

Please take a look and either add what you think I've missed or let me know what I've forgotten to put in and I try to add it.

### DIFF
--- a/doc/Gabc.tex
+++ b/doc/Gabc.tex
@@ -679,20 +679,44 @@ Some examples:
   }
 }
 
-\subsubsection{Horizontal Episemata}\label{horizontalepisemata}
+\subsection{Rhythmic signs}
 
-A horizontal episema may be adjusted with the addition of numbers.  Multiple
-numbers may be added to combine their effects.
+\paragraph{\textit{Punctum Mora}}
+
+The \textit{punctum mora} (dots after the note) are denoted by adding a period ( \texttt{.} ) after the note. When there are two \textit{punctum mora} after a neume, the character is simply doubled: ( \texttt{..} ).  You cannot place more than two \textit{punctum mora} after a note.
+
+{
+  \gresetinitiallines{0}\gresetlyriccentering{firstletter}%
+  \gresetclef{invisible}\gresetlastline{justified}%
+  \ttfamily
+  \gabcsnippet{( ) g.(g.) h.(h.) g..(g..) ( )}
+}
+
+\paragraph{\textit{Ictus}}
+
+The \textit{ictus} (also called the vertical \textit{episema}) is denoted by an apostrophe ( \verb='= ) after the note. Gregorio will place the episema above or below the note according to the context, however you can force its position with \verb='0= (for always below) and \verb='1= (for always above).
+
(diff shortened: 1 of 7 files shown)


rpspringuel:
You got much more ambitious with your start than I did.  It's going to take me a while to proofread everything.

One thing I do notice, is that this PR is against `develop`.  I think we should base the gabc documentation effort against `release-5.0` with a daughter branch which pulls in `develop` and documents the gabc syntax which will be new in 5.1 (dotted barlines, for the moment).  It would be highly useful to upload a pure documentation change to CTAN for those aspects of gabc notation which are already in the version loaded there (the 5.0 series).

henryso:
I can do that, but this documentation has been this way for so long that I think we might as well cut a 5.1 soon with the new enhancements and the documentation.  It's almost time for TeX Live 2018, and we cut it pretty close last time.

rpspringuel:
The pretest for TeXLive is usually at the start of April.  Admittedly, we did cut it really close (and perhaps just a bit over) last year, so we should plan on being ready early, but it's still a solid 4 months away.  Unless this documentation process really drags through the end of the year, I expect to be cutting 5.0.4 for CTAN at about the same time that we cut 5.1.0 (which, I agree, should be soon).

henryso:
There's actually two things that would need to be taken out.  Let's try to wrap up the review and update in this branch for the 5.1.0 documentation relatively quickly.  Then I'll grab a copy of the Gabc.tex file, re-base this branch, adjust it to take out the 5.1.0 specific things and re-pull it against release-5.0 if that still makes sense when the time comes.

Note: I will be away for a good deal of December and, though I will have internet and all that fun stuff, I can't say how much time I will be able to spend on this.

rpspringuel:
Well, let's make this as quick as possible then.

I do have to say I'm not happy with the topic order/organization at this point (and I realize that's not your fault, as you're borrowing most of it from the website).  I have some ideas for what might make more sense, but need a bit of time to flesh them out.  Hopefully I'll have something later today that others can comment on.

rpspringuel:
Please look at https://github.com/gregorio-project/gregorio-project.github.io/pull/67/commits/610593acb3301c322187dfcb32ba67d4fea57e99 for the start of my ideas on restructuring.

rpspringuel:
You know, that js file isn't all that easy on the eyes so I just committed an attempt to make that structure concrete, at least for the first part of the gabc documentation.  I haven't touched the note syntax yet (except to extract those elements which I'm moving into one of the earlier sections).  This should be easier for people to study and comment on.

rpspringuel:
I don't think this is done, but I'm going to suggest merging this immediately and making a release so that we're ready for the TeXLive pretest.  We can come back to adding to the docs after we've prepped everything for TeXLive 2018.

henryso:
I agree.  Merge it when you're ready.

PR #1409, opened 2018-03-15
Script for uploading package to CTAN

rpspringuel:
This adds a script which makes use of ctan-o-mat (a script which comes with TeXLive) to upload the package to CTAN.  Naturally the CTAN package file must be built before it can be uploaded, so we check for it before trying to upload, but do not build it if it is not there.
The script attempts to scrape as much information as possible from the repository in creating the upload, but the user should always supply the text of the release message when invoking the script (as the first argument).  This can be skipped, but it is not recommended so the script asks for confirmation before letting that happen.
Finally, if it is needed, a second message can be added which is directed to the CTAN upload managers.  This is always the second argument of the script and can be left blank without a problem.

I've tested everything except the actual upload step, but can't test that until we have something to upload to CTAN.  We should probably hold off on merging this PR until we have a chance to test the script fully (i.e. whenever 5.1.1 is released).

Feel free to pick my coding apart and make suggestions for improvement, however.


ctan_upload.sh
```diff
@@ -0,0 +1,72 @@
...
+VERSION=`./VersionManager.py --get-current`
+FILE_VERSION=`echo $VERSION | sed s:[.]:_:g`
+pkgfile="gregorio-$FILE_VERSION.pkg"
+
+# copy config file
+cp "ctan-o-mat.config" "gregorio-$FILE_VERSION.pkg"
+
+# changelog entry
+changes=$(awk "/$VERSION/ {p=1; next}; /[0-9]+\.[0-9]+\.[0-9]+/ {p=0}; p" "CHANGELOG.md")
+perl -s -i -0777 -pe 's/<<<CHANGELOG>>>/$to/' -- -to="$changes" "$pkgfile"
```
henryso: Why not do this stuff with sed, so perl is not required?

rpspringuel:
I tried and ran into problems with the changelog because it's a muli-line substitution.  sed complained about unescaped newlines.  PERL was the only solution I could make work there, and just used it for the others for consistency.  If you know how to make that substitution then I'd be glad to make the change. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

henryso:
OK, didn't realize that.  I guess, then document the need for perl.

rpspringuel:
Is that sufficient or did you have something else in mind?

henryso:
That's going above and beyond (which is fine, of course).  I was only supposing you would put something in a README somewhere.

rpspringuel:
I thought about it, but couldn't figure out which README it belonged in.

rpspringuel:
I've done this merge locally and will push it directly.

Issue #1406, opened 2018-03-12
Copyright dates need to be updated


rpspringuel:
New year.  Copyright dates need to be updated to reflect that.  There are lots of these scattered through the files and it's difficult to keep track of them all, so it might be useful to make VersionManager capable of handling this to make life easier in the future.

eschwab:
This could be added into VersionManager. I would propose to have the script check the copyright on each run by having the current year included with the version in `.gregorio-version` Then it will automatically update the copyright year in all the files the first time a release is made in a new year. Unless it is better to manually invoke the updating of the year (which one would have to remember to do).

rpspringuel:
I don't have a problem with having the VersionManager update the year automatically with each run, but there are a whole lot more files which contain copyright notices than they're are which contain version statements.  It would be wasteful of resources to check every file for both version statements and copyright notices.  The list of files containing copyright notices should be separated from the current list of those with version statements so that each file is only checked for what it actually contains.

eschwab:
Agreed on separate file lists.

To clarify: VersionManager would only check `.gregorio-version` for the year. If the year noted in that file was old, then VersionManager would update all the necessary files with the new year.

rpspringuel:
That makes even more sense than what I was thinking of.

eschwab:
Should this be based on CTAN or develop?

rpspringuel:
This is a ctan compatible update because it doesn't involve a change to the gregorio executable.

Issue #1405, opened 2018-03-12
Multiple README files


rpspringuel:
From the folks at CTAN:

> Please note
     http://mirror.ctan.org/help/ctan/CTAN-upload-addendum.html#uniquefilenames
     As a matter of fact he have here three different files
                 README.md
                 doc/README.md
                 fonts/sources/README.md
     (in previous releases as well; I wonder why I apparently
     didn't mention it earlier).
     (Users often send us _identical_ README.md and doc/README.md
     so that one is tempted to delete the second one because of
     http://mirror.ctan.org/help/ctan/CTAN-upload-addendum.html#noduplicatefiles,
     so you see how dangerous this can be ...)
     My wish would therefore be that you _please_ rename the
     README.ms's in doc/ and fonts/sources/ to sth. slightly
     different.  Thanks in advance!

Issue #1403, opened 2018-02-17
VersionManager and multiple Unreleased sections in CHANGELOG


rpspringuel:
Because of the possibility of different CTAN and manual install releases, it's possible that there would be two "Unreleased" sections in the CHANGELOG.md on the develop branch: one for the changes which can be pushed to `ctan` between TeXLive releases (and which were originally made on the current `release-` branch) and one for changes which cannot (and which were originally made on the `develop` branch.

We should probably differentiate these sections (perhaps by changing the second "unreleased" to either "ctan" or "develop", as appropriate).

Additionally `VersionManager.py` needs to have its handling of this situation improved.

rpspringuel:
I've done the rearranging of the CHANGELOG sections on the ctan and develop branches as part of the 5.1.0 release.  `VersionManager.py` still needs to be modified to account for this.

PR #1410, opened 2018-03-15
CTAN related fixes/upgrades

rpspringuel:
The original commit is so that `make maintainer-clean` will remove `gregoriotex.ctan.zip` if it is present.


rpspringuel:
Rather than open a new PR, I'm adding some other CTAN related changes to this one.  I expect that I'll release 5.1.1 shortly after those changes are made (even though, so far at least, they are purely developer or documentation changes).

rpspringuel:
Just addressed #1405 in the simplest possible manner by adding the folder name to the name of the file too.  That is `fonts/README.md` -> `fonts/fonts_README.md` etc.  Let me know if this isn't an acceptable solution.

rpspringuel:
Modifications to VersionManager to address #1403 have now been uploaded.

henryso:
This looks OK to me.  I'd merge it, but I don't know if you're done.

rpspringuel:
Looking at the issues I've assigned to 5.1.1, there's only #1406 and #1362 to deal with.  @eschwab has assigned a separate PR in #1411 for the former, the latter though no one has touched yet.  Given the additional work involved in that issue, it probably makes sense to give it its own PR, so I would say this is ready for merge.

rpspringuel:
In reviewing #1411 I found a couple other files which I think that `make maintainer-clean` ought to remove.  Please take a look and confirm.

henryso:
I think it's fine for maintainer-clean to clean those.  Just make sure that distcheck, et al, still work correctly.

rpspringuel:
Thanks for pointing that out.  When I ran `make distcheck` and `make ctan` I found a couple problems that I've now fixed (and tried to generalize them to fix all targets at the same time).  I think I've checked all the primary targets now, but may have missed some of the more obscure ones (being less familiar with make than some).  If you have a list of particular targets that you'd like checked, please let me know.

rpspringuel:
Okay, I could probably continue to piddle around with this forever so I'm going to force myself to stop here.  If there are no objections, I'll merge this on Saturday and then will start working on the 5.1.1 release on Sunday or when #1411 is merged (which ever comes second).

PR #1411, opened 2018-03-16
Fix 1406 - Copyright updating

eschwab:
Version Manager now automatically updates the copyright year of all the Gregorio Project owned files. There may be a missing file here or there. Should not conflict with changes in #1410 

Ready for review.


rpspringuel:
When files are first created, they sometimes are given a single copyright year instead of a range (i.e. `(C) 2018` instead of `(C) 2018-2018`).  You probably ought to check for that and append a hyphen and the current year instead of simply replacing the final year.  That should be as simple as looking for the hyphen in a line identified as containing a copyright and doing `re.sub(fileyear, fileyear+'-'+CURRENTYEAR, line)` instead of `re.sub(fileyear, CURRENTYEAR, line)`

VersionManager.py
```diff
@@ -64,6 +66,104 @@
...
+                   "contrib/Makefile.am",
+                   "contrib/gabc.lang",
+                   "contrib/gprocess",
+                   "contrib/checkSyllabation.py",
+                   "configure.ac",
+                   "examples/Makefile.am",
+                   "Makefile.am",
+                   "install.sh",
+                   "debian/copyright",
+                   "debian/copyright",
```
rpspringuel: Any reason to have this repeated?

rpspringuel:
Hmm...  Based on `git clean -nx`, at least some of those files that I listed may be generated files:

 - aclocal.m4
 - m4/libtool.m4
 - m4/ltoptions.m4
 - m4/ltsugar.m4
 - m4/ltversion.m4
 - m4/lt~obsolete.m4

If that's the case, then they obviously shouldn't be listed.

VersionManager.py
eschwab: When I made the initial list of files grep found two locations in this file and returned the filename twice. I did not scan the list for duplicates. Removed.

eschwab:
Here is a short annotated version of @rpspringuel 's list of files.

FILE - reason
```
m4/00gnulib.m4 - Free Software Foundation (exclude)
m4/canonicalize.m4 - Free Software Foundation (exclude)
m4/double-slash-root.m4 - Free Software Foundation (exclude)
m4/nocrash.m4 - Free Software Foundation (exclude)
m4/pkg.m4 - not ours (exclude)
m4/ax_check_compile_flag.m4 - not ours (exclude)
m4/ax_check_link_flag.m4 - not ours (exclude)

aclocal.m4 - git clean (exclude)
m4/libtool.m4 - git clean (exclude)
m4/lt~obsolete.m4 - git clean (exclude)
m4/ltoptions.m4 - git clean (exclude)
m4/ltsugar.m4 - git clean (exclude)
m4/ltversion.m4 - git clean (exclude)

windows/gregorio-resources.rc - reference to another file (exclude)
CONTRIBUTORS.md - only the word appears, not a year (exclude)

debian/manpage.xml - added comment to the proper line (added)

fonts/greextra.sfd - Matthew Spencer
contrib/gabc-syntax.plist - Tomasz Grabowski
```
Should we update the copyrights for the last two files?

rpspringuel:
Thanks for looking into all that.  I didn't have time to do much more than generate the list when I was looking earlier.

greextra.sfd has had some pretty major modifications over the years (in particular gresym.sfd was merged into it in 2015) so I'm going to say that this one should have the Copyright updated.  It also looks like the current copyright in the file might be due to a copy/paste error anyway, as it refers to greciliae.  The font first appeared when mensural et al. clefs and custos were added in 2010. 
 @eroux, can you clarify the original history of this font?  In particular, did Matthew Spencer have anything to do with those original glyphs (i.e. only the font name in the copyright statement is in error) or was the whole statement accidently copied from greciliae?

gabc-syntax.plist: looking at the history, this one hasn't been changed since it's original commit so probably leave it off.

eschwab:
Added the check for a single copyright year and append the currentyear as a range.

I also changed the copyright name of greextra and updated the year. The question of whether that copyright statement is accurate still remains to be resolved. For now it is not included in the list of copyright_files.

rpspringuel:
No time to test right now, but the check for range looks good.

The change to greextra, however, doesn't read right to me, but without more accurate information from @eroux I cannot figure out how to fix it.

I'll test tomorrow (Saturday) and if there are no problems go ahead and merge even with greextra the way it is.  If that happens I'll open an Issue to track the fix for it.  That should set me up for releasing 5.1.1 sometime during Holy Week (and in time for the end of the TeX Live pretest, which is scheduled for Apr 2).

rpspringuel:
Not having push access to your fork of the repository, I'll resolve these conflicts locally on my machine and push them directly to the repository rather than wait you do it so that this can be merged online.

PR #1415, opened 2018-03-25
Reverse1363

rpspringuel:
Obviously the biggest issue with this PR is that it creates a compatibility problem with TeX Live 2017.  Earlier versions will work, as will TeX Live 2018.  Is it a good idea to have that kind of hole in the support?  The change is not needed to make TeX Live 2018 work, its just a bit cleaner code.

In any event, I don't think this should be merged until after TeX Live 2018 is released (sometime this summer).


henryso:
In my opinion, this PR should be abandoned.  The present method works in all versions of TeX Live, and unless it causes problems, we should leave it be.

rpspringuel:
I don't have a problem with that.  Now that the PR exists, even if closed, we can resurrect it should the work-around ever become problematic.

henryso:
Closing since there are no objections.

PR #1420, opened 2018-04-13
Rewrite Scribus tool

rpspringuel:
This is a rewrite of the Scribus tool to make it simpler.  Instead of Scribus writing a lua script and calling texlua on it to write two files (the gabc and tex file) and then run lualatex on the result, I make use of the filecontents package to make the writing of the gabc file part of what calling lualatex on the tex file does.  Thus Scribus writes the tex file and calls lualatex on it directly.
Tested as working on both Ubuntu and macOS using Scribus 1.4.6.


eschwab:
Tested and works on Arch using Scribus 1.4.6

Issue #1416, opened 2018-03-26
File names with quotes cause executable not found errors


rpspringuel:
If the file name of the main project file contains a single `'` (such as might happen if the file name contains a single possessive noun), then something goes wrong and GregorioTeX is unable to find the executable (regardless of whether `--shell-escape` is needed/active or not).  I suspect that the quote in the filename creates a quote balancing error in something which gets passed to the shell, but I haven't had the opportunity to track it down.

rpspringuel:
Okay, I think what's happening is that the `...test.gsnippet` file takes its name from the main project file and that file name is not protected the same way that gabc files of actual scores are when we attempt to run gregorio on them.  The result is that the quote messes up the shell command.  We should probably modify the filename for the `...test.gsnippet` so that is it protected in the same way that we do for the gabc files.

rpspringuel:
I just did some testing and my diagnosis was wrong.  Even with the `test_snippet_filename` and `snippet_filename` protected in the same way we protect the filenames of gabc files, the problem still occurs.

PR #1426, opened 2018-06-20
Protect filenames

rpspringuel:
I've created a function for protecting file names (substituting `-` for a bunch of invalid characters) and used it to protect snippet related files.  This prevents projects which have these characters in the main project file's name from not finding the executable (#1416).


rpspringuel:
Just realized that I haven't created a test for this issue.  I'll work on that later today.

rpspringuel:
Problem with creating a possible test.  Adding a single quote to a file name raises the following error from the testing script: `gxargs: unmatched single quote; by default quotes are special to xargs unless you use the -0 option`

@henryso do you have any suggestions for how we might deal with this in the script?

rpspringuel:
I've been thinking about this and since the problem is in trying to mix this unusual filename with `xargs` perhaps we should run this test without the benefit of `xargs`?  Perhaps as an optional "extra" at the end?

henryso:
Perhaps the harness can be rewritten to use the null separator so that `-0` can work.  I haven't thought about how difficult that would be to do, though.

henryso:
I think I've figured out a way to make this work.  However, I think it will require a change to the viewer settings in gregorio-test.rc.  Also, it'll take a bit more doing before I'm done.

henryso:
I have something that works, but I'd like to wait for #1429 to be merged so that it's easier to make the changes I need to make to `harness.sh`.

henryso:
On second thought, looks like I'd need to get this against the ctan branch.  I'll try to rework it tomorrow.

henryso:
I slept on it and decided I could just bring the changes in wholesale.  @rpspringuel Please try your tests with this code, and merge gregorio-project/gregorio-test#328 if it is satisfactory or provide feedback for me to fix.

henryso:
This PR is getting a bit difficult for me to follow, but I think as long as the tests work (which they seem to do) and this change works on Windows as well, then we're good to go ahead with it.

rpspringuel:
This PR is the one that would be closed without merging.  #1450 is I think a simpler and therefore better fix. 

henryso:
OK, I'll put my comment there.

Issue #1423, opened 2018-05-04
`latin` is not considered a valid alias for `Latin`


rpspringuel:
In the gabc files, if one puts `latin` as the language header, then gregorio raises an error about being unable to find the language and falling back onto the default `Latin`.  Other languages don't have this problem because we establish aliases for them in the `gregorio-vowels.dat`.  The `Latin` rules, however, are hard-coded into the source so I think the alias needs to be established there as well.

eroux:
I'm not sure making the change now is relevant, but if I were to do this part again, I would use [BCP47](https://en.wikipedia.org/wiki/IETF_language_tag) lang tags instead of names...

rpspringuel:
Do you mean somehow embedding the tag in document meta data, or using the tag as the contents of the language header?

If the latter, it's easy enough to define aliases for the those tags.  One could even go so far as to modify the documentation to indicate a preference for those tags.  So long as the names currently defined continued to exist as aliases, there would be no backwards compatibility issues.

rpspringuel:
Question for those more knowledgeable in C than I:

Right now the [code](https://github.com/gregorio-project/gregorio/blob/master/src/characters.c#L105-L121) which sets checks for the `Latin` language is as follows

```
void gregorio_set_centering_language(char *const language)
{
    if (!read_vowel_rules(language)) {
        if (strcmp(language, "Latin") != 0) {
            gregorio_messagef("gregorio_set_centering_language",
                    VERBOSITY_WARNING, 0, _("unable to read vowel files for "
                        "%s; defaulting to Latin rules"), language);
        }

        gregorio_vowel_tables_init();
        gregorio_vowel_table_add(DEFAULT_VOWELS);
        gregorio_prefix_table_add("i");
        gregorio_prefix_table_add("I");
        gregorio_prefix_table_add("u");
        gregorio_prefix_table_add("U");
    }
}
```

If I read this correctly, then the line which hard codes for `Latin` as the language name is line [108](https://github.com/gregorio-project/gregorio/blob/master/src/characters.c#L108).  To fix this issue, I think changing `strcmp` to `strcasecmp` would do it, but that is only supported on POSIX compliant systems (i.e. not Windows).  Is this is a compile time thing or a runtime thing?  If it's compile time, then I think that we can make that change because we compile our Windows executables on a Linux cross compiler (which should be POSIX compliant).

[...]

PR #1427, opened 2018-06-20
Language

rpspringuel:
Addresses #1423, both the original issue and the suggestion that @eroux made about language codes.

Note that I've separated the commits for the two possibilities so that the latter can be reversed if we decide against it.


rpspringuel:
Since this issue only ever raised a warning, I don't think it's possible to create a test for it.

PR #1431, opened 2018-09-19
Correct GitHub URL.

EdwardBetts:


rpspringuel:
Thanks for the catch.

Issue #1395, opened 2017-12-13
Not finding scores in TEXINPUTS


rpspringuel:
Gregorio seems not to be looking down the TEXINPUTS path for scores.  I don't have time to create a MWE right now, but need to put this reminder down so that I come back to this when I get the chance.

rpspringuel:
Along similar lines (or as a possible alternative) it might be useful to have a `\grescorepath` function which works similar to `\graphicspath`.

Issue #1393, opened 2017-12-04
Errors when compiling from another directory


jperon:
Here's an example: if I am in the `/home/user` folder, then compile the `gabc. tex` document that is in `/tmp/test`: `lualatex -shell-escape -output-directory=/tmp/test /tmp/test/gabc.tex`

- if this document contains a command `\gregorioscore`, I get the error `The file gabc-score.gabc does not exist`;
- if it contains a command `\gabcsnippet`, the message obtained is `error: kpse prohibits write to file /tmp/test/gabc.gtmp`.

I could of course chdir into the document's folder, but I encounter this problem in trying to develop a [filter for pandoc](https://github.com/jperon/luapandoc-filters/blob/master/filters/gabc.lua), and pandoc doesn't chdir before calling `lualatex`.

Here is a MWE to test:

```
\documentclass{article}
\usepackage[autocompile]{gregoriotex}
\begin{document}
\gabcsnippet{(c4) A(f)ve.(c)}
\gregorioscore{gabc-score}
\end{document}
```

rpspringuel:
The first error sounds like a conflict between the working and relative file path.  GregorioTeX assumes that relative file paths are relative to the working directory and that this is the same as the directory of the file being compiled.  I don't know if this assumption can be broken (it depends on how it is linked to the other error).

The second error is a kpathsea error and is thus related to file security.  Kpathsea won't let you write to arbitrary directories in order to prevent a malicious document from making unintended changes to your system.  We can't change that, but you might be able to disable that security restriction with an option to the luatex call or an appropriate conf file. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

jperon:
I can confirm that, when setting the environment variable `openout_any=a`, the first error disappears ; the second doesn't with the MWE, but as pandoc "stays" in the original folder, it finds the score. So as is, it works.

Just in case ; wouldn't it be possible to detect when kpse prohibits to write to `output-directory`, and then default to a subfolder of current directory (that would be writeable) to avoid failing compilation ?

rpspringuel:
I would have expected setting `openout_any` to eliminate the second error, not the first.  My understanding of what is going on my be flawed. 

[...]

PR #1440, opened 2018-12-05
Kpse file search

rpspringuel:
When looking for files, GregorioTeX will now employ kpse.find_file if it does not find the file is not in the expected location.  This allows for TEXINPUTS to specify paths to scores and addresses the bug part of #1395.

The one thing I'm not sure about is that I make no attempt to strip off directories before doing the kpse based search.  Should I?


rpspringuel:
Note, I have not yet run a complete battery of tests.

rpspringuel:
I've done some further testing and have things working a bit better.  However, I'm running into #1393.  I've included a patch here for that issue when kpse is used to find the file, but it would probably be better to figure out a more general solution.

rpspringuel:
Just discovered an interesting quirk: `kpse.find_file` appears capable of traversing directories with spaces in the name while the `\input` statement which actually reads in the gtex file is not.  This results in an interesting bug where the code finds a gtex file which is ready for importing, but cannot actually read it in.  Not sure if this is simply something that needs documenting or some sort of code fix.

henryso:
Is there a way to read a file in a la `\input` which does work with spaces?  If so, then I vote we use it; if not, it's a TeX quirk, and we document it.

rpspringuel:
The best that I can think of is to read in the file in lua using `io.open` and then `tex.sprint` line by line.

henryso:
That seems a bit too much.  My vote is we document it and if there are enough requests for this, then we consider alternatives.

eroux:
I agree with @henryso , let's not "fix TeX"...

rpspringuel:
Okay, this now appears to be complete and behaving the way I want it to.  Please check it carefully as I cannot currently check the kpse search in the test repository.

henryso:
I *don't* think my review is good enough to fully accept this, but it looks OK to me.

henryso:
This has been sitting here for a while with no comments to the contrary, so let's merge.

PR #1448, opened 2019-03-23
Update for Lua 5.3

rpspringuel:
Turns out there is an integer/float issue that needed to be dealt with.  This change works under both TeX Live 2018 and TeX Live 2019.
This change does not affect any tests.


rpspringuel:
TeX Live is looking to freeze the Pretest as soon as Monday.  I will merge this over the weekend and publish a bugfix release if there are no objections.

henryso:
I have no objection.

eroux:
lgtm!

PR #1449, opened 2019-03-24
Enumerate ancient font options

rpspringuel:
Generally when there is a limited number of options, we enumerate them.  I've therefore listed the possible ancient fonts.  Technically a user could add their own, but I'm not sure that's realistic.


jakubjelinek:
LGTM

PR #1450, opened 2019-03-30
Simpler fix for filenames

rpspringuel:
This fix for problematic filenames seems to work and is simpler than what I was working on in #1426.
It is compatible with gregorio-project/gregorio-test#328 after the changes I proposed on henryso/gregorio-test#1 are pulled in.
The corresponding PR which adds the test for this problem is gregorio-project/gregorio-test#331

If this is approved, #1426 can be closed without merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][CTAN]
+### Fixed
+- Project names with quotes in them no longer cause GregorioTeX to not be able to find the executable.  See [#1416](https://github.com/gregorio-project/gregorio/issues/1416).
 
 
 ## [5.2.0] - 2019-03-10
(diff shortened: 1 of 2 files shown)


henryso:
As long as this works on Windows, I'm OK with it.

rpspringuel:
I haven't checked that yet.  Thanks for the reminder.

rpspringuel:
Found an issue here.  Quoting the executable name fouls up the security restrictions:  `gregorio` is an allowed executable, but `"gregorio"` is not (when `shell_escape=p`).  I'm going to need to change that up and do some more testing.

henryso:
Maybe just remove the escaping of the executable and document that gregorio shouldn't be installed in a directory with a space in it?

rpspringuel:
Well, the executable never gets called with its full path anyway.  We rely on it being in `PATH`.  So I think just removing the quotes around it should be sufficient. 

rpspringuel:
Just tested on Windows (with f0b263) and stuff seems to be working fine.

rpspringuel:
TeX Live is looking to freeze the Pretest as soon as Monday.  I'm planning to merge #1448 over the weekend to make sure that fix is in TeX Live 2019.  Does anyone think this should be merged too?

henryso:
I think we should go ahead with this.

Issue #1457, opened 2019-04-17
Scribus having trouble with gabc headers in 5.2.1


domoknight:
The score is rendering in Scribus, but the gabc headers appear to be ignored. 

Using the sample gabc:

> name: myscore;
> mode: 6;
> commentary: howdy;
> %%
> (c3) Pó(eh/hi)pu(h)lus(h) Si(hi)on,(hgh.)

No mode number is shown.  Also, even though I'm using
\gresetheadercapture{commentary}{grecommentary[3mm]}{string}
no commentary is printed.

An excerpt from the logfile is as follows:

```
Module gregoriotex Info: capturing header commentary using grecommentary[3mm], s
tring=true, name=nil on input line 24
Module gregoriotex Info: scribus_file-score.gabc has been modified and scribus_f
ile-score-5_2_1.gtex needs to be updated. Recompiling the gabc file. on input li
ne 25
Module gregoriotex Info: compiling the score scribus_file-score.gabc... on input
 line 25
Module gregoriotex Info: will use gregorio-5_2_1 on input line 25

Module gregoriotex Warning: warning:no name specified, put `name:...;' at the be
ginning of the file, can be dangerous with some output formats on input line 25
Module gregoriotex Warning: *** end of warnings for scribus_file-score.gabc *** 
on input line 25
```

domoknight:
This appears to be related to the rewrite of the Scribus tool (900_gregorio.xml).  Somewhere along the way, the gabc header info is not getting passed along.  If I use the 900_gregorio.xml file from the 5.1.1 build, the header data passes through just fine. (annotation text appears and gresetheadercapture works)

rpspringuel:
Thanks for the info.  I’ll have to look into that, but it may not be until after Easter. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rpspringuel:
I've had a chance to do some additional investigation, and the problem appears to be the use of the `filecontents` environment.  This environment adds a comment header which uses `%%` to mark itself and therefore fouls up the executable's ability to detect the proper header (we use `%%` to mark the transition from headers to score).  The correct environment to use is `filecontents*` which doesn't add the comment header.  However, I haven't had the chance to fully test this because of other problems I'm having with my installation.  If someone else can test and verify the diagnosis, then I'll put up the PR with a fix.  Otherwise this will have to wait until I fully sort out the problem with my system.

domoknight:
[...]

PR #1459, opened 2019-04-22
Fixes to Scribus tool

rpspringuel:
The use of `filecontents` rather than `filecontents*` was leading to a comment header that made it impossible for Gregorio to find the gabc headers in the temporary score file.
Some of the indenting (which makes the file more human readable) was leading to errors in the formatting of the created files because they are processed in a way which handles whitespace differently from XML.

This is not a file which is uploaded to CTAN so I've made this PR against develop (for inclusion in the next enhancement release).  However, since the file is not uploaded to CTAN, this commit could be cherry-picked if the next release turns out to be a bugfix rather than an enhancement.  Some rewriting of the CHANGELOG would be needed in that event.

See #1457.


rpspringuel:
Barring objections I'll merge this later this week.

PR #1464, opened 2019-05-21
Stafflinethickness

### LINKED ISSUE #1461: problem updating \gre@stafflinefactor
This was happening on 5.2.1

If \greconffactor is different than what the staff size is set to, it seems to cause a problem when loading the new space file.  It multiplies by the staff size but divides by \greconffactor.  Anyway I think that is what was going on.  So if you have the staff size set to something big like 23 and the regular \greconffactor of 17, when you load the new spacing, \gre@stafflinefactor gets set to 31  (23 * 23 / 17).  It's easy to workaround so it's not a big deal.  I'm not sure if anything else is affected, but the line thickness was certainly quite noticeable.

rpspringuel:
Fixes #1461 

Staff line thickness scales with the size of the staff, but was not being set in gsp-default.tex.  As a result, whenever the default spacings were loaded and the staff size wasn't currently set to the default, the line thickness would change.  This change puts the setting of the staff line thickness into gsp-default so that it will scale as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][CTAN]
+### Fixed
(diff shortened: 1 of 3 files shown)


rpspringuel:
This has been checked out by asfbloom so I'm merging.

Issue #1444, opened 2019-03-03
No ligatures in syllable with the style tags using LuaLaTeX


rpspringuel:
As reported [here](https://groups.google.com/d/msg/gregorio-users/pSx6H-I-1oc/7gNGhD2FAwAJ) by Sr. Maria Ruth:

> In GregorioTeX 5.1.1, when a style tag is used just for a part of a syllable, part that corresponds to a ligature, LuaLaTeX never typesets the ligature [liga_lualatex_1], while LuaTeX does, with the exception of `<c>` and `<ul>` [liga_luatex].  In "gregoriotex.sty", the macros `\GreItalic`, `\GreSmallCaps`, `\GreBold`, and `\GreTypewriter` use `\textit`, `\textsc`, `\textbf`, and `\texttt` in their definition.  I think those LaTeX's control sequences should be replaced by `\itshape`, `\scshape`, `\bfseries`, and `\ttfamily` because they cut off (they armour) each character in the ligature, while the last ones do not [liga_lualatex_2].
>
> Anyway I think the overall procedure which manages the style tags within a syllable should be improved.

[liga_issues.zip](https://github.com/gregorio-project/gregorio/files/2923776/liga_issues.zip)

rpspringuel:
The proposed fix works for bold and italics.  The sample for small capitals and typewriter is inconclusive because no ligature is involved.

`\underline` does not have a flag equivalent in normal LaTeX.  To fix it for this case we would need to test the various packages which improve underlining and add that dependency.

Color is already a switch, but its doing something that already breaks the ligature.  Based on [this answer](https://tex.stackexchange.com/a/491988/49327) it appears that we need to switch from xcolor to luacolor to make this work.  Is switching this dependency acceptable?

rpspringuel:
Correction.  We would need to load luacolor *in addition* to xcolor, not instead of it.

rpspringuel:
Additional correction: the luacolor solution only works for LuaLaTeX.  The package doesn't appear to be compatible with LuaTeX, so I'm looking for an additional option.

rpspringuel:
After further reflection, I cannot come up with a situation where small capitals or typewriter would actually make use of ligatures.  As a result, while will make the change according to the pattern for the ones which are problematic, I'm not going to worry about testing it at this point.

[...]

PR #1466, opened 2019-05-30
Ligatures

rpspringuel:
Corrected interaction issues between text styles and ligatures.  The LaTeX commands like `\textit` insert italics correction, preventing ligatures being formed between their arguments if two occur sequentially.  On the other hand the switches like `\itshape` do not.  Since gregorio breaks up syllables around the vowel and then applies the formating commands to each part, this behavior showed up.  We switch to using the switches to avoid this (except for underlining, for which a switch does not exist).  We also fix this for color tags by loading `luacolor` which changes how the `\color` tag is implemented to allow ligatures to span groups.  See [#1444](https://github.com/gregorio-project/gregorio/issues/1444).


eroux:
looks good to me!

Issue #1460, opened 2019-05-03
gsp-default.tex does not include a version check


BillyV2018:
Gregoriotex 5.2 in TexLive 2019, checked latest updates of May 3 at 10 AM.

GreSet variables work okay, except the \grechangestaffsize. According to documentation defaults to 17. I want to reduce it to 15. That provokes an error 'undefined control sequence'.
Adding 'pt' is not a solution. Leaving it to default is the workaround for the moment.
The relevant paragraphs from the error log:

! Undefined control sequence.
\gre@changedimenfactor ...ifgre@scale@bar@minimis 
                                                  \gre@changeonedimenfactor ...
l.22 \grechangestaffsize{15}
                           % 17 is de standaard maat
The control sequence at the end of the top line
of your error message was never \def'ed. If you have
misspelled it (e.g., `\hobx'), type `I' and the correct
spelling (e.g., `I\hbox'). Otherwise just continue,
and I'll forget about whatever was undefined.

! Missing number, treated as zero.
<to be read again> 
\gre@space@skip@bar@minimis 
l.22 \grechangestaffsize{15}
                           % 17 is de standaard maat
A number should have been here; I inserted `0'.
(If you can't figure out why I needed to see a number,
look up `weird error' in the index to The TeXbook.)

! Illegal unit of measure (pt inserted).
<to be read again> 
\gre@space@skip@bar@minimis 
l.22 \grechangestaffsize{15}
                           % 17 is de standaard maat
Dimensions can be in units of em, ex, in, pt, pc,
cm, mm, dd, cc, nd, nc, bp, or sp; but yours is a new one!
I'll assume that you meant to say pt, for printer's points.
To recover gracefully from this error, it's best to
delete the erroneous units; e.g., type `2' to delete
two letters. (See Chapter 27 of The TeXbook.)

! Missing ) inserted for expression.
<to be read again> 
\gre@space@skip@bar@minimis 
l.22 \grechangestaffsize{15}
                           % 17 is de standaard maat
I was expecting to see `+', `-', `*', `/', or `)'. Didn't.

! LaTeX Error: Missing \begin{document}.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.22 \grechangestaffsize{15}
                           % 17 is de standaard maat
You're in trouble here.  Try typing  <return>  to proceed.
If that doesn't work, type  X <return>  to quit.

[...]

PR #1467, opened 2019-05-30
Gsp default

rpspringuel:
Addresses underlying cause of #1460 

This PR moves `gsp-default.tex` to be an internal file.  This allows for us to version check it to make sure we're loading one which properly defines all needed distances and penalties.  As part of this `\grecreatedim` is moved to the internal namespace as `\gre@createdim`.

`gsp-sample.tex` is now provided in the doc folder as an example space configuration.  Right now it's simply a duplicate of `gsp-default.tex` except for the following:

1) It has no version check (version checking is only for internal files)
2) It uses `\grechangedim` (an User macro) instead of `\gre@createdim` (an internal macro)
3) The comments have been edited to make them slightly more user friendly (please double check these)


rpspringuel:
Oh, I should note here:  **This is a user interface change and warrants a bump in the major version number when released.**

henryso:
If this will be an internal file, maybe it should get a name change as well?  `gregoriotex-sp-default.tex` perhaps?

rpspringuel:
That'll make it harder for people to reload the defaults after changing to a custom space.  As is `\greloadspaceconf{default}` will reload the default spacing configurations, if we change the name, then that won't work.

Oh... That does make me realize something though, because rescaling is handled within `\grechangedim` (which is not used in `gsp-default.tex`) we're going to run into a scaling problem when reloading the defaults if the current staff size is not 17.  I'm going to need to add some special handling for loading the defaults to account for that.

Now that I think about it, I guess that makes it relatively easy to change the name.  If there's already some special handling of the defaults, that special handling can be extended to the file name too.

henryso:
I merged the test by accident, so I'm going to merge this as well.  The name change bit can be a separate pull request.

Issue #1261, opened 2016-12-06
Font size not taken into account when spacing lyrics


rpspringuel:
In the following document the larger than normal lyrics text causes it to overlap with the low notes of the staff.  

```
% !TEX program = LuaLaTeX+se
\documentclass[12pt]{article}

\usepackage{fullpage}
\usepackage{libertine}
\usepackage{gregoriotex}

\begin{document}

\Large
\gabcsnippet{(c3) Hel(g)lo(g) nó(dc)}

\end{document}
```

This happens because the distance between the bottom line and the text baseline (`spacelinestext`) is a fixed value (currently 0.60617cm) which is only appropriate for font sizes of roughly 10 to 12 pt.  I propose changing the value to em units.  The current value is about 1.72em if the base font is 10pt and 1.44em if the base font is 12pt.  I thus think 1.5em is a good starting point.  I would also change this distance so that by default it is fixed with respect to the staff size.

Comments?

eroux:
Well, I don't have a strong opinion on that, I don't find the current behavior particularly buggy... but if you change it, the `em` unit doesn't make any sense, the `ex` unit does

cantatedomino:
I do very much appreciate the hard work of all the Gregorio developers, and the updates in Version 6, but this change did break the spacing in my engravings; Setting the Salve Regina (solemn tone) shifted the text in all the lines of chant to spacing required by the lowest note in the entire piece, rather than the lowest note for that particular line.I have a work-around that adds \grechangedim{spacelinestext}{0.60617 cm}{scalable} to my LaTeX preprocessor file.

PR #1469, opened 2019-06-12
Lyric fontsize

rpspringuel:
Note: this includes #1467 because it would conflict with it otherwise.  This is only a spacing change (#1261), and so could be included in ctan, but given the above, and that it's really such a minor issue, I've pulled it against develop instead.


henryso:
This looks OK to merge to me.  In your opinion, how "breaking" is this to people in the middle of a project?  In other words, if you are in the middle of the project, would you want to be cautious when updating?  If so, then this is probably left out of CTAN until a release with a proper UPGRADE.md entry or at least better warning.

rpspringuel:
If the project contains only scores at size 17 (the default) or are using their own spacing configurations, then the change isn't breaking at all.  For any project which contains scores at other sizes but still use the default spacings (just scaled) then this will make a difference.  I think the difference is an improvement, but others might disagree.  If you look at some of the ancillary tests which changed you can get a sense for how much of a difference the change makes.

This is already against develop, and thus as is won't go to CTAN until next year.  To push it to CTAN earlier would require me to redo the PR (and not include #1467).  I can add an UPGRADE.md entry for it.

henryso:
In that case, I have no issues with this.

Issue #1475, opened 2019-09-12
Parenthesized divisio minima


henryso:
A request from the mailing list.

It would be nice to have the ability to typeset a parenthesized divisio minima like you see in the Graduale Triplex for when Solesmes edited "out" the marking.

I think it would be also good to have support parenthesized virgula as well with a similar syntax.

henryso:
@eroux @rpspringuel Any thoughts on the notation?

rpspringuel:
I don’t see a potential clash on the use of `?`.  The parenthesis say “optional” to me much as the dotted divisio maior does.  

✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️
Fr. Samuel, OSB
(R. Padraic Springuel)
St. Anselm’s Abbey 
4501 South Dakota Ave, NE
Washington, DC, 20017
202-269-2300
(c) 202-853-7036

PAX ☧ ΧΡΙΣΤΟΣ

henryso:
The challenge now is to draw nice-looking parentheses.  I've tried a little, but mine don't yet look good.  I will keep at it.

henryso:
I started the coding for this, and I need to modify how `#7` of `\GreSyllable` and `\GreBarSyllable`.  In tracing through the code, I see constructs like:

```
\ifnum#7>9 %
  \ifnum#7<20 %
    % do something
  \else %
     % do something else
  \fi %
\fi
```

Best I can tell, `#7` will be 20/40/60 + some `AT_` constant as per the `gregorio` executable.  The `AT_` constants range from 0 through 9, so `\ifnum#7>9` is looking for an alteration.  However, the code in `\ifnum#7<20` will never be executed in this construct.  Am I missing something?

@eroux @rpspringuel Any background here?

henryso:
@eroux @rpspringuel Disregard.  I found it after looking at it with less sleepy eyes.  Values between 9 and 20 are bars.

PR #1476, opened 2019-10-31
Added parenthesized virgula, divisio minima, flat, natural, and sharp.

henryso:
Implements #1475.

There are also a number of episema-related bug fixes that I found and fixed incidentally.


henryso:
@rpspringuel Everything seems to be working now, so this is ready for review.

PR #1480, opened 2019-12-03
Gsp default

rpspringuel:
Fixes problems with #1467 (which was merged early).

Also merges in changes on ctan branch.


henryso:
The changes are unfortunately really hard to review due to all the merging, so I think I will just merge this and hope for the best.

PR #1483, opened 2020-01-02
Simplify example file

adunning:
Use KOMA-Script article class to achieve the intended appearance with fewer additional packages. Improve font loading with Libertinus package.

### DIFF
--- a/examples/main-lualatex.tex
+++ b/examples/main-lualatex.tex
@@ -5,7 +5,7 @@
 \documentclass{scrartcl} % set document class: manual at https://ctan.org/pkg/koma-script
 
 % Load packages:
-\usepackage[osf,p]{libertinus} % set font
+\usepackage[osf,p]{libertine} % set font
 \usepackage{gregoriotex} % enable Gregorio score inclusion
 \usepackage[latin]{babel} % set language
 
--- a/examples/main-lualatex.tex
+++ b/examples/main-lualatex.tex
@@ -12,10 +12,6 @@
 \setkomafont{section}{\normalfont\centering\huge\scshape} % section heading style
 \setcounter{secnumdepth}{-\maxdimen} % remove section numbering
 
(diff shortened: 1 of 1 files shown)


henryso:
This is a pretty opinionated change.  As I have no opinion really on this (I prefer other packages entirely myself), @rpspringuel, what do you think?

adunning:
The resulting appearance is very similar – this only aims to simplify the setup, to allow new users to focus on Gregorio itself.

rpspringuel:
Personally, I don't consider the KOMA-Script classes to be simplifications.  They're more all-inclusive and put more thing "under the hood" so to speak, which can lead to less appearing in the project file (especially when the defaults are acceptable) but one of the reasons I like using LaTeX is because I like the way it exposes things.  The granularity of packages is a feature, not a bug, for the way I work.

I realize this isn't for everyone, so I too don't have really strong feelings about this.  Perhaps the best compromise would be to have multiple examples which model different possibilities.

adunning:
In teaching LaTeX to new users, I have found that many tend to see a long list of packages that they don't understand, and blindly copy them into every new document, because it vaguely works, but without knowing why.

For this reason, I've found that KOMA-Script is a better place for new people to start, since it has a manual covering most of what one would want when getting started. For advanced users, there's no question that bare-bones packages are often desirable.

I think you have already achieve the aim of giving different models on the [examples page](http://gregorio-project.github.io/gregoriotex/example.html).

adunning:
Are the changes I made to this acceptable, or would you prefer to stick with the `article` class? (Again, my aim here is to make the file as simple as possible to make it clear what Gregorio itself needs.)

rpspringuel:
Based on the absence of any significant objections, these changes seem acceptable.  However, the PR needs to be against the ctan branch.  We don't apply patches directly to master.

PR #1501, opened 2020-12-18
Fix broken link

benyanke:


rpspringuel:
Please pull this against the ctan branch.  We don't apply patches directly to master.

PR #1505, opened 2021-01-26
Make end of element unbreakable by default

### LINKED ISSUE #1504: Unison notes on the same syllable should by default not allow line breaks
Here is a example illustrating the problem:

Screenshot_20210123_211723

The actual behaviour is that this code:

```gabc
%%
(hff) (::)
```
gives following output from `gregorio`:

```tex
% File generated by gregorio 5.2.1
\GregorioTeXAPIVersion{5.2.1}%
\GreBeginHeaders %
\GreEndHeaders %
\GreBeginScore{fdfb0937e15f22d0431f6d53899bc3d9b4e7544f}{10}{8}{0}{0}{}{4}{\GreInitialClefPosition{3}{0}}%
\GreSetLargestClef{c}{3}{3}{c}{0}{3}%
\GreScoreOpening{%
}{%
\GreSetInitialClef{c}{3}{3}{c}{0}{3}{1}%
}{%
}{\GreSyllable}{\GreSetNoFirstSyllableText}{}{\GreUnstyled}{1}{\GreSetNextSyllable{}{}{}{}{}}{}{15}{}{%
\GreSyllableNoteCount{3}%
\GreGlyph{\GreCPFlexusLongqueueTwoNothing}{10}{\GreNextCustos{8}{}}{1}%
{%
\GreGlyphHeights{10}{8}%
}{}{}%
\GreEndOfElement{0}{0}{2}%
\GreGlyph{\GreCPPunctum}{8}{\GreNextCustos{9}{}}{0}%
{%
\GreGlyphHeights{8}{8}%
}{}{}%
}%
%
\GreFinalDivisioFinalis{%
}%
\GreEndScore %
\endinput %
```

IMO, the last `\GreEndOfElement{0}{0}{2}%` should be by default `\GreEndOfElement{0}{1}{2}%`, so that the two notes stay on the same line.

In fact, I think that, within a melism, any breakable space should be explicit (which was the behaviour "once upon a time", but I don't remember when this changed): for example, in `hGFgv`, the `gv` should stay with the `hGF`, unless a cut is added: `hGF/gv`. Perhaps it's more debattable when notes aren't unison ones, but (IMO again) it's really expected for unison ones.

jperon:
This is a proposal to fix #1504 if there is no objection. It restores the old behavior: line breaks are allowed between neumatic elements only if a `/` is typed.

### DIFF
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -4140,7 +4140,7 @@ static void write_syllable(FILE *f, gregorio_syllable *syllable,
                                         && element->next->type == GRE_ALT
                                         && element->next->next->type ==
                                         GRE_ELEMENT))) {
-                    fprintf(f, "\\GreEndOfElement{0}{0}{%u}%%\n",
+                    fprintf(f, "\\GreEndOfElement{0}{1}{%u}%%\n",
                             note_unit_count);
                 }
                 break;


rpspringuel:
This PR needs to be against develop, not master.  It also needs some matching tests in gregorio-project/gregorio-test.

Issue #1495, opened 2020-06-17
Line breaking issue with long word


rpspringuel:
We've got some issues cropping up with long words not triggering line breaking the way they used to.  I haven't had time to investigate, but came across it while working on updating things for TeX Live 2020.  The problem also occurs in TeX Live 2019.

See https://github.com/rpspringuel/gregorio-test/commit/4d6c02e696058cde2c1f85c9a9043c68aa0fa9c8 for tests which show the problem.

rpspringuel:
Possibly #1184 again.

PR #1508, opened 2021-02-01
Saves

rpspringuel:
Fixing problem identified by user: https://groups.google.com/g/gregorio-users/c/u3LmnGYnhwU

We were reusing a register to save and restore the value for the `\hyphenpenalty`.  The register was used both at the whole score level and around a No Line Break Area.  As a result, when we attempted to restore the `\hyphenpenalty` at the end of the score, we ended up setting it to what it was just before the No Line Break Area, not what it was before the score.

This PR does some refactoring of all registers used to save values for restoration so as to not just fix this one problem, but hopefully make it harder to repeat the same mistake in the future.

There's also a couple of minor fixes here and there which were discovered while reviewing the documentation for this fix.

**Note:** the problems identified in #1495 still exist in this PR.


henryso:
I'm fine merging this when ready

rpspringuel:
Another thread on the mailing list which is probably the same problem fixed here: https://groups.google.com/g/gregorio-users/c/yDZsp7255rU/m/7TmEI_43AAAJ

PR #1511, opened 2021-02-03
Visibility

rpspringuel:
Add the ability to hide translation and above lines text (separate switches).  Addresses issue from mailing list: https://groups.google.com/g/gregorio-users/c/J-V0dkBMevY


henryso:
Not that you need my approval, but this looks OK to me.

PR #1513, opened 2021-02-03
Modify vowel message

### LINKED ISSUE #1470: Searching for Latin in gregoriotex-vowels.dat
If the language is `Latin`, `latin`, `lat`, or `la` then is there really a need to search the gregoriotex-vowels.dat files for vowel definition rules?  The only reason I can think of is to allow a user to define alternate vowel definitions for Latin.  Is that really reasonable?

If it is not, then we should modify `characters.c` lines [63-87](https://github.com/gregorio-project/gregorio/blob/master/src/characters.c#L63) so as to skip over this block when the language is Latin.

On the other hand, if it is reasonable that a user might want to define an alternate set of vowel determination rules for Latin, then we should probably add an informational message (or modify the existing ones) so that when the user sees "Could not find Latin..." they don't worry about it.

rpspringuel:
There is no Latin vowel definitions in gregorio-vowels.dat by default because they are built into the binary itself.  The user can, however, define their own alternative, so it is conceivable that there might be.  We thus print a little extra information when were falling back to the internal Latin vowel definitions so that the user isn't confused by the "Could not find" message.

Fixes #1470.

No testing changes.

### DIFF
--- a/src/characters.c
+++ b/src/characters.c
@@ -86,7 +86,7 @@ static bool read_vowel_rules(char *const lang) {
         }
     }
     if ((strcmp(language, "Latin") == 0 || strcmp(language, "latin") == 0 || strcmp(language, "la") == 0 || strcmp(language, "lat") == 0) && status == RFPS_NOT_FOUND) {
-        gregorio_messagef("read_rules", VERBOSITY_INFO, 0, "Falling back on internal Latin");
+        gregorio_messagef("read_rules", VERBOSITY_INFO, 0, "Falling back on internal Latin vowel rules");
     }
     if (status == RFPS_ALIASED) {
         gregorio_messagef("read_rules", VERBOSITY_WARNING, 0,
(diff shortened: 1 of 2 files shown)


src/characters.c
```diff
@@ -85,6 +85,9 @@ static bool read_vowel_rules(char *const lang) {
             break;
         }
     }
+    if ((strcmp(language, "Latin") == 0 || strcmp(language, "latin") == 0 || strcmp(language, "la") == 0 || strcmp(language, "lat") == 0) && status == RFPS_NOT_FOUND) {
+        gregorio_messagef("read_rules", VERBOSITY_INFO, 0, "Falling back on internal Latin");
```
henryso, line 89: Perhaps the message should be "Falling back on internal Latin vowel rules".

henryso (changes requested):

src/characters.c
rpspringuel: Done.

I also modified a message a few lines later to make the two messages more consistent.

henryso (approved):

PR #1524, opened 2021-02-11
Update to scribus configuration file

### LINKED ISSUE #1523: Change to line 51 of 900_gregorio.xml to make it work in Scribus
When using gregorio with the latest version of Scribus and the latest version of TexLive, I needed to change line 51 of 900_gregorio.xml to read

`\begin{filecontents}[force]{scribus_file-score.gabc}`

Without this addition of the **[force]** option, LaTex would refuse to overwrite scribus_file-score.gabc, so that I was only able to make render frames with the default preview score.  It looks like it is due to changes in the way that the LaTex kernel handles the filecontents package.  See here: https://www.latex-project.org/news/latex2e-news/ltnews30.pdf

rpspringuel:
There have been some changes to the filecontents environment that need to be addressed.
Fixes #1523

### DIFF
--- a/contrib/900_gregorio.xml
+++ b/contrib/900_gregorio.xml
@@ -47,10 +47,9 @@
 		    \pagestyle{empty}
     		\setlength{\textwidth}{$scribus_realwidth$ pt}
 
-        \begin[overwrite]{filecontents}{scribus_file-score.gabc}
-    </preamble>
+        \begin{filecontents*}[overwrite]{scribus_file-score.gabc}</preamble>
     <postamble>
-        \end{filecontents}
+        \end{filecontents*}
 
         \begin{document}
         $scribus_greconf$
--- a/contrib/900_gregorio.xml
+++ b/contrib/900_gregorio.xml
@@ -46,7 +46,7 @@
     	\setlength{\textwidth}{$scribus_realwidth$ pt}
 
         \begin{filecontents*}[overwrite]{scribus_file-score.gabc}
-    </preamble>
+</preamble> <!-- this line's indentation is purposefully off because of the way that scribus interprets the preamble section -->
     <postamble>
         \end{filecontents*}
 
--- a/contrib/900_gregorio.xml
+++ b/contrib/900_gregorio.xml
@@ -41,14 +41,13 @@
 		                paperheight=$scribus_realheight$ pt,
           					left=0cm,top=0cm,right=0cm,bottom=0cm,nohead,nofoot]{geometry}
         \usepackage[autocompile]{gregoriotex}
-        \usepackage{filecontents}
     		\title{Scribus-Latex-File}
     		$scribus_additionalgreheaders$
     		\author{Scribus}
 		    \pagestyle{empty}
     		\setlength{\textwidth}{$scribus_realwidth$ pt}
 
-        \begin{filecontents}{scribus_file-score.gabc}
+        \begin[overwrite]{filecontents}{scribus_file-score.gabc}
     </preamble>
     <postamble>
         \end{filecontents}
--- a/contrib/900_gregorio.xml
+++ b/contrib/900_gregorio.xml
@@ -36,18 +36,17 @@
     </empty-frame-text>
     <!-- you can change the basic headers here -->
     <preamble>
-		    \documentclass[$scribus_grefontsize$]{extarticle}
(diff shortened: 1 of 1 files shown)


EagerDSC (commented):
Moving </preamble> up the line does not have the intended effect.  It removes the entire first line from the scribus render frame editor so that gregorio returns the following error `Module gregoriotex Warning: warning:no name specified, put `name:...;' at the b
eginning of the file, can be dangerous with some output formats on input line 2
0`.  I can still work around this by adding the blank line to the beginning of my gabc input in the scribus render frame editor.  Is there a reason that gregorio does not ignore white space not contained between `:` and `;` preceding the `%%` in a `.gabc` file?  (I'm pretty new to LaTex and gregorio.)

rpspringuel:
Can you post a copy of the tex file that Scribus creates?  I'm need to see what the `filecontents*` environment contents end up being to make more educated guesses as to how to make this work.

> Is there a reason that gregorio does not ignore white space not contained between : and ; preceding the %% in a .gabc file? 

The reason is mostly that we never thought about that possibility.  We always assumed that gabc file writing would be done by hand.  Unfortunately, scribus, while using an XML file to define a rendering frame, doesn't really understand the indentation of an XML file and thus sometimes thinks XML indentation should be part of the target file it's supposed to create.  Further, we actually need to write two files (the gabc and the tex file) while the render frame tool only expects to need to write one.  Thus we have to design the tex file to create the second automatically in a way that is compatible both with how Scribus interprets the XML and with how `filecontents` interprets its contents.  Getting such indirection right is complicated (especially since none of the developers actively use Scribus and thus testing requires one of us to relearn it each time we run into problems).

@henryso, Could the lexer be modified to ignore spaces in header names?

EagerDSC:
**If `</preamble>` is up the line in `900_gregorio.xml` then here are the two situations.**

**_Given the default input to the Scribus render frame editor, these are gabc, gtex, and glog files that are generated._**
[DefaultInput.zip](https://github.com/gregorio-project/gregorio/files/5976548/DefaultInput.zip)

**_Given the default input preceded by a blank line, these are the gabc, gtex, and glog files that are generated._**
[WithBlankLineFirst.zip](https://github.com/gregorio-project/gregorio/files/5976558/WithBlankLineFirst.zip)


**If `</preamble>` is down on its own line in `900_gregorio.xml`, then here are the two situations:**

**_Given the default input to the Scribus render frame editor, these are gabc, gtex, and glog files that are generated._**
[DefaultInputPreambleOwnLine.zip](https://github.com/gregorio-project/gregorio/files/5976568/DefaultInputPreambleOwnLine.zip)

**_Given the default input preceded by a blank line, these are the gabc, gtex, and glog files that are generated._**
[WithBlankLinePreambleOwnLine.zip](https://github.com/gregorio-project/gregorio/files/5976569/WithBlankLinePreambleOwnLine.zip)

rpspringuel:
I need to see the **tex** files (not the gtex).  I need to know what scribus itself is writing.  All of these files are produced by LuaTex after scribus has already (mal-)formed the tex file.

EagerOrthodox:
I think these are what you are looking for, despite the fact that they appear to have no file extension.

[WithBlankLinePreambleUpLine.zip](https://github.com/gregorio-project/gregorio/files/5983278/WithBlankLinePreambleUpLine.zip)
[WithBlankLinePreambleOwnLine.zip](https://github.com/gregorio-project/gregorio/files/5983279/WithBlankLinePreambleOwnLine.zip)
[DefaultInputPreambleUpLine.zip](https://github.com/gregorio-project/gregorio/files/5983280/DefaultInputPreambleUpLine.zip)
[DefaultInputPreambleOwnLine.zip](https://github.com/gregorio-project/gregorio/files/5983281/DefaultInputPreambleOwnLine.zip)

rpspringuel:
Okay, one more go.  Based on what those files looked like, I think this should eliminate the need for the extra blank line in the input box.  @EagerDSC @EagerOrthodox Please test and report back.

EagerOrthodox (approved):
This version of the file seems to be working!  Thank you for being so attentive to a minor issue.

EagerDSC (approved):
This version of the file works.  Thank you!

rpspringuel:
Thanks for testing.  Glad we could get it working.

PR #1526, opened 2021-02-12
Name font calibration distances

### LINKED ISSUE #1521: Magic spaces
While working on #1491 I found a magic space and while looking to fix that, I've found a couple of more:

[gregoriotex-signs.tex:1213](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-signs.tex#L1213):`			\gre@dimen@temp@three=\dimexpr(3700sp * \gre@factor)\relax %`
[gregoriotex-signs.tex:1215](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-signs.tex#L1215):`			\gre@dimen@temp@three=\dimexpr(4500sp * \gre@factor)\relax%`
[gregoriotex-signs.tex:1218](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-signs.tex#L1218):`			\gre@dimen@temp@three=\dimexpr(2500sp * \gre@factor)\relax %`
[gregoriotex-spaces.tex:173](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L173):`	\global\gre@dimen@stafflineheight=\dimexpr(1500 sp * \gre@stafflinefactor)\relax %`
[gregoriotex-spaces.tex:182](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L182):`	\global\gre@dimen@interstafflinespace=\dimexpr((31500sp * \gre@factor) %`
[gregoriotex-spaces.tex:220](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L220):`	\global\gre@dimen@constantglyphraise = \dimexpr((-22000 sp * \gre@factor) %`
[gregoriotex-spaces.tex:815](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L815):`	\global\gre@dimen@glyphraisevalue = \dimexpr(15750 sp * \gre@factor * \gre@count@temp@three)\relax %`
[gregoriotex-spaces.tex:999](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L999):`		\global\gre@dimen@additionaltopspace=\dimexpr(15750 sp * \gre@count@temp@one * \gre@factor)\relax %`
[gregoriotex-spaces.tex:1006](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L1006):`		\global\gre@dimen@additionaltopspacealt=\dimexpr(15750 sp * \gre@count@temp@one * \gre@factor)\relax %`
[gregoriotex-spaces.tex:1013](https://github.com/gregorio-project/gregorio/blob/7f0023a0d025f5fc44314b205b25d513c6ce973d/tex/gregoriotex-spaces.tex#L1013):`		\global\gre@dimen@additionaltopspacenabc=\dimexpr(15750 sp * \gre@count@temp@one * \gre@factor)\relax %`

The ones in gregoriotex-spaces.tex all appear to be related to the staff lines (the distance between them, their thickness, the vertical distance between note positions, etc.).  The ones in gregoriotex-signs.tex seem to be related to positioning the mora after a punctum inclinatum.

Should I convert all these to configurable dimensions, or should they remain as hard-coded internal values?  In either case, I'm going to try to work out the sensical relations between the numbers and store them in registers so that they can be referred to by name.

rpspringuel:
There were several "magic distances" still in the code.  All of these distances are related to calibrating things to the fonts and so should not be user editable.  However, having those numbers scattered through the codebase is harder to maintain (if you change one, you have to figure out which of the others also needs to change, by how much, and where they all are), so based on relationships between the numbers, I've reduced them to 6 named dimensions which are then used as needed in the various calculations.

Fixes #1521.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - GregorioTeX will now look in additional places for scores using.  Users can specify paths to look in using `\gresetgregpath`.  If scores are not found in the current working directory or the list of provided paths, we fall back on kpse to try and find them.  See [#1395](https://github.com/gregorio-project/gregorio/issues/1395).
 - Added the ability to toggle the visibility of the translation and above lines text.  See [user request on mailing list](https://groups.google.com/g/gregorio-users/c/J-V0dkBMevY).
 
-### Deprecated
-- `noteadditionalspacelinestext` is removed in favor of calculating the actual amount of additional space needed by low notes.
+### Changed
+- `noteadditionalspacelinestext` now calculates the actual amount of additional space needed by low notes.  To enable the old behavior (when `noteadditionalspacelinestext` is set by the user), use `\gresetnoteadditionalspacelinestext{manual}`.  See [#1521](https://github.com/gregorio-project/gregorio/issues/1521) and [#1526](https://github.com/gregorio-project/gregorio/pull/1526).
 
 
 ## [5.2.1] - 2019-04-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - GregorioTeX will now look in additional places for scores using.  Users can specify paths to look in using `\gresetgregpath`.  If scores are not found in the current working directory or the list of provided paths, we fall back on kpse to try and find them.  See [#1395](https://github.com/gregorio-project/gregorio/issues/1395).
 - Added the ability to toggle the visibility of the translation and above lines text.  See [user request on mailing list](https://groups.google.com/g/gregorio-users/c/J-V0dkBMevY).
 
-### Removed
-- `noteadditionalspacelinestext` is removed in favor of calculating the actual amount of additional space needed by low notes.  Because the replacement is an automatic calculation, we skip straight to marking `noteadditionalspacelinestext` as obsolete (as we generally require deprecated code to still be functional).
+### Deprecated
+- `noteadditionalspacelinestext` is removed in favor of calculating the actual amount of additional space needed by low notes.
 
 
 ## [5.2.1] - 2019-04-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - GregorioTeX will now look in additional places for scores using.  Users can specify paths to look in using `\gresetgregpath`.  If scores are not found in the current working directory or the list of provided paths, we fall back on kpse to try and find them.  See [#1395](https://github.com/gregorio-project/gregorio/issues/1395).
 - Added the ability to toggle the visibility of the translation and above lines text.  See [user request on mailing list](https://groups.google.com/g/gregorio-users/c/J-V0dkBMevY).
 
+### Removed
+- `noteadditionalspacelinestext` is removed in favor of calculating the actual amount of additional space needed by low notes.  Because the replacement is an automatic calculation, we skip straight to marking `noteadditionalspacelinestext` as obsolete (as we generally require deprecated code to still be functional).
+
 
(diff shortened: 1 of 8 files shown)


rpspringuel:
One question on this remains:

Right now we have `noteadditionalspacelinestext` which is the amount of space added per level of really low note and is approximately `((\gre@dimen@interstafflinedistancebase + \gre@dimen@stafflinethicknessbase) / 2 )` (the vertical distance between notes a half-step apart).  Should we replace this distance with the formula or is there a reason why when adding space between the notes and the lyrics, one would want to add something other than the distance actually required by the low note position?

rpspringuel:
It should be noted that currently in the code `((\gre@dimen@interstafflinedistancebase + \gre@dimen@stafflinethicknessbase) / 2 )` is used for high notes above the staff.  For consistencies sake, if `noteadditionalspacelinestext` it to be retained, then we should do something similar for the above staff notes.

henryso (approved):
I'm going to trust that your math is right.  If the tests pass, then it probably is, or at least is close enough.

rpspringuel:
What about the issue with `noteadditionalspacelinestext`?

rpspringuel:
What about the issue with `noteadditionalspacelinestext`?  Is it more reasonable to have the amount of space added (both above and below the staff) be based how much outside the staff the high/low notes are, or allow the user to customize the amount of space added per step outside the staff?

henryso:
If it works, then go for it.  I know there's some weirdness in spaces above and below the staff.

rpspringuel:
I've pushed the changes to eliminate `noteadditionalspacelinestext` in favor of the automatic calculation.  Tests with extremely low notes are altered slightly, as the default value for that distance was not exactly what the calculation would have predicted.  If everything looks good, then this is ready for merge.

henryso:
Is there a way to set things up so that the spacing of those low notes are not changed?  I suspect that some people (like me, for instance) will have scores already typeset and want to retain the old spacing, even if it's wonky.

rpspringuel:
I could probably toggle the behavior with a conditional. Let me look at that after Mass. 

rpspringuel:
Okay, I think I've got that worked out.  Please look it over and let me know if you see any problems.

rpspringuel:
Note that the associated test PR has changed due to some of the monkey business I did to extract the old test results for the backwards compatibility tests.  That process was much harder than it really should have been.

henryso:
I don't think this is deprecation.  I see this as an alternative that doesn't really go away.

rpspringuel:
Hmmm...  That’s going to require a slightly different approach. 

Now, currently the additional space added for high notes is calculated while the additional space for low notes is manually set. If we’re going to preserve the ability to manually set the space for low notes, should we add the ability to do so for high notes?  If so, should the user be able to independently decide whether to use the manual or automatic setting for low and high notes (requiring two switches with two positions each), or does it suffice to have there be a single manual vs automatic setting?

henryso:
I think that there should be a "legacy" switch only for the lower notes as the upper notes have been working as they've been working thus far and no one has been asking for a way to make the high note spacing work like the low note spacing.  We'd only need this for the people who have things typeset and laid out and for whom such a change would cause a big mess and a need to re-layout the work.

rpspringuel:
Okay.  I can work on that tomorrow. 

rpspringuel:
Converted.  I'll be offline on Ash Wednesday so if you spot anything I'll deal with it on Thursday.

henryso:
Looks OK to me.

PR #1529, opened 2021-02-25
May it be better this tittle?

DocOc98:


rpspringuel:
No.  This is not better.  the `doc_` prefix (and others like it) are necessary to keep the build and packaging scripts happy so that only the necessary files end up in the distributions.

PR #1552, opened 2021-07-29
Fixed loss of open styles on syllables starting with a forced center.

### LINKED ISSUE #1551: gabc <c> tag not working with { } for setting syllable in chant text.
Hi,

Thanks again for a magnificent piece of software!

This is not the most earth-shattering problem, but the following example struck me as unexpected behavior.  The last syllable in the second example is not colored red.  

% !TEX TS-program = lualatex
\documentclass{article}

\begin{document}

\gabcsnippet{(c4) <c> Nar(c)ra(egf)bo(f.) </c> (::)  }

\gabcsnippet{(c4) <c> Nar(c)ra(egf){bo}(f.) </c> (::) }

\end{document}

henryso:
Fixes #1551.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,12 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][develop]
-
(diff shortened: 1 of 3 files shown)


henryso:
@rpspringuel This one is deep in the C (yacc) code, but if you feel like reviewing it, I'd appreciate it.  If not, I'll merge in a few days.

rpspringuel:
When I run the tests I only get one failure: gabc-output/bugs/fix-601.gabc.

The placement of the accent over the `æ` is slightly displaced.  That's an `<sp>` tag that opens and closes in the same syllable, however, so I don't think its related.  Probably just a minor font thing.

henryso:
I would agree.  I don't get that (or maybe it's lower than the tolerances I have set) and that's a compose character.

Issue #1558, opened 2022-01-20
Fusion Enhancements


rpspringuel:
Graduale Novum, in the offertory Recordare on p. 340 of the first volume, namely the first four notes, a torculus resupinus terminating with a virga:

![Recordare neume example](https://user-images.githubusercontent.com/7671682/150415718-6dc18562-8307-495b-a9f3-4452fd662e61.jpg)

Sandhofe's 2002 Nocturnale Romanum, page [126] in the Common for Virgins and non Virgins, Responsory 6 for Virgins, Offerentur Regi, sort of a porrectus with a fifth note stacked on top of the fourth (which should be a quilisma), at the end of the syllable Re in Regi: 

![chant_snippet](https://user-images.githubusercontent.com/7671682/150415940-e1be302b-b429-4475-9a84-728738770753.jpeg)

As best I can tell, the current rules for fusion prohibit both of these combinations (the latter is explicitly spelled out in the documentation, but the there's no mention of a virga as a primitive element in the fusion rules).  @henryso Can we modify the fusion rules to make these possible?

Reference mailing list thread: https://groups.google.com/g/gregorio-users/c/2NpiJsjEQnU/m/Ho_JZCxIAgAJ?utm_medium=email&utm_source=footer&pli=1

henryso:
It's possible. It would require virga endings as well as a "quilisma pes" endings to be drawn and added to the glyph set along with the changes in the fusion code.

rpspringuel:
Don't those glyphs already exist outside of fusion?

henryso:
Not the ones with little nubs to connect to the lines of the previous glyph.

henryso:
The quilisma-pes will end up being 5 glyphs given the ambitus.

rpspringuel:
Where are we in terms of available glyph space?

rpspringuel:
Also, has anyone looked at using glyph variants to define the connecting nubs rather than creating two separate glyphs (one connecting, one not)?

henryso:
We're close, but maybe it can fit.  It depends on how many we need for the quilisma-pes.  I haven't looked into glyph variants.  There are nine for every shape within fusion and varying numbers for starting glyphs and ending glyphs.  Quilisma-pes is actually a new can of worms because it'll be the first two-note ending glyph.

henryso:
[...]

PR #1560, opened 2022-02-05
Implemented virga fusion and shape change on clivis.

henryso:
Implementation for #1558

### DIFF
--- a/src/dump/dump.c
+++ b/src/dump/dump.c
@@ -5,19 +5,19 @@
  * Copyright (C) 2007-2021 The Gregorio Project (see CONTRIBUTORS.md)
  *
  * This file is part of Gregorio.
- * 
+ *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the Free
  * Software Foundation, either version 3 of the License, or (at your option)
  * any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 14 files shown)


henryso:
Don't know how I closed this.  Maybe I pressed some hot key in the window.

src/dump/dump.c
```diff
@@ -424,9 +424,13 @@ void dump_write_score(FILE *f, gregorio_score *score)
                                         gregorio_texverb(note->texverb));
                             }
                             if (note->choral_sign) {
-                                fprintf(f, "         Choral Sign            \"%s\"\n",
+                                fprintf(f, "         choral sign            \"%s\"\n",
```
rpspringuel, line 427: Why is this changing?  It doesn't seem related to what's being implemented here.

rpspringuel (commented):
The tests of the new shapes look good.  My only issue is what seems to be a superfluous change to the capitalization of the dump output for the choral sign.  Am I missing something as to why that needs to change?

src/dump/dump.c
henryso: This wasn't there before.  I added it (and then de-capitalized it) because I expected a change to affect that.

rpspringuel: I read the changes as showing the "Choral Sign" was there before and the only change is the change in capitalization.  What am I missing?

henryso: I must be conflating this with something.  So yes, I only changed the capitalization to be in line with the other text.

rpspringuel: Ah, I hadn't noticed the capitalization inconsistency.  I have no further problems with this change.

Issue #1579, opened 2023-07-23
Update scripts to Python 3.×


michalrus:
Hey, I noticed the scripts under `fonts/` are still expecting Python 2.7 in FontForge, while Python 2 has reached end of life on Jan 1, 2020.

Recent FontForge versions also do not support it, I tried forcing 2.7, and this happened:

```
fontforge>   Could NOT find Python3 (missing: Python3_EXECUTABLE Python3_INCLUDE_DIRS
fontforge>   Python3_LIBRARIES Development Interpreter Development.Module
fontforge>   Development.Embed) (Required is at least version "3.6")
```

There's a tool helping with automated translation – https://docs.python.org/3/library/2to3.html – perhaps it could be used here? :pray:

Thank you for considering this!

michalrus:
So far, as a workaround, this works:

```
find -iname '*.py' -not -name 'checkSyllabation.py' | xargs ${python3}/bin/2to3 -w

sed -r 's/(newfont\[glyphnumber\]\.width) = (width)/\1 = int(\2)/g' -i fonts/squarize.py
```

eschwab:
Of course you can fork and make a pull request.

I may be able to update the scrips to python3 in mid November, but if @MaestroGlanz gets to it first, all the better.

PR #1582, opened 2023-11-17
Update python scrips to Python 3.x

eschwab:
Addresses #1579

### DIFF
--- a/VersionManager.py
+++ b/VersionManager.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python
 
 """
     A script that manages the VERSION of gregorio.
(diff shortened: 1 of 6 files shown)


rpspringuel:
I think this should be against ctan.  This update doesn't change anything that would prevent a new CTAN upload and changes on the ctan branch are automatically merged into the develop branch.

VersionManager.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python
```
rpspringuel, line 1: I think it would be better if this specified `python3` rather than just `python`

eschwab:
New PR made against `ctan`

Issue #1588, opened 2024-02-27
[6.0.0] Gregorio renders multiple frames in Scribus with the same gabc score


domoknight:
When using multiple gregorio frames in the same Scribus document, the same gabc score gets processed for several of the frames (instead of the intended gabc scores).

In the image below, only the highlighted frame is supposed to show the "Hagios Ischyrós" (pardon the spelling error in the doc)
![image](https://github.com/gregorio-project/gregorio/assets/8115490/cbdb58a8-1242-41d6-9167-44a52e784aee)

Sure enough, that's the gabc we're looking for:
![image](https://github.com/gregorio-project/gregorio/assets/8115490/bec1143d-a80c-4fb4-8894-8dc8267ed55c)

But the other frames should not be showing that chant. Here is the bottom left frame:
![image](https://github.com/gregorio-project/gregorio/assets/8115490/aa3a4f91-5d49-4417-910d-5f0f81ee5f4f)

The PDF files are getting different names, but it seems like when multiple instances of LuaHBTeX are running at once, the same gabc score gets used multiple times. I only see one file in the Temp directory with the name "scribus_file-score.gabc", and it once again is the Hagios Ischyrós:
![image](https://github.com/gregorio-project/gregorio/assets/8115490/531da689-321c-4721-a37b-efb1e5161f49)

Is there a way to make the generated gabc file have a dynamic name?

PR #1591, opened 2024-03-08
Vary filename for scores in scribus render frames

rpspringuel:
Addresses #1588 

@domoknight please test to see if this works.  None of the developers regularly use Scribus so it's difficult for us to test this sort of thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][CTAN]
-
+### Fixed
+- Multiple Scribus render frames were all using the same file name, which would result in the same score appearing in all render frames.  This change makes the score files use an available Scribus variable to force multiple file names.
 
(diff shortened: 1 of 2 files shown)


domoknight:
I get the following message multiple times, and no gabc file is generated:

> (c:/texlive/2022/texmf-dist/tex/latex/libertine/LKey.tex))
> ! Undefined control sequence.
> <argument> filec@ntents@checkdir C:Users 
>                               Admin AppData Local Temp scribus _temp_render_hg
> l.11 ...\Temp\scribus_temp_render_hg3448-score.gabc}

rpspringuel:
So, the `$scribus_file$`s are full path filenames, which on a windows system includes "\" which messes with TeX's ability to process the filename (as something like `\Users` gets interpreted as a macro and TeX tries to expand it).  Based on what I can find, I don't think Scribus provides a shorter filename without the path, so let's try a different route: `\jobname`.  I've pushed a change that tries to name the score file using `\jobname` instead of `$scribus_file$`.  Please try this and let me know if it works.

domoknight:
Thank you so much! That appears to have fixed it!

Here are the files I see generated in the Temp dir:
image


And here is the resulting page in Scribus:
image

rpspringuel:
Excellent. Merging. 

PR #1592, opened 2024-04-01
Update python scrips to Python 3.x

eschwab:
Addresses #1579


rpspringuel:
To make this work I need to modify line 815 in squarize.py to make an explicit casting of `width` to `int`.  With that change (which already exists on the develop branch), everything works for me.  I'm merging and pulling this into develop.

Issue #1585, opened 2024-01-27
Segmentation Fault when the First Word is One Letter and Bolded


jglowe:
I did a brief search and didn't see this issue reported anywhere, so here it goes.

I came across a segmentation fault when the score is bold and the first syllable is one letter long. 

Below is an example of the error and the platform:

```
$ neofetch
                    'c.          REDACTED@REDACTED
                 ,xNMM.          ---------------------------------------------
               .OMMMMo           OS: macOS 14.2.1 23C71 arm64
               OMMM0,            Host: MacBookAir10,1
     .;loddo:' loolloddol;.      Kernel: 23.2.0
   cKMMMMMMMMMMNWMMMMMMMMMM0:    Uptime: 37 days, 7 hours, 20 mins
 .KMMMMMMMMMMMMMMMMMMMMMMMWd.    Packages: 202 (brew)
 XMMMMMMMMMMMMMMMMMMMMMMMX.      Shell: bash 5.2.21
;MMMMMMMMMMMMMMMMMMMMMMMM:       Resolution: 1440x900
:MMMMMMMMMMMMMMMMMMMMMMMM:       DE: Aqua
.MMMMMMMMMMMMMMMMMMMMMMMMX.      WM: Quartz Compositor
 kMMMMMMMMMMMMMMMMMMMMMMMMWd.    WM Theme: Blue (Dark)
 .XMMMMMMMMMMMMMMMMMMMMMMMMMMk   Terminal: tmux
  .XMMMMMMMMMMMMMMMMMMMMMMMMK.   CPU: Apple M1
    kMMMMMMMMMMMMMMMMMMMMMMd     GPU: Apple M1
     ;KMMMMMMMWXXWMMMMMMMk.      Memory: 2517MiB / 16384MiB
       .cooc,.    .,coo:.




$
$ gregorio --version
Gregorio 6.0.0 (kpathsea version 6.3.5).
Copyright (C) 2006-2021 Gregorio Project authors (see CONTRIBUTORS.md)
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$ cat working.gabc
%%
(c4)
O(h) please(gh) help(h) me(g) main(h)tain(i)ers(h.)
$ gregorio working.gabc
$ cat not_working.gabc
%%
<b>(c4)
[...]

PR #1594, opened 2024-04-02
Fix segfault for first syllables of one letter with a style.

eschwab:
See issue #1585

I am no expert in what the c is really trying to do in that if statement, but this fixes the reported issue and does not seem to have broken anything else.


henryso (approved):
Looks good to me. Thanks for tracking it down.

eschwab:
For the record: The cause of the segfault was that at line 1215 of characters.c `current_character->next_character` was null.

PR #1600, opened 2024-10-13
Create Vim plugin structure for gabc.vim and add ftdetect script

aquohn:
Create a proper Vim plugin structure for `gabc.vim` and add a simple script to detect `gabc` filetypes. This allows the Vim syntax file to be managed by standard Vim plugin managers.


rpspringuel:
I don't use vim and so can't check this.  Are there any other devs that can double check this work before I merge?  If not I'm willing to accept it as-is because the files changed are all in contrib.

eschwab:
I also am not a vim wizard. Should be good to go. However I cannot remember or find the discussion if this should be against `develop` or `ctan`, certainly `master` is not correct.

rpspringuel:
Normally, you’d be right. This should be against ctan, as it makes no changes to the binary. However, since I’m planning an imminent release, I’m not worried about fixing this as all three primary branches will shortly converge.

Issue #1373, opened 2017-08-09
automatic custos with flat rendering as clef on second pass


asbloomf:
[test.zip](https://github.com/gregorio-project/gregorio/files/1212932/test.zip)

If I comment out \gre@updatelinesclef in @updateleftbox, the problem goes away (but other problems are introduced).

rpspringuel:
First reported here: https://groups.google.com/forum/#!topic/gregorio-users/6tVHry453EY

asbloomf:
Any work on this?  I see it still happens in 5.1.1

adunning:
Here is another simple example, in case it is helpful:

<img width="492" alt="Screenshot 2020-05-11 at 17 09 35" src="https://user-images.githubusercontent.com/1084407/81584094-3324e200-93aa-11ea-854c-1ab979158171.png">

```tex
\documentclass[paper=A5,fontsize=12pt,DIV=12]{scrbook}

\usepackage{gregoriotex}

\begin{document}

\gabcsnippet{(c4) Re(f)gí(g)na(f) cǽ(g)li(h.)  *() læ(ixi)tá(h)re,(g_) al(ixi)le(h)lú(g)ia(f.) :(;) Qui(f)a(j_) quem(j) me(k)ru(j)í(ixi)sti(h) por(f)tá(g)re,(h_) al(ixi)le(h)lú(g)ia(f.) :(:) Re(j)sur(j)réx(k)it,(j_) si(j)cut(f) dix(g)it,(f_) al(g)le(h)lú(ixi)ia(j.) :(;) O(j)ra(f) pro(g) no(ixi)bis(h) De(g)um,(f.) al(e)le(g)lú(gg)ia.(f.) (::)}

\gabcsnippet{(cb4) Re(f)gí(g)na(f) cǽ(g)li(h.)  *() læ(i)tá(h)re,(g_) al(i)le(h)lú(g)ia(f.) :(;) Qui(f)a(j_) quem(j) me(k)ru(j)í(i)sti(h) por(f)tá(g)re,(h_) al(i)le(h)lú(g)ia(f.) :(:) Re(j)sur(j)réx(k)it,(j_) si(j)cut(f) dix(g)it,(f_) al(g)le(h)lú(i)ia(j.) :(;) O(j)ra(f) pro(g) no(i)bis(h) De(g)um,(f.) al(e)le(g)lú(gg)ia.(f.) (::)}

\end{document}
```

olivierberten:
[...]

Issue #1473, opened 2019-08-09
Custos renders off the staff


FGasper:
```
name: 1st Sunday of Advent ;
user-notes: OFFERTORY Ad te Domine;
commentary: Psalm 25 ;
annotation: ii ;
centering-scheme: english;
%fontsize: 12;
%spacing: vichi;
%font: OFLSortsMillGoudy;
%width: 5;
%height: 8.5;
%%
(f3)To(c) you,(ef/efef/HHhg) O(f) Lord,(F/GhvGFggf) (,) I(f) lift(fe/f//gwh) up(gfHHhvFEfHHh) my(f) soul;(fgf) (:)
```
^^ The above renders with gregorio 5.2.1 (LuaLaTeX Version 1.10.0  from TeX Live 2019/MacPorts 2019.50896_0) thusly, with the custos hanging off the end of the staff:
<img width="544" alt="image" src="https://user-images.githubusercontent.com/844867/62785285-58d53180-ba8e-11e9-9593-db102e27bc1a.png">

Nokel81:
Having done some investigation myself, it seems to be that Gregorio is unwilling to spit up long numes. The solution that I have come up with is adding `z` (or break) characters in the middle of the numes.

Nokel81:
A possible solution would be to have a character which means "you can break here" or have a set of rules of where gregorio will break a long nume.

FGasper:
Is it possible to put the long neume itself at the start of a new line?

-F

Nokel81:
Yes it is possible to manually do it by placing a `z` at the beginning of the nume.

`I(f) lift(fe/fgwh) up(zgfHHhvFEfHHh)`

FGasper:
That would seem to defeat one of the benefits of Gregorio: that it (like TeX) separates content from presentation.

I’m hardly capable of making the change myself, and Gregorio is already such a fantastic tool that I hesitate to say, but would Gregorio be able to determine for itself when to break before a long neume?

-F

rpspringuel:
Gregoriotex already does do some breaking.  So it ought to be just certain sequences that are causing the problems.  If you can identify the problematic sequence and where the appropriate breakpoint is in the sequence we can probably look into modifying the breaking logic.  

The idea of an optional break may also be possible, but I would need to look further at that to determine how it might be reasonably accomplished. 

[...]

PR #1605, opened 2025-01-20
Stop custos from erroneously changing into clef

### LINKED ISSUE #1604: Custos turns into clef
Here's a situation where a custos mysteriously changes into a clef, but only after running LuaLaTeX twice:

```
\documentclass{article}
\usepackage[autocompile,allowdeprecated=false]{gregoriotex}
\begin{document}
\parbox{2.2in}{\gabcsnippet{(c4)A(f)a(f)a(f)a(f)a(f)a(f)a(f)a(f)a(f)a(ixi)a(k)}}
\end{document}
```

Result of first run (after deleting `bug.gaux`):

Image

Result of second run:

Image

Things that seem to matter:

- The first note on the second line is flatted
- The second line has to have that re or higher; making it lower makes the problem go away?

Thank you!

davidweichiang:
`\GreNextCustos` was setting `\gre@box@temp@width` to a custos, then sometimes calling `\gre@custosalteration` which changed `\gre@box@temp@width` to a clef, then using `\gre@box@temp@width`, causing a clef to appear in place of the custos. This changes the order to avoid the problem; not sure if another fix is preferable.

Closes #1604.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Fixed the interaction between hyphens and styles.  See [#1538](https://github.com/gregorio-project/gregorio/issues/1538).
 - Fixed the loss of ongoing styles when a syllable starts with a forced center.  See [#1551](https://github.com/gregorio-project/gregorio/issues/1551).
 - Fixed first syllables of one letter with a style causing a segfault.  See [#1585](https://github.com/gregorio-project/gregorio/issues/1585).
+- Fixed a bug that caused a custos to sometimes change into a clef. See [#1373](https://github.com/gregorio-project/gregorio/issues/1373).
 
 ### Added
 - Added a configurable setting `\gresetunisonbreakbehavior` to control automatic line breaks between unison notes above a syllable.  Defaults to `breakable` for backwards compatibility, but may be set to `unbreakable` if that behavior is desired.  See [#1504](https://github.com/gregorio-project/gregorio/issues/1504).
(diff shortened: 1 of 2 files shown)


davidweichiang:
Closes #1373

rpspringuel:
Let me look at the failing tests before approving this.  My guess is that they're due to the fact that we haven't released in so long.  Generally speaking each new TeX Live release breaks a bunch of the visual tests in what turn out to be inconsequential ways.

davidweichiang:
I think the tests were failing on my end because of different versions of `sed` regular expression syntax.

davidweichiang:
With https://github.com/gregorio-project/gregorio-test/pull/356, all tests pass except for two (fix-601 and bar-substitution).

rpspringuel:
fix-601 fails for me, but it's an inconsequential visual change in the combining diacritic in the lyrics, so I'm not concerned by it (and have already pushed the changed expectation).

bar-substitution passes for me.

Looking at the test associated with this PR, however, I notice that the custos is being pushed out off the staff.  I want to know if that's the result of just how narrow the parbox is to make the error appear with such a small score or if it's an indicator of a measurement problem (the width of the alteration and the space between it and the custos not being taking into account).  If it's a measurement problem, then the `\global` was serving a purpose, even if it was doing it in a way that I don't like (and created the bug).

davidweichiang:
Isn't this #1473? I would think that since the `\global` had to do with measurement of a clef, it shouldn't be involved with placing the custos correctly.

If the parbox is widened to 1.1in, the overflow goes away -- is that preferable?

davidweichiang:
(And if the parbox is narrowed to 0.9in, there is overflow even if there's no flat.)

rpspringuel:
If widening the parbox makes the overflow go away, then it's almost certainly an example of #1473.  My preference is that the test only expose one bug so that when that bug is fixed the accepted test expectation looks right in all respects.

davidweichiang:
Ok, done!

rpspringuel:
This is all good for me, though we do need a CHANGELOG entry before merging.  I don't see any other documentation issues raised by this fix.

davidweichiang:
Ok, done.

Issue #1541, opened 2021-05-20
Kpsewhich error not appearing in log


rpspringuel:
The following kpsewhich based error is sent to stderr but does not get added to the glog file that Gregorio itself generates:

    154: in function `gregorio_out_name_ok': error:kpse prohibits write to file <gtex location>

This can make debugging difficult when auto compile (as this error doesn't end up in any log file).  Is it possible for the executable to copy stderr messages to glog or should we add a redirect for stderr to the command call in the Lua code which executes the autocompile?

henryso:
glog is currently being written by the lua code.  The executable should *not* be writing to the same file as it will cause buffering to be messed up and the file to be corrupted..  If anything, the lua code should append stderr from the executable to the glog.

rpspringuel:
I can make that change, but I’ll point out that we’re already passing the glow file to the executable using the -l argument. If we don’t want the executable to write to glog directly, then we’ll need to change that too. 

✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️✝️
Fr. Samuel, OSB
(R. Padraic Springuel)
St. Anselm’s Abbey 
4501 South Dakota Ave, NE
Washington, DC, 20017
202-269-2300
(c) 202-853-7036

PAX ☧ ΧΡΙΣΤΟΣ

henryso:
Yeah, either that should change or the lua code needs to close the file before calling the executable.

rpspringuel:
Is the glog file still needed after the executable is called (I haven't had a chance to check yet myself)?  If it is, is it reasonable for the Lua code to close the file, call the executable, and then immediately reopen the file?

henryso:
Yes, that's reasonable.  The executable currently *overwrites* the `-l` file, so this would need to be changed for the behavior you want.

henryso:
The executable should also be audited to ensure that it writes to said file instead of `stderr` once option parsing is completed.

rpspringuel:
My computer has decided to go on the fritz again, so I can’t complete my analysis, but I think that support.c may have the offending line. It’s the only place where gregorio_set_error_file is called with stderr as hard coded argument. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

henryso:
[...]

PR #1606, opened 2025-01-22
Stderr fix

rpspringuel:
I'm not sure why we never considered this for a merge.  It looks fully ready to me so I'm putting it here for comments.


rpspringuel:
Barring objections I'll merge this on Monday.

eschwab:
I was hoping to test it, but cannot get time for that until Tuesday. The code looks fine though.

rpspringuel:
I can wait until Tuesday.  Please do review it if you have the time.

eschwab:
Had a chance to look at this today. Works as expected.

tex/gregoriotex.lua
```diff
@@ -798,8 +870,8 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
     extra_args = extra_args..' -D'
   end
 
-  local cmd = string.format('%s %s -W -o %s -l %s "%s"', gregorio_exe(),
-      extra_args, gtex_file, glog_file, gabc_file)
+  local cmd = string.format('%s %s -W -o %s -l %s "%s" 2> %s', gregorio_exe(),
```
davidweichiang, line 873: I think this line might be causing an error for me sometimes. The error looks like this:
```
(.../tex/latex/latexconfig
/epstopdf-sys.cfg))ignored arguments: 2> tmp-gre/alma-6_0_0.glog
Usage: gregorio [OPTION]... [-s | INPUT_FILE]
Try 'gregorio --help' for more information.
Proceeding anyway...

(.../tmp-gre/alma-6_0_0.gtex
```
Sometimes it actually causes the run to pause and sometimes it finishes.

rpspringuel: That's supposed to be file redirection to make sure that output to stderr gets recorded in the glog file (makes debugging easier).  Not sure why it would be rendering it as "ignored arguments".  Do we need to quote the filename of the glog file?

davidweichiang: I can't reliably reproduce the problem (even though I've seen it many times), so I'm also stumped.

davidweichiang: Found the problem: if `--shell-escape` is not given, but `gregorio` is listed in `shell_escape_commands` in `texmf.cnf`, then `2>` gets interpreted as an ordinary argument. Gregorio will output a warning about extra arguments, which is harmless, but the glog file will be empty.

rpspringuel: `gregorio` will *always* be in the listed `shell_escape_commands` for TeX Live (and I think MiKTeX too).  I'll ask the TeX Live people about this.

davidweichiang: Right, which is a good thing, but presumably the restriction on redirects is a security measure. (For example, they certainly wouldn't want to allow `os.execute("gregorio | rm -rf *")`.)

Perhaps an alternative is to make gregorio itself write the .glog file.

rpspringuel: It already can.  That's what the `-l` option is all about.  There are just some errors that the `-l` option doesn't capture.  See #1541.

The point of the redirection is to try and capture those errors too, as they sometimes are needed to diagnose what's going wrong.

davidweichiang: I see, so ultimately this error message I'm seeing is unintended but mostly harmless. I've verified that if gregorio outputs an error, it does appear in the glog.

I mentioned above that sometimes TeX was stopping, and I am now guessing that it was for some other reason (still don't know what).

rpspringuel: So long as it's harmless, then I see no need to change things.  If it works as expected with the `--shell-escape` option, then the error messages can be captured when needed even if normal operation doesn't capture them because of the security measures.

Issue #1575, opened 2023-02-14
Soft flat?


MRoth1910:
I'm working on a Kyriale. In the gabc score of Agnus IX found at [Gregobase](https://gregobase.selapa.net/chant.php?id=707), the flat of "miserere" and "dona" is only on the first note, as per the source (`mi(gxfgh)se(hvGF)`) However, in my case, it appears that I'm stuck, for now, with a line break which would require inserting an additional flat. **But** I'd rather not insert a flat into the gabc, as that would make the file less portable (although I will if it means that the flat is printed correctly).

I mean "soft" just like in "soft hyphen", only printed if the line needs to be broken at that point; a soft flat would be printed in a case like mine, where the first note is already flat but where the second note to be sung with a flat could unfortunately wind up on another line.

This has some similarities to the problem mentioned in #157, but I think the problem is **only** at the ends of lines in a word with several flattened notes, due to the less predictable nature of line breaks, which cannot be as easily adjusted, unlike when setting the original scores by hand. I'm far from an expert, but awkward breaks like in Agnus IX are fairly rare…

Since a word division, a natural, or a bar cancels the flat, the first flat of a word or the first one after a bar, can be coded as normal without trouble. Nothing needs to be coded one way or the other to return the note to natural after a bar either even if theory one could add the natural  just to be safe (cf. the [Gradual of Ss Peter and Paul](https://gregobase.selapa.net/chant.php?id=307)) If I understood the other discussion correctly, a line break is known already by Gregorio, and these two things together mean that only the line breaks need to be considered for printing some of the flats, not the other rules.

xmarteo:
Is there a rule that a flat gets canceled after a line break? If not, why print it a second time?

xmarteo:
OK.
This is clearly harder to do than "like soft hyphens in lyrics, but with flats", because this one relies on LaTeX, who simply /always/ inserts a hyphen when a word is spread over two lines, whereas the soft flat requires inspection of the presence of a Si before the next bar or word when a word is spread over two lines.

MRoth1910:
> Signum ♭ molle ante notam si positum minuit acuitatem notæ uno semitono. Effectus autem ejusmodi signi se extendit usque ad finem lineæ, nisi prius ♮ quadro evictus fuerit.

[...]

PR #1607, opened 2025-01-24
Add option to print only the first flat on each line

### LINKED ISSUE #157: Improvement of flats line by line
Flats can have convoluted meanings, and occur until the end of the line, so Gregorio should know be able to add flats or naturals at beginning of line.

More details explanation, from a mail on the mailing-list.

Currently this is the part of score production that is most laborious for me, because the flats and naturals have to be readjusted manually each time depending on the width and text/chant size of the score. If we were able to have a more absolute system of flats that could appear or not appear depending on the context, that would go a long way towards making an individual GABC file much more flexible.

As far as I know, there are two different systems in 19th/20th century chant books for notating flats and naturals.

1) In the Vatican edition, a flat applies until a natural, or a division line, or a new word.

2) In Dominican chant books, a flat applies until a natural or the end of a line.

I think that the best way to notate flats in GABC should be to put a flat indication on every note that ought to be flat, and then have Gregorio determine when to actually print the flats in relation to the rules of the flats. If this change was to be implemented, perhaps it could be done in a way that left that current system in place (and would thus be backwards compatible) but which added a flat or natural to a note in a way analogous to the notations that add a rhythmic mark to a note, i.e., by putting a letter after the pitch/height indication. As an offhand suggestion, you could use % for flat: [e.g.: A(i%)men.(j)]

A specific letter-notation for natural would not be necessary because a natural would be added automatically if the flat was not present in the notation.

davidweichiang:
This uses `post_linebreak` to figure out the first flat on each line, and records this information in the `.gaux` file so that `\gre@alteration` can print only the first flat. It needs two or three runs to converge.

Closes #157. (But not #1575.)


davidweichiang:
Here's an example of the same score set at two different widths (and where GregoBase is currently wrong):

image

tex/gregoriotex-signs.tex
```diff
@@ -2695,16 +2697,59 @@
...
+    \fi
+  \else
+    \global\gre@alteration@isvisibletrue
+  \fi
+}
+
+\def\gre@alteration@visible#1#2#3#4#5#6{%
+  \gre@trace{gre@alteration@visible{#1}{#2}{#3}{#4}{#5}{#6}}%
+  \ifcase#2\relax
+    \or\def\alterationchar{\gre@fontchar@flat}\def\alterationhole{\gre@fontchar@flathole}% flat
```
rpspringuel, line 2741: We try to avoid unprefixed macros in order to prevent name conflicts with other packages (regardless of how unlikely we think they are).  Our prefix system is:
- `\gre@`: Internal macros that users should never interact with in their documents
- `\Gre`: Macros written to the gtex file by the executable. Advanced users might interact with these to do special things in their document
- `\gre`: user facing macros.  We expect users to make use of these macros regularly.

With that in mind, `\alterationchar` and `\alterationhole` should probably be named `\gre@alterationchar` and `\gre@alterationhole`.

rpspringuel:
Just one thing that I see right now (and it's not feature related), but it may just be my relative unfamiliarity with the desired feature.

Do you think that you could create a matching branch in the test repository and commit proposed tests showing the behavior you're trying to fix.  Said tests could be based on the scores where things are currently incorrect, but if the problems are not glaringly obvious at a glance, some text explaining the problem should be added to the test.

I wouldn't commit the "corrected" expectations yet so that the image comparison failures can be used to highlight the changes and make sure they are what is desired.

davidweichiang:
Done, though I'm not sure if the test is what you're looking for yet.

rpspringuel:
Thinking about this, a modification to the gabc syntax may be necessitated by this idea: we might need to differentiate between a flat (or other alteration) that should always show and one that is allowed to disappear according to the ruleset chosen in the document.

Currently we have `gx`, `gy`, and `g#` which produce their respective alteration.  I would assign these to the forced alteration set (in keeping with their current behavior, where all alterations are forced).  `gX` and `gY` could be used for optional flats and naturals (ones subject to the ruleset), but what would the equivalent for the sharp be?

davidweichiang:
I added the header option `alteration-scope:line` to indicate that all the flats in the score are "optional." But in the example, there is one flat that doesn't really need to be there, so I think it makes sense to always have the current "hard" flat available. In that case, I think the header option will no longer be needed.

I like `gX` and `gY` (though I am unsure if `gY` is really needed). I imagine that the "soft" sharp would be very rare, so maybe something like `g#*`?

tex/gregoriotex-main.tex
```diff
@@ -2142,18 +2142,4 @@
 % We load the default space configuration.
 \greloadspaceconf{default}%
 
-\newif\ifgre@alteration@line
-\gre@alteration@linefalse
-\def\gresetalterationscope#1{%
```
rpspringuel: I'd actually keep some sort of setting here, primarily for future use.  Right now the settings would be `none` (all alterations are considered hard, even if they were written in the gabc as soft) and `line` (soft flats which are not the first on a line are suppressed).

Eventually we might add other rules for when soft flats are suppressed (as per #157) and so having this setting in place makes sense (though probably not as a simple boolean).

rpspringuel:
Looking back over #157 and #1575, I noticed that naturals (or sharps, I suppose) would also cancel the flat. @davidweichiang, have you taken that into account in this change?  (I'm working on my own multi-line initial branch to try and get that finished and so don't have time to poke at this thoroughly myself).

The other points where alterations might get canceled are word-breaks and bar lines.  We'll want to implement those possibilities eventually (and the combinations there-of that specify the Vatican and Dominican rulesets), but if we can just get naturals and line breaks for now, I think that's sufficient.

tex/gregoriotex-main.tex
davidweichiang: The argument for not needing a setting is that any other convention (flats last until the end of the word or a bar) can be expressed using hard flats. It's only when flats last until the end of the line that soft flats are needed.

However, if there's a setting, should it be set in the gabc header or using a TeX command? Originally I thought it should be in the gabc header because it affected the semantics of the notes (that is, what pitches they stand for). But now that hard and soft flats have different letters, I think this setting only affects the appearance of the output and it's more appropriate as a TeX command.

davidweichiang:
The current logic is:

- Hard flats (`x`), sharps (`#`), and naturals (`y`) are always printed. So are the parenthesized (`?`) versions.
- A soft flat (`X`) is printed iff it is the first alteration (of any type) on the line or the previous alteration on the line is not `x` or `X`.
- Similarly for soft sharp (`#*`).
- A soft natural (`Y`) is printed iff the previous alteration on the line is not a natural (either `y` or `Y`).

So, for example, if the alterations on a line are `YXXYYX`, then they would be printed as
- `Y`: nothing
- `X`: flat
- `X`: nothing
- `Y`: natural
- `Y`: nothing
- `X`: flat

tex/gregoriotex-main.tex
rpspringuel: While bar lines and word breaks are always visible in gabc and thus any convention that relies solely on those for determining the visibility of an alteration can be implemented manually in the gabc much easier than one based on line breaks (which we purposely try to leave undefined in the gabc), if you went that way switching between conventions would require recoding the gabc.  Additionally, conventions which combine bar lines and/or word breaks with line breaks could not be represented this way.

Having a setting allows for a source which has been encoded with soft alterations to be used with any available convention.

davidweichiang: That's fine with me -- is it better for this to be in the gabc header or in the TeX source? I am thinking the latter, as I got the impression that a lot of things have been moved out of the header.

In any case, it seems that it should actually be three flags for bar lines, word breaks, and line breaks, for a total of eight (!) conventions. So something like
- `\grealterationscope{bar,word}` (Vatican)
- `\grealterationscope{line}` (Dominican)
- `\grealterationscope{bar,word,line}` (the variation mentioned in #1575)
Does that look right?

I do want to note that in the PR as it stands, it's possible to represent a convention combining line breaks and word breaks, as discussed in #1575. If one wrote `mi(gxfgh)se(gXhvGF)`, then the flat on `mi` would always appear, and the flat on `se` would only appear if a line break occurred between the syllables.

What this PR does not fix is if a line break occurs between the flat and the note itself, e.g., between `gXhv` and `GF`. I don't know what to do about that and am hope it just doesn't happen much.

rpspringuel: The goal for differentiating between header appropriate and TeX appropriate is as follows:
- Things which are related to appearance should be TeX settings.  
- Things related to source should be specified in headers.

Alterations are clearly a borderline case.  In scores encoded purely with soft alterations, the convention can be switched at-will and so is clearly appearance related (a TeX setting).  In scores encoded with hard alterations (generally according to the convention in the source), then we're dealing with something that is source related (and thus should be a header).

My personal inclination is to avoid the new header and push people towards re-encoding scores to use the soft alterations, but I'm open to counter arguments.


> What this PR does not fix is if a line break occurs between the flat and the note itself, e.g., between gXhv and GF. I don't know what to do about that and am hope it just doesn't happen much.

I take your point, but `gXhvGF` isn't a breakable neume.  In fact #1473 is at least partly due to GregorioTeX being very adverse to breaking neumes.  It would take a much longer melisma than that to actually create the problem and I doubt the flat would be placed so far away from the note it's supposed to effect in such a melismatic sequence (though, if you have an example, I'm open to being proved wrong).  I think this is a corner case that is so rare as to be ignorable.

davidweichiang: Done, and glad to hear that the corner case is unlikely!

davidweichiang: In that case, this would close #1575 as well.

davidweichiang:
I tried to explain the current logic in the documentation, and I hope it was clear enough. I think this is ready to go but am of course happy to make further changes if needed.

Here's a summary of the user-facing changes:
- Three new alterations, soft flat (`X`), natural (`Y`), and sharp (`#*`). Not a lot of thought was put into the choice of `#*`.
- A new command named `\gresetalterationeffect`. I don't know what the convention is about whether the name should include `set`. And maybe `scope` was a better choice than `effect`.
- Theoretically the argument is a comma-delimited list of keywords, but the only meaningful choices are
  - `\gresetalterationeffect{note}` (the default): print all soft flats and sharps, and no soft naturals
  - `\gresetalterationeffect{line}`: Dominican

rpspringuel:
Do we use `*` for anything else?  I don't recall off hand.  If not then `*` could be the soft sharp all by itself.

davidweichiang:
I don't see `\\*` in the parser code, but that seems like a nice character to spend on something that will never(?) get used...

How about `##`, which is sort of like a big `#`?

rpspringuel:
I'm not to worried about "spending" characters.  Adding new gabc syntax has become a very rare event at this point and we still have lots of unused characters.

davidweichiang:
I'll change it to `##` if it's okay with you. Has a clearer connection to `#` and is easy to type.

eschwab:
I have been loosely following this PR, when it is ready I would like to review the documentation for ease of comprehension.

As for `##` or `#*` I concur that `##` is a better symbol. 

There is also `♯` U+266F the actual sharp, but that would easily be confused with `#`.

davidweichiang:
Thanks! IMO the documentation is ready for review, but it will be very helpful to have a third pair of eyes on it.

davidweichiang:
Actually, no, I'm not happy with the documentation. I'll work on it more and let you know when it's really ready.

rpspringuel:
Don’t forget we have the doc_check.sh script to help with the documentation process. It won’t write documentation for you, but it’ll help you identify which macros still need to be documented. 

davidweichiang:
@eschwab Hopefully that's better now!

doc/Command_Index_gregorio.tex
```diff
@@ -530,6 +530,18 @@ \section{Gregorio Controls}
...
 \end{argtable}
 
+\macroname{\textbackslash GreFlatSoft}{\#1\#2\#3\#4\#5}{gregoriotex-signs.tex}
+Macro to typeset a soft flat.
+
+\begin{argtable}
+  \#1 & integer & Height number of the flat.\\
+  \#2 & \texttt{0} & The flat is not part of the clef.\\
+  & \texttt{1} & The flat is part of the clef.\\
+  \#3 & \TeX\ code & signs to typeset before the glyph (typically additional bars, as they must be "behind" the glyph)\\
```
rpspringuel, line 540: Please use typographical quotes around behind.

doc/Command_Index_gregorio.tex
```diff
@@ -810,6 +822,18 @@ \section{Gregorio Controls}
...
 \end{argtable}
 
+\macroname{\textbackslash GreNaturalSoft}{\#1\#2\#3\#4\#5}{gregoriotex-signs.tex}
+Macro to typeset a soft natural.
+
+\begin{argtable}
+  \#1 & integer & Height number of the natural.\\
+  \#2 & \texttt{0} & The natural is not part of the clef.\\
+  & \texttt{1} & The natural is part of the clef (doesn't happen).\\
+  \#3 & \TeX\ code & signs to typeset before the glyph (typically additional bars, as they must be "behind" the glyph)\\
```
rpspringuel, line 832: Same comment about quotes.

doc/Command_Index_gregorio.tex
```diff
@@ -1080,6 +1104,18 @@ \section{Gregorio Controls}
...
 \end{argtable}
 
+\macroname{\textbackslash GreSharpSoft}{\#1\#2\#3\#4\#5}{gregoriotex-signs.tex}
+Macro to typeset a soft sharp.
+
+\begin{argtable}
+  \#1 & integer & Height number of the sharp.\\
+  \#2 & \texttt{0} & The sharp is not part of the clef.\\
+  & \texttt{1} & The sharp is part of the clef (doesn't happen).\\
+  \#3 & \TeX\ code & signs to typeset before the glyph (typically additional bars, as they must be "behind" the glyph)\\
```
rpspringuel, line 1114: And a third time.

davidweichiang: This was a copy-paste from the existing documentation for `\GreFlat` and friends. I've changed them throughout the whole manual.

davidweichiang:
@rpspringuel and @eschwab I believe this is ready to go -- your remaining comments would be welcome!

davidweichiang:
Done (not sure how much detail to go into, since it's complicated to explain)

doc/Gabc.tex
```diff
@@ -490,17 +492,51 @@ \subsubsection{One-Note Neumes}
...
+appending characters to pitch letters.
+
+Flats, sharps, and naturals each have a ``soft'' version (\verb|X|, \verb|##|, and \verb|Y|, respectively). Soft flats are printed if no previous flat (on the same pitch) is in effect. Similarly for soft sharps. Soft naturals are printed if a previous flat or sharp (on the same pitch) is in effect.
+What ``in effect'' means is controlled by the \verb|\gresetalterationeffect| macro (see \S\ref{setalterationeffect}). Soft alterations currently have two use cases.
+
+For Dominican chant (in which flats take effect until the end of a line or until cancelled by a natural), use \verb|\gresetalterationeffect{line}|, and type a soft flat (\verb|X|) in front of \emph{every} note that is to be sung a half-step flat. Naturals should also be soft (\verb|Y|), but only naturals that cancel flats need to be typed. For example:
+
+\begin{center}\parbox{5in}{%
+  \gresetinitiallines{0}\gresetlyriccentering{firstletter}%
+  \grechangestyle{translation}{\ttfamily}%
```
eschwab, line 520: There is no translation in the gabcsnippet. Does this line need to be here? Same with the `\grechangedim{translationheight}` two lines below.

tex/gregoriotex-signs.tex
```diff
@@ -2695,16 +2699,113 @@
...
+    \or\def\gre@alteration@char{\gre@fontchar@flatparen}\def\gre@alteration@hole{\gre@fontchar@flatparenhole}% parenthesized flat
+    \or\def\gre@alteration@char{\gre@fontchar@naturalparen}\def\gre@alteration@hole{\gre@fontchar@naturalparenhole}% parenthesized natural
+    \or\def\gre@alteration@char{\gre@fontchar@sharpparen}\def\gre@alteration@hole{\gre@fontchar@sharpparenhole}% parenthesized sharp
+    \else\gre@error{Invalid alteration type}% soft alterations should have been changed to hard
+  \fi
+  \setbox\gre@box@temp@width=\hbox
+    attr \gre@attrid@alteration@pitch=#1\relax
+    attr \gre@attrid@alteration@type=#2\relax
+    {\gre@pointandclick{\gre@alteration@char}{#6}}%
+  \ifnum#3=0\relax %
```
eschwab, line 2808: Coding style question: `%` is probably not needed after `\relax` Do we want to be consistent with omitting the `%`? 

`:2764, :2795, :2805, :2806` vs `:2808`

Also the `\relax % Comment` construction shows the gobbled whitespace VS. `\relax% Comment` For readability I am fine with `\relax % Comment` and omitting the `%` after other `\relax`'s that have no comment.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
 - Added a configurable setting `\gresetunisonbreakbehavior` to control automatic line breaks between unison notes above a syllable.  Defaults to `breakable` for backwards compatibility, but may be set to `unbreakable` if that behavior is desired.  See [#1504](https://github.com/gregorio-project/gregorio/issues/1504).
 - Added the ability to fuse upwards to a virga.  See [#1558](https://github.com/gregorio-project/gregorio/issues/1558)
 - Added the ability to use the "stroke" form of a clivis instead of the default two-notes form by specifying `[shape:stroke]` after the clivis to change.  See [#1558](https://github.com/gregorio-project/gregorio/issues/1558)
+- Added new alterations, soft flat (X), soft natural (Y), and soft sharp (##), that only appear when necessary, as determined by a new option `\gresetalterationeffect`. These can be used for Dominican chant. See [#157](https://github.com/gregorio-project/gregorio/issues/157) and also [#1575](https://github.com/gregorio-project/gregorio/issues/1575).
```
eschwab, line 21: May be good to add a `Defaults to ...` statement as seen above on linie `18`

doc/Command_Index_User.tex
```diff
@@ -1408,6 +1408,23 @@ \subsubsection{Ancient Notation}
...
+      & \texttt{line} & effect ends at the end of the line \\
+\end{argtable}
+
+Soft flats are printed if no previous flat (on the same pitch) is in effect. Similarly for soft sharps. Soft naturals are printed if a previous flat or sharp (on the same pitch) is in effect.
+
+Thus, under \verb|\gresetalterationeffect{note}|, all sharps and flats lose their effect immediately, so all soft sharps and flats are printed, but no soft naturals are printed.
+Under \verb|\gresetalterationeffect{line}|, all sharps and flats are in effect until the end of the line (unless cancelled by a natural), which is the convention used in Dominican chant scores.
+
+(Argument \#1 is actually a comma-separated list of keywords, but currently, it must not be empty, and \texttt{note,line} has the same effect as \texttt{note}. In the future, other options like \texttt{word} and \texttt{bar} may be added.)
+
```
eschwab, line 1427: The most confusing thing for me is I cannot easily tell if I need to use this macro or not. Perhaps a quick note about the default behavior of gregoriotex could be added. (cf. note on changelog)

Also why is `note, line` equal to `note`? Does the first or last option have priority? Does this macro need multiple options?

tex/gregoriotex-signs.tex
davidweichiang: This happens a lot in existing code (>1000 times) and I was following the convention. I agree that it does not need to be there, but its removal should perhaps be another (huge) PR.

doc/Command_Index_User.tex
davidweichiang: I'll make these changes, but also have a question. I'm doubtful that `\grealterationeffect{note}` has any real use. It just seemed like the most neutral thing to do. Would it be preferable to instead make `line`, which is actually useful, the default? This would not break any existing code, because it only affects soft alterations, which are not used in existing code.

rpspringuel: I think making soft alterations default to `line` is fine.  As you say, this doesn't break existing scores because hard alterations aren't affected and were the only thing that existed previously.

Along those line, however, a note in UPGRADE about how to modify scores to make use of the new feature might be in order.

rpspringuel:
Test results all look good.

tex/gregoriotex-signs.tex
rpspringuel: I agree that this sort of style thing should await a different PR.

davidweichiang:
What's the convention about `\gresetalterationeffect` vs. `\grealterationeffect`?

davidweichiang:
I've tried to incorporate all the suggestions above. Thanks!

davidweichiang:
Done! Thanks, hopefully this will be a big time-saver for some people.

eschwab:
Let me just build and test it. If all good, I'll merge it in.

avbop:
I don't find myself using gregorio very much these days, but this is a feature I've been interested in for a long time. Thank you for your work on it!

PR #1612, opened 2025-02-05
Fix #1611

### LINKED ISSUE #1611: Errors in restricted \write18 mode
Earlier (https://github.com/gregorio-project/gregorio/pull/1606/files#r1938549811) I reported a "mostly harmless" error when running in restricted \write18 mode (i.e., without the --shell-escape option but with gregorio listed among the `shell_escape_commands` in texmf.cnf). I mentioned a more serious error, and finally figured out what causes it.

- LuaTeX is run in restricted \write18 mode.
- The .gtex file exists but can only be found by kpathsea (gregoriotex.lua:967), so it is an absolute pathname.
- The .gabc file is newer than the .gtex file.
- gregorio is run and rewrites the .gtex file.
- gregoriotex.lua:924 tries to open the .gtex file in mode 'a', but because it is an absolute pathname, the open fails.
- The TeX run stops, and whether the user types 'x' or enter, the .gtex file does not get included.
- If the user re-runs TeX, everything is fine.

I would be happy to help with the fix but don't know what the right fix is. The fixes I can think of are:

- Open the .gtex file in mode 'r' instead of 'a'. (I didn't understand why it was opened for writing anyway.)
- Don't use kpathsea to find the .gtex file. (Just keeping in ./tmp-gre seems fine to me.)
- Attempt to change the absolute pathname into a relative pathname.

davidweichiang:
After running gregorio, open the .gtex file read-only to work in restricted \write18 mode (#1611).

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -899,12 +899,15 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
         .."'%s' with %s.\nPlease check your score file.", gabc_file,
         gregorio_exe())
   else
-    local gtex = io.open(gtex_file, 'r')
-    if gtex == nil then
-      err("\n Unable to open %s", gtex_file)
-    else
-      gtex:close()
-    end
+    -- The next few lines would open the gtex file for writing so that LuaTeX records the fact that gregorio has written to it
+    -- when the -recorder option is used.
+    -- However, in restricted \write18 mode, the gtex file might not be writable. Since we're the sole consumer of the gtex file, it should be okay not to record the write.
+    --local gtex = io.open(gtex_file, 'a')
+    --if gtex == nil then
+    --  err("\n Unable to open %s", gtex_file)
+    --else
+    --  gtex:close()
+    --end
     local glog = io.open(glog_file, 'a+')
     if glog == nil then
       err("\n Unable to open %s", glog_file)
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -899,9 +899,7 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
         .."'%s' with %s.\nPlease check your score file.", gabc_file,
         gregorio_exe())
   else
-    -- open the gtex file for writing so that LuaTeX records output to it
-    -- when the -recorder option is used
-    local gtex = io.open(gtex_file, 'a')
+    local gtex = io.open(gtex_file, 'r')
     if gtex == nil then
       err("\n Unable to open %s", gtex_file)
     else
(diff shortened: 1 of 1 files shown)


tex/gregoriotex.lua
```diff
@@ -899,9 +899,7 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
         .."'%s' with %s.\nPlease check your score file.", gabc_file,
         gregorio_exe())
   else
-    -- open the gtex file for writing so that LuaTeX records output to it
-    -- when the -recorder option is used
-    local gtex = io.open(gtex_file, 'a')
+    local gtex = io.open(gtex_file, 'r')
```
eschwab, line 902: The comment should probably stay but change `writing` to `reading`. Would then match the comment on line `1080`

See 5200037adf991cc545680d45cd377e3226a9fb11

davidweichiang:
I see, the .gtex file is being opened for writing because gregorio has written to it, and now the Lua code "pretends" to write to it to let the LuaTeX recorder record it. So changing 'r' to 'a' is not the right solution.

davidweichiang:
Since the .gtex file is consumed by this code, I believe that the recording is not necessary. It should be enough to open the .gabc file for reading so that the recorder knows that this code depends on the .gabc file.

https://tex.stackexchange.com/questions/245336/how-can-one-manually-add-a-line-to-the-pdflatex-recorder-fls-file

rpspringuel:
I wasn't experiencing this problem myself, so I need to clarify how to reproduce it: I should compile a file with a plain vanilla TeX Live installation (no development version of Gregorio installed locally) without the `--shell-escape` option?

davidweichiang:
The problem will surface with a local development version of Gregorio, as long as it's not this branch. Yes, you'd have to drop the --shell-escape option but ensure that gregorio is listed in texmf.cnf.

A less aggressive solution than the current one (which is to comment out that whole block) would be to attempt to open the gtex file for writing, but just comment out the line that raises an error. Then there would be no change in behavior for users using --shell-escape, and for users not using --shell-escape, the change in behavior would be, I believe, both invisible and harmless.

tex/gregoriotex.lua
```diff
@@ -899,14 +899,15 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
...
-      err("\n Unable to open %s", gtex_file)
-    else
-      gtex:close()
-    end
+    -- The next few lines would open the gtex file for writing so that LuaTeX records the fact that gregorio has written to it
+    -- when the -recorder option is used.
+    -- However, in restricted \write18 mode, the gtex file might not be writable. Since we're the sole consumer of the gtex file, it should be okay not to record the write.
+    --local gtex = io.open(gtex_file, 'a')
+    --if gtex == nil then
+    --  err("\n Unable to open %s", gtex_file)
```
rpspringuel, line 907: This is the line you're talking about, right?  What about simply converting the error to a warning?  Something to the effect that the `-recorder` option might not work correctly if you're making manual edits to the gtex file without the `--shell-escape` option.

davidweichiang:
Right, that would work. I do believe that the recorder will work fine,
though.

eschwab:
This solution seems better to me.

davidweichiang:
Agreed -- the last commit does that and tries to explain in the warning message.

davidweichiang:
Is there still time to merge this? The error this fixes will affect many people, I think.

eschwab:
I don't see a need for a changelog entry since this is a downgrade from `err` to `warn` with better warning language and documentation in the code.

rpspringuel:
I'm all for just merging this now.  In fact, I'll do it.

PR #1618, opened 2025-02-10
Fix computation of top and bottom pitches of glyphs

### LINKED ISSUE #1613: Glyph top/bottom not computed correctly
This code
```
\documentclass{article}
\pagestyle{empty}
\usepackage{gregoriotex}
\gresetinitiallines{0}
\begin{document}
\gabcsnippet{<alt>alt</alt>kI(kI) b(b)}
\end{document}
```
produces output (on the 2nd run)

Image

because the top/bottom pitch are not computed correctly. The gaux file has

```
 ["line_heights"]={
  ["17ea64d7de758f5a97f2bf4f6305c6fb40b31111.1"]={
   [0]={1,11,4,0,1},
   ["last"]=0,
  },
```

but the 11 should be a 13.

The reason is that in `\GreGlyph`, the box containing the glyph (`\gre@box@temp@width`) is created before the calls to `\GreGlyphHeights`. When the box is actually printed, it retains the wrong top/bottom attributes. There's a one-line fix (to unbox and rebox the glyph at the right time), but maybe the root of the problem is that `\GreGlyphHeights` shouldn't be slipped into argument `#5` of `\GreGlyph`. Should the top/bottom pitches be made into their own arguments instead, or can they be computed?

davidweichiang:
This is just a one-line change.

Closes #1613.


eschwab:
This works for me.

rpspringuel:
Concur.

Issue #1319, opened 2017-02-25
`spacebeneathtext` causes syllable hyphen to drop down into translation line.


matthewgonzalez:
This bug has been plaguing me on a project I am working on that relies heavily on the translation feature.

**Gregorio Version:** 4.2.0

**The problem:** At the end of a line, the hyphen drops down into the translation text (see screenshot)
**Culprit:** I've narrowed this down to some issue with setting `spacebeneathtext`
**What should happen:** The hyphen should display with its respective line and syllable, even when adjusting `spacebeneathtext`

I've attached a minimum example. Pardon the large margins; they were set this way purposely to force the hyphen situation.

![screen shot 2017-02-25 at 4 00 33 pm](https://cloud.githubusercontent.com/assets/1775184/23335138/5569104c-fb74-11e6-8440-14ee4429b92b.png)

[translation-hyphen-bug.zip](https://github.com/gregorio-project/gregorio/files/801548/translation-hyphen-bug.zip)

rpspringuel:
I don't have time to track down which algorithm is responsible for inserting that hyphen at the moment, but based on the described symptoms, it must be failing to add `spacebeneathtext` to the height of the hyphen.

matthewgonzalez:
This seems to be an issue with the hyphens which are added via `gregoriotex.lua` (to my knowledge, there are two other types of hyphens which are added in `gregoriotex-syllable.tex`).

The code in question is the following (sidenote, `adddash` is mispelled on the last line!):
```
if adddash==true then
          local dashnode, hyphnode = getdashnnode()
          dashnode.shift = currentshift
          hyphnode.font = currentfont
          insert_after(line.head, lastseennode, dashnode)
          addash=false
end
```
https://github.com/gregorio-project/gregorio/blob/e72026e4a2770d307b119b90f8e6dfbf0d812bdd/tex/gregoriotex.lua#L548-L554

The `currentshift` isn't being updated due to the following conditional (I'm not sure why the conditional is there in the first place):

```
if currentshift == 0 then
              currentshift = n.shift
end
```
https://github.com/gregorio-project/gregorio/blob/e72026e4a2770d307b119b90f8e6dfbf0d812bdd/tex/gregoriotex.lua#L514-L516

If this conditional is removed, the hyphen assumes the correct shift adjustments and doesn't get pushed down.

I point these out to help you properly fix this. I'm not super familiar with your codebase.

rpspringuel:
[...]

PR #1619, opened 2025-02-12
Fix vertical position of hyphens at line breaks (#1319)

davidweichiang:
The code for inserting hyphens was intertwined with a couple of other things, so this factors it out into its own block.

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -476,11 +476,10 @@ end
 local function dump_nodes_helper(head, indent)
   local dots = string.rep('..', indent)
   for n in traverse(head) do
-    local ids = format("%s", has_attribute(n, part_attr))
+    local ids = format("%s", has_attribute(n, glyph_id))
     if node.type(n.id) == 'penalty' then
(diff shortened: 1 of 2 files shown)


tex/gregoriotex.lua
```diff
@@ -830,8 +804,6 @@ local function post_linebreak(h, groupcode, glyphes)
           new_score_last_syllables[syl_id] = syl_id
         end
       end
-      -- we reinitialize the shift value, because it may change according to the line
-      currentshift=0
```
davidweichiang, line 834: This was the bug: currentshift should have been reset at every syllable, like currentfont is.

rpspringuel:
This looks good to me.

davidweichiang:
I also think this is ready.

Some more obscure effects of this change:
- If a syllable is written in two different fonts, I believe the old code chose the first font, and the new code chooses the last font
- I believe the old code would suppress the line-break hyphen if an explicit hyphen occurred anywhere in the last syllable; the new code only suppresses if an explicit hyphen occurs at the end of the last syllable.

Issue #1503, opened 2021-01-21
Line height adjustment is off when linebreak is in a clef change


olivierberten:
![funnycustos](https://user-images.githubusercontent.com/213117/105418211-03b6bc80-5c3d-11eb-8032-65761160f40c.png)

https://gregobase.selapa.net/chant.php?id=13362

rpspringuel:
I think that the automatic eol custos should be suppressed in this instance as the clef-change custos already provides the necessary information.  Is that what you'd want?

olivierberten:
Well... I thought the custo was wrong but I just realize the actual bug is the whole second line is shifted down by half a line (except for the clef)... It's been fixed on Gregobase because the gabc code wasn't really optimal but it might be worth checking how this happened...

rpspringuel:
>the whole second line is shifted down by half a line (except for the clef)

Huh?  Can you post a full MWE so that I can see this?

olivierberten:
```tex
% !TEX TS-program = lualatex
% !TEX encoding = UTF-8

\documentclass[12pt]{article}
\usepackage{geometry}
\geometry{paperwidth=16cm,paperheight=20cm}
\usepackage{gregoriotex}
\usepackage{fullpage}

\usepackage{fontspec}
\defaultfontfeatures{Ligatures=TeX}
\setmainfont{Linux Libertine O}

\pagestyle{empty}
\begin{document}

[...]

PR #1620, opened 2025-02-12
Fix bug causing staff to shift up (#1503)

davidweichiang:
This is just a one-line fix (an off-by-one error).


rpspringuel:
This looks good to me.

davidweichiang:
I also think it's ready, unless you want more documentation of the line_ids (the keys of the line_heights table) that seems to have caused some confusion.

rpspringuel:
I think I must have missed something.  What sort of confusion are you talking about?

rpspringuel:
Okay, now I see what you're saying.  Some additional documentation about how the `line_heights` table is constructed (and what each number means) is probably in order, but I'm not sure where to put it at this moment.  Logically it would be in our Lua documentation (since this is something created and used by Lua code), but as you've probably noticed our Lua code is not especially well documented (and doesn't have a designated spot).

Issue #1434, opened 2018-10-26
Custo at end of last line of a score


dpk:
For reasons of space I wish to print a chant which has two alternate endings by printing the common beginning once, followed by the two alternative endings on separate pages. At the end of this common opening, I’d like a custo to show the note the ending starts on (they both start on the same note).

I can of course use the `+` mark in my gabc score but that leaves a custo hanging in the middle of the line, which I find aesthetically unpleasing. What I'd ideally like is the custo to be pushed to the rightmost position on this final line, like all the other custos on the other lines are.

More generally, a way to do an `\hfill` to push parts of a score to the end of a line, regardless of the fact there’d be a gap in the middle, would be nice.

davidweichiang:
This appears to work:
```
%%
g(g) g(g[gv:\hfill]) (h+)
```

<img width="355" alt="Image" src="https://github.com/user-attachments/assets/b4417d53-7641-458e-be48-f956218d3e1b" />

PR #1627, opened 2025-02-17
Trim staff lines on last line

### LINKED ISSUE #1418: Adjustable line length
Right now the staff lines are drawn first on every line and their length is based purely on `\hsize` at the time the score begins (in the case of the first line, this is suitably shortened to account for the initial).  While I don't think there is any way to get around the "drawn first" principle, I think that the basis of their length can be adjusted, allowing for more dynamic line lengths.

What I think should be possible is the following:

1) certain element can be known to end a line (e.g. an end of line custos or a final bar line) and should be able to "save" the value of its right most extent and which line of the score its on, much like high/low elements save their heights, and this value could be used in a second pass to determine the length of the staff lines for that line

2) the ragged line break `Z` could do the same thing for the arbitrary element it is attached to when there is no end of line custos

3) the `\grechangenextscorelinedim` command could be expanded to allow the user to manually set the line length of a particular line in the next score.  In this case, decision on where to break the syllables on the line should also be affected.

I think idea 3 is actually fairly independent of the other two and may be the easiest to implement, but it also leaves most of the work on the user.

davidweichiang:
Adds an option `\gresetlastline{trimmed}` that trims the staff lines on the last line. Closes #1418.

Also removes some unneeded code related to paragraph shape, and makes `adjust_fullwidth` a little more general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Changed
 - Variable line heights are now computed in one pass instead of two. Per-line adjustments using `\grechangenextscorelinedim` and `\grechangenextscorelinecount` are also done in one pass, but only work on dimensions/counts related to line heights.
 - Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the option `\gresetlastline` was ignored; now, a final `Z` or `z` is ignored, and `\gresetlastline` determines whether the last line is ragged, justified, or trimmed.
+### Deprecated
 - The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
 
 ## [6.1.0] - 2025-02-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ### Changed
 - Variable line heights are now computed in one pass instead of two. Per-line adjustments using `\grechangenextscorelinedim` and `\grechangenextscorelinecount` are also done in one pass, but only work on dimensions/counts related to line heights.
 - Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the option `\gresetlastline` was ignored; now, a final `Z` or `z` is ignored, and `\gresetlastline` determines whether the last line is ragged, justified, or trimmed.
+- The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
 
 ## [6.1.0] - 2025-02-28
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,11 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 ## [Unreleased][CTAN]
 ### Added
 - Added new option `\gresetlastline{trimmed}`, which sets the last line ragged and also trims the staff lines. See [#1418](https://github.com/gregorio-project/gregorio/issues/1418).
+
 ### Changed
 - Variable line heights are now computed in one pass instead of two. Per-line adjustments using `\grechangenextscorelinedim` and `\grechangenextscorelinecount` are also done in one pass, but only work on dimensions/counts related to line heights.
 - Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the option `\gresetlastline` was ignored; now, if `Z` is used with `\gresetlastline{justified}` or `z` is used with `\gresetlastline{ragged}` or `\gresetlastline{trimmed}`, a warning is printed; in a future release, this will be an error.
+
 ### Deprecated
 - The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 
 ### Changed
 - Variable line heights are now computed in one pass instead of two. Per-line adjustments using `\grechangenextscorelinedim` and `\grechangenextscorelinecount` are also done in one pass, but only work on dimensions/counts related to line heights.
-- Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the option `\gresetlastline` was ignored; now, if `Z` is used with `\gresetlastline{justified}` or `z` is used with `\gresetlastline{ragged}` or `\gresetlastline{trimmed}`, a warning is printed; in a future release, this will be an error.
+- Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the appearance of the last line would sometimes depend on `Z` versus `z` and sometimes depend on `\gresetlastline{ragged}` versus `\gresetlastline{justified}`. Now, the appearance of the last line always depends on `\gresetlastline`.
 
 ### Deprecated
 - The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][CTAN]
+### Added
+- Added new option `\gresetlastline{trimmed}`, which sets the last line ragged and also trims the staff lines. See [#1418](https://github.com/gregorio-project/gregorio/issues/1418).
 
 
 ## [6.1.0] - 2025-02-28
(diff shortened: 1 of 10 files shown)


rpspringuel:
Can you make a test to illustrate this?

davidweichiang:
Done!

davidweichiang:
I think this one should be ready to go. Thanks!

rpspringuel:
Reviewing this I get some interesting failures:

gabc-output/glyphs/no-notes-2.gabc
gabc-output/glyphs/no-notes.gabc

In both cases the staff lines have been removed entirely, leaving only the initial clef (as there are no notes in these tests, hence their name).  I'm not sure we want to do that.  I could conceive of someone wanting to create staves with no notes (and maybe even no clef) to create exercise pages or a template for composing.

davidweichiang:
Evidently, I assumed that because this was adding a new feature that was off by default, the tests did not need to be run -- sorry! With the default `\grelastline` setting, the lines certainly should not disappear, so I'll have to track this down.

davidweichiang:
Here's what I think is happening in the original code:

Because the line with no notes has only two boxes (one containing the staff lines and one containing the clef), `post_linebreak` deletes it.
https://github.com/gregorio-project/gregorio/blob/18301fde54e1216c5e19d4d514f7b1fdfd63dcf3/tex/gregoriotex.lua#L738-L741

However, `post_linebreak`, instead of returning the new head of the list of lines, returns `true`, which is equivalent to returning the original head. This resurrects the deleted line. (This only happens when the deleted line is the first line.)
https://github.com/gregorio-project/gregorio/blob/18301fde54e1216c5e19d4d514f7b1fdfd63dcf3/tex/gregoriotex.lua#L879-L880

In this PR, all of the above is true, but the lines are initially made zero-width and expanded to full-width in `post_linebreak`. Since the code that does that not see the deleted line, the lines remain at zero width.

Possible solutions:
- Don't delete an empty line if it's the first line. This will preserve the original behavior, but it doesn't make that much sense to me. For example, if you wanted to make a sheet of blank lines with `(z) (z) (z) (z) (z)`, it would not work.
- Come up with a better heuristic for when to delete a line. 
- Delete all blank lines, but invent a new gabc symbol for an intentionally blank line.

It would help to know what bug 20974 was.

rpspringuel:
That's probably a tracking number from the old Gna! tracker.  We switched to GitHub when that system was shutting down in 2014, but only transferred over the open bugs.  Unfortunately I don't think any of the data from that system exists on the Internet anymore.

davidweichiang:
The last commit, which implements the first suggested solution above, makes the no-notes test pass; there are some very slight differences in another set of tests that I went ahead and accepted.

davidweichiang:
However, if a score ends with `Z` (as many of the tests do), neither the existing option `\gresetlastline{justified}` nor the new option `\gresetlastline{trimmed}` have any effect. The reason is that the final `Z` creates an empty last line, which is deleted.

rpspringuel:
>However, if a score ends with Z (as many of the tests do), neither the existing option \gresetlastline{justified} nor the new option \gresetlastline{trimmed} have any effect. The reason is that the final Z creates an empty last line, which is deleted.

Frankly, I'm inclined to think that's an inappropriate use of `Z`.  Whether the last line is set ragged, justified, or trimmed should be controlled by the TeX setting, not something in the gabc.  Otherwise you get the situation where someone tries to change the TeX setting but doesn't see their output change the way they expect.

If a score ends with `Z` the user should be understood as *asking* for a blank line at the end of the score with a penultimate line which is set raggedly.  Equivalently, a score ending with `z` is also asking for a blank line with the penultimate line being set justified. What do others think?

davidweichiang:
I think it would be cleanest not to delete empty lines at all. But a lot of existing scores that have a final `z` or `Z` would suddenly have an empty line.

If all the final `z` and `Z`s were removed, I believe there would still be unwanted empty lines. There's a penalty of 0 (`finalpenalty`) at the end of every score, and if a break occurs there then the last line is empty. I'm not sure why that penalty is there.

rpspringuel:
I think that penalty is supposed to be for page breaks, not line breaks (as there is no line to be broken at the end of the score).  The idea (if I understand correctly) is that after a score is a good place to insert a page break, but not necessarily the best (hence it defaulting to 0, rather than a negative value).  It should be issued in vertical mode, though I've never actually checked that.

davidweichiang:
A penalty for page breaks makes a lot of sense, but `finalpenalty` is definitely being used in horizontal mode. I've verified that removing it has no effect on any tests. 

I've also verified that not deleting any lines, but changing the handling of `Z` and `z` based on the flag `\ifgre@endofscore` to detect whether it's at the very end of the score, works. It breaks two tests, and I think both are improvements.

The question remains how `\gresetlastline` should interact with a final `Z` or `z`. I would be fine with ignoring a final `Z` or `z` and letting `\gresetlastline` control the appearance of the last line.

rpspringuel:
`\grechangecount` should be modified to have a special case that detects an attempt to change `finalpenalty`.  It should issue a deprecation warning in this instance (since `finalpenalty` is being eliminated, no alternative needs to be given).  The lines of code that implement this should be marked as `DEPRECATED`, even though they are new.

I agree with ignoring a final `z` or `Z` and letting `\gresetlastline` rule, but I would issue a deprecation warning if one is detected.  That way there's something in the log to alert users to the fact that they may have been relying on the previous behavior.  If possible, there should be two messages: one for `z` indicating that `\gresetlastline{justified}` should be used instead and one for `Z` indicating that `\gresetlastline{ragged}` is the new alternative.  An argument could be made that the message should be suppressed if the setting already matches what the last line calls for, but I would save that for the obsoletion stage, when we escalate from a warning to an error.

davidweichiang:
Will do for `finalpenalty`, but for final `z` and `Z`, there's a fair number of tests (32) that have this, which makes me wonder if some more thought is needed. I know that in parallel situations (e.g., Pascal not allowing trailing semicolons, JSON not allowing trailing commas, and of course LaTeX adding a blank line after trailing `\\`, which is precisely the same as the current case), I always prefer allowing the trailing separator.

Picking up on your comment about distinguishing between matching vs. clashing options, what about

| | trailing `Z` | trailing `z` |
|-|-|-|
| `justified` | warning -> error  | silent |
| `ragged` | silent | warning -> error |
| `trimmed` | silent | warning -> error |

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Variable line heights are now computed in one pass instead of two. Per-line adjustments using `\grechangenextscorelinedim` and `\grechangenextscorelinecount` are also done in one pass, but only work on dimensions/counts related to line heights.
 - Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the option `\gresetlastline` was ignored; now, a final `Z` or `z` is ignored, and `\gresetlastline` determines whether the last line is ragged, justified, or trimmed.
+- The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
```
rpspringuel, line 14: This should be in a `### Deprecated` section.

UPGRADE.md
```diff
@@ -4,6 +4,10 @@ This file contains instructions to upgrade to a new release of Gregorio.  These
 
 ## [Unreleased][develop]
 
+## [Unreleased][CTAN]
+
+- The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
```
rpspringuel: I would use a three-hash subsection header rather than a bulleted list here.

rpspringuel:
So basically, we issue warnings (and eventually errors) if the indications clash, but stay silent about indications which match?  That sounds reasonable, but I would add a note to `UPGRADE.md` about the change and advise people to remove trailing `z` or `Z` in gabc scores when they encounter them.  As for our tests, I think we should remove the trailing `z` and `Z` from all of them.

Looking at tests, I see three new failures:
gabc-output/bugs/fix-836.gabc
gabc-output/eos-coverage-2.gabc
tex-output/bar-spacing/lyric-centering.tex

The first two look like the results of your recent changes (and would match with what I expect the new behavior to be) but the third is weird.  Vertical spacing between scores has been affected in a way I would have associated with the previous PR, not this one.  Did that somehow slip through?

davidweichiang:
I think what's happening in lyric-centering.tex is that the inter-line glue of the line of text "Forced Centers," etc. was being computed using the depth of the deleted empty 3rd line in the old code (0pt), but using the depth of the correct 2nd line in the new code (2.42pt because of the descender on "g"). The depth of the descender is supposed to be subtracted from the inter-line glue, so the lines now appear slightly closer together.

CHANGELOG.md
rpspringuel: Please add a blank line between sections.

rpspringuel:
A deleted empty line would explain the changes.  Go ahead and accept those new test results.

davidweichiang:
Done. But `fix-836` ends with `(h+z)` to get a manual custos at pitch `h`. So I changed the gabc using the hack from #1434 to get rid of the final `z`, which further changes its appearance.

CHANGELOG.md
davidweichiang: Done.

davidweichiang:
On second thought -- I think your original suggestion is much simpler. I will revert to letting `\gresetlastline` override `Z/z`, update the warnings, and rerun the tests.

tex/gregoriotex-main.tex
```diff
@@ -238,7 +238,7 @@
 }%
 
 \def\GreFinalNewLine{%
```
rpspringuel, line 240: Was this previously a no-op?  Perhaps we should mark this macro for removal altogether.

davidweichiang: Yes, it was previously a no-op, generated only when the final break follows a bar. I think it might make sense to keep it or to remove it, but in any case I think that final breaks should be treated the same whether or not they follow a bar.

davidweichiang:
I'm think I'm done with the requested changes, here and in the tests.

rpspringuel:
Everything looks good to me.

davidweichiang:
Great. Is it okay to merely "discourage" final line breaks, or did you actually want to deprecate them and eventually make them an error?

rpspringuel:
I'd like to end up in a place where a final line break results in a blank staff at the end of the score for the purposes of those who might be making exercises or composing or the like.  For now, I think we should simply discourage final line breaks in the gabc by ignoring final `z` and `Z`.  Later on we can start respecting them in the new way as a new feature.  Since blank staff lines aren't a feature anyone is clamoring for, I think we can afford some transition time for people to get used to the current changes before introducing more.

davidweichiang:
Regarding blank staves, a related issue is that gregorio forbids a line break at the very beginning of the score. For an entirely blank score, I figured that `\gresetnotes{invisible}` would be a reasonable workaround.

I think this is ready to merge, then, thanks!

Issue #1628, opened 2025-02-17
Redirection warning


rpspringuel:
The warning about ignored arguments reared its head with Akira Kakuto in his testing and it bothered him enough that he emailed me a bug report (with a patch that removes the redirection).  Perhaps we need to reassess just how "harmless" this warning is and see if there isn't something we can do about it while still leaving open the option of capturing the error it's meant to capture.

Refer to #1541 and #1606 for past discussion.

rpspringuel:
Quoting Akira's messages for reference:

> I find a command line for gregorio is better if 2> %s is removed,
since message file is set by -l %s:

```--- gregoriotex.lua.orig	Thu Feb 13 07:39:06 2025
+++ gregoriotex.lua	Mon Feb 17 07:40:22 2025
@@ -1143,8 +1143,8 @@
    extra_args = extra_args..' -D'
  end

-  local cmd = string.format('%s %s -W -o %s -l %s "%s" 2> %s', gregorio_exe(),
-      extra_args, gtex_file, glog_file, gabc_file, glog_file)
+  local cmd = string.format('%s %s -W -o %s -l %s "%s"', gregorio_exe(),
+      extra_args, gtex_file, glog_file, gabc_file)
  res = os.execute(cmd)
  if res == nil then
    err("\nSomething went wrong when executing\n    '%s'.\n"
```

> In the original case, unnecessary warnings:
> Module gregoriotex Warning: ignored arguments: 2> tmp-gre\Populus...,
> etc. are shown and written in tmp-gre/...glog file, at least on
> Windows.
> If the new patch is applied, no such warnings are shown.

Second message:

> I believe that the above unnecessary warnings are shown in all
platforms in TeX Live.
In TeX Live, gregorio runs in restricted mode without
--shell-escape. In the restricted mode, all arguments are quoted.
Thus the redirection 2> foo becomes a usual argument.
I confirmed that if I execute like
> 
> `lualatex --shell-escape main-lualatex.tex`
> 
> the warnings are not shown. In this case, however, the final
argument of
> 
>  `local cmd = string.format('%s %s -W -o %s -l %s "%s" 2> %s', gregorio_exe(),
     extra_args, gtex_file, glog_file, gabc_file, glog_file)`
> 
> should be changed, for example, like
> 
>  `local cmd = string.format('%s %s -W -o %s -l %s "%s" 2> %s', gregorio_exe(),
     extra_args, gtex_file, glog_file, gabc_file, "gerror.err")`
[...]

PR #1629, opened 2025-02-17
Remove redirection in restricted mode

rpspringuel:
Since the redirection doesn't work in restricted mode, we only add it when `--shell-escape` has been given.  Further, rather than reusing the glog file, we create a special file for capturing the redirection error.  This addresses #1628.


rpspringuel:
This may be supplanted by #1632.  Converting to draft to prevent accidental merges.

Issue #1289, opened 2017-02-01
Cross compiling on Ubuntu 16+


rpspringuel:
In preping the beta I've been trying to setup cross-compiling on my Ubuntu 16.04 VM and have discovered that the mingw32 port is not in the repositories for that version of Ubuntu.  As a result, I can't build the Windows executable on that VM.  There is a mingw-w64 port, which should be able to build both 32 and 64 bit executables, but it doesn't work with our build setup as currently structured, and I'm not in a position to figure out what changes are needed at the moment.

This doesn't present too big a problem for now, as I can simply switch to my Ubuntu 14.04 VM (where the mingw32 port is readily available) and build the executable there.  However, in the long term we're probably going to have to look into switching the tool chain around for building the Windows executable.

rpspringuel:
In a similar vein, there is now a mingw-w64 cross compiler port for MacPorts, so ideally this update should allow for compilation of the Windows executable on both Mac and Linux.

Also, I found a page which shows a simple compiler selection script for mingw-w64:
https://sourceforge.net/p/mingw-w64/wiki2/UsingLinuxBinaries/

henryso:
Does this mean we would no longer support win32-only systems?  If so, do we care?

rpspringuel:
The mingw-w64 package contains both a 32 & 64 bit compiler.

rpspringuel:
The priority on this is now greatly increased as I'm having trouble getting my Ubuntu 14.04 VM to keep going.

eschwab:
On an up-to-date Linux installation I can cross-compile for Windows. However using `mingw-w64` it seems that it only supports 64bit. It might support 32bit, but that would require some more digging and testing. I think it is fine to drop building 32bit, since I presume that most users would be on 64bit boxes.

Or we could say that building on Linux for Win is 64bit only, and building on Mac for Win can do both. If I am looking at #1630 correctly, it appears that the Mac build is only targeting 32bit. It that correct?

rpspringuel:
To be honest, I wasn't trying for either, just to get it working.  This was my first foray into getting cross-compiling working on my Mac and I'm working mostly by trial and error from examples found online, not any real understanding of what I'm doing.

Checking it now, it does seem that what I get is a 32bit executable.  I suppose that I should be able to compile a 64bit executable too, but it would take some further fooling around to make it happen.

PR #1630, opened 2025-02-18
Getting Windows cross-compilation to work on my Mac

rpspringuel:
I don't know if this would solve #1289 on Linux machines, but at least on my Mac I can now build a windows executable and installer with this.

If someone with a native Linux machine could look at this and see if the same kind of thing would work on Linux hosts, I'd appreciate it.  I'm still trying to sort out some issues with my Ubuntu VMs.


eschwab:
Building on native Linux, with some tweaks I might have a working gregorio executable. Just need to test it in the Win11 box. Have not tried making the installer.

rpspringuel:
@eschwab, once you've made sure the executable works, can you double check the Linux compilation instructions (and either submit the updates, or tell me what needs to be changed and I'll do it)?

rpspringuel:
I'm particularly interested if the differ from the Mac instructions significantly.  I'm thinking of making a makefile in the Windows folder that builds the installer (like I did with the macosx folder) but that doesn't really make sense if the build process on Mac and Linux are significantly different.

rpspringuel:
Ugh.  I'm doing some testing and it seems that there might be a dll linking problem in my build.  I keep getting an error about lipssp-0.dll being missing.

rpspringuel:
Okay.  So that was a simple fix.  There was a `-static` flag missing.

eschwab:
Never having used gregorio on windows, is the gregorio-<version>.exe portable? That is can I just plop it on a flash drive and run it via powershell? Or does it need to be installed with the installed built with InnoSetup?

rpspringuel:
It's mostly portable.  It needs `gregorio-vowels.dat` to be in a location that kpse can find to work.  That means you need a minimal TeX distribution of some sort installed (so that kpse is available) and have to, at least, bring a copy of `gregorio-vowels.dat` with you onto the machine.  However, a full installation is not necessary.  If you put your test executable, `gregorio-vowels.dat`, and your sample gabc file all in the same directory, you should be able to run the test executable with PowerShell (or the Command Prompt) from that folder.

build.sh
```diff
@@ -92,7 +105,7 @@ then
   esac
   ARCHFLAGS="$ARCHFLAGS"
   CFLAGS="-arch $ARCH -g -O2 $CFLAGS"
-  LDFLAGS="-arch $ARCH $LDFLAGS" 
+  LDFLAGS="-arch $ARCH $LDFLAGS"
 fi
```
eschwab, line 110: Missing a final `fi` to close the `MINGWCROSS` test.

eschwab: nevermind, that was an issue on my end. Indentations were all messed up.

PR #1632, opened 2025-02-20
Fix some errors related to .glog file

davidweichiang:
- Make gregorio overwrite the .glog file instead of appending to it
- If there is an error opening the .glog file, write the error to stderr (previously the error could go nowhere)
- More fixes not implemented yet

Addresses #1628


rpspringuel:
I've switched this to be against release-6.1 instead of develop.  Please make sure that this hasn't broken anything.

tex/gregoriotex.lua
```diff
@@ -1160,48 +1108,37 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
         .."See the documentation of Gregorio or your TeX\n"
         .."distribution to automatize it.",
         cmd, tex.formatname, tex.jobname)
-  elseif res ~= 0 then
+  else
+    -- Copy the contents of glog_file into warnings.
+    -- Open it for writing so that the LuaTeX recorder knows that gregorio wrote to it.
     local glog = io.open(glog_file, 'a+')
     if glog == nil then
```
davidweichiang, line 1115: I'm unsure why we had to downgrade this to a warning for the gtex file, but not for the glog file.

tex/gregoriotex.lua
```diff
@@ -1237,101 +1174,80 @@ local function locate_file(filename)
...
-      gtex_filename = lfs.normalize(gtex_filename)
-      tex.print(string.format([[\input %s\relax]], gtex_filename))
-      return
-    else
-      err("The file %s does not exist", gabc_filename)
-      return
-    end
+    gabc_dir, base = string.match(gabc_file, "(.-)([^\\/]*)$")
   end
-  if not gabc_file then
```
davidweichiang, line 1308: This seems to have used the gtex file even if the original gabc file was gone. I didn't keep this case, but it could be easily added back in.

tex/gregoriotex.lua
```diff
@@ -1237,101 +1174,80 @@ local function locate_file(filename)
...
-  local cleaned_filename = input_name:gsub("[%s%+%&%*%?$@:;!\"\'`]", "-")
-  local gabc_filename = string.format("%s%s.gabc", file_dir, input_name)
-  local gabc_file = locate_file(gabc_filename)
-  local gtex_filename = string.format("%s%s-%s.gtex", file_dir, cleaned_filename,
-      internalversion:gsub("%.", "_"))
-  local gtex_file = locate_file(gtex_filename)
-  local glog_file = string.format("%s%s-%s.glog", file_dir, cleaned_filename,
-      internalversion:gsub("%.", "_"))
-  if not gtex_file then
-    clean_old_gtex_files(file_dir..cleaned_filename)
```
davidweichiang, line 1266: This was cleaning gtex files in the wrong place, and I just deleted it. But it can be put back and made to clean gtex files inside tmp-gre.

rpspringuel: What do you mean by "wrong place"?  Before the tmp-gre system, gtex files were placed in the same directory as the gabc files and all this was meant to clean-up those gtex files.  We probably should still clean those files up, or at least indicate to the users (perhaps in UPGRADE.md) that they should be deleting those files themselves.

tex/gregoriotex.lua
rpspringuel: We want to keep that case.  A "missing" gabc file may actually be one that we've purposely hidden from the document because we've made manual modifications to the gtex file.

tex/gregoriotex.lua
```diff
@@ -1237,101 +1174,80 @@ local function locate_file(filename)
...
-  if gtex_timestamp < gabc_timestamp then
-    log("%s has been modified and %s needs to be updated. Recompiling the gabc file.", gabc_file, gtex_file)
-    compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
-  elseif force_gabccompile then
+
+  -- Set up output directory
+  local output_dir = base_output_dir..sep..gabc_dir
+  info('Output directory: %s', output_dir)
+  if not lfs.exists(output_dir) then
+    local ok, message = lfs.mkdirp(output_dir)
```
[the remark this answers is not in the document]
rpspringuel: Is that new?  I'll add some Windows testing to my todo list.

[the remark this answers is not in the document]
davidweichiang: No, the manual makes it sounds old:

image

tex/gregoriotex.lua
davidweichiang: By "wrong place" I mean the "old place". I can restore the function and the call to it, and are you saying you want to clean them up in both the old place and the new place?

(My opinion is that programs should avoid removing files without asking the user.)

davidweichiang:
It's difficult to test all the possible combinations, but the core cases definitely work against `release-6.1`. Only one test is really failing, fancyhdr-hooks, which is something new, right?

tex/gregoriotex.lua
eschwab: We decided to remove these files to prevent excessive clutter of gtex files made for old versions of gregorio. Since these gtex files were created by gregorio we felt that gregorio should be allowed to remove them without user input.

tex/gregoriotex.lua
davidweichiang: Will do

rpspringuel:
Yes, fancyhdr-hooks is new. How is it failing?

davidweichiang:
fancyhdr-hooks:

image

rpspringuel:
That’s breaking fancyhdr-hooks back to what it was before the fix. Are you testing from a branch which doesn’t have that fix applied?

davidweichiang:
It appears to be because my fancyhdr is too old (v4.1, which was in TeXLive 2024) and hooks have only been around since v4.5 (Dec 2024).

tex/gregoriotex.lua
rpspringuel: I agree with @eschwab.  We create these files so we should be allowed to delete them.

davidweichiang:
> We decided to remove these files to prevent excessive clutter of gtex files made for old versions of gregorio. Since these gtex files were created by gregorio we felt that gregorio should be allowed to remove them without user input.

OK, this code is restored; actually it now deletes more files than it used to:
- .glog as well as .gtex
- in both the old location (same as .gabc) and new location (tmp-gre)
- whenever gregorio is run (previously it was just when the .gtex file was missing)

rpspringuel:
I'm updating my Windows VM tonight.  I'll do some testing tomorrow(Saturday) and let you know if I find any problems.

rpspringuel:
So, it turns out that applying 5 years of updates takes more than a day.  It's been a really long time since I used my Windows 10 VM and I didn't keep it up to date so I've spent the day downloading updates, installing them, and restarting the VM and I'm not even close to done yet (the Edge icon is still the old stylized "e" and not the newer pinwheel, just to give you an idea of how far I still have to go).  I'm going to leave the machine to do some unattended updates overnight, but tomorrow I'll snapshot it regardless of where it is in the update cycle in order to do some testing and I'll finish the updating process later.


Note to self: when I get around to creating the Windows 11 VM I need to boot it regularly to install updates.

rpspringuel:
Okay, some testing on my Mac first to establish a baseline behavior that I'll use to compare to Windows.  Here's my testing document setup, which includes scores from several locations:

[Test_gregorio_dev.zip](https://github.com/user-attachments/files/18931606/Test_gregorio_dev.zip)

Compiling from the main directory (the location of the tex file):
This one raises an error when trying to run the executable on `alongside/incipit3.gabc`. Because the path to this file is provided as a relative path with a parent `..` designator, kpse objects to writing the gtex file, though the glog file is written just fine.  It is, however, not under `tmp-gre`, but alongside it because the parent designator has been used to navigate out of `tmp-gre`.

Compiling from Test_gregorio_dev (parent of main):
Fails completely.  Paths are take as relative to working directory (not directory of Main.tex) and thus gabc files cannot be found.

Compiling from alongside:
Scores in main cannot be found.  Score in alongside is found, but kpse error rears its head.

**Note:** All these errors can be resolved by using `\gresetgregpath{{../}}` and recasting the gabc file locations relative to Test_gregorio_dev.  In this case we see `tmp-gre/alongside` being created for the score from the alongside directory.

Compiling from subdirectories:
`subdir/incipit3.gabc` coincidently works because I set up a lower level that has the same name (i.e. subdir/subdir actually exists) but the rest cannot be found.

Setting `\gresetgregpath{{../}}` doesn't work to fix these errors, but `\gresetgregpath{{../../}}` does.  This implies that the path is not actually relative to the document but to the working directory and that compiling from the parent directory worked before because kpse is capable of descending through the tree and finding files even when the top level of the path hasn't been specified and finally that compiling from alongside worked because the folder has the same relative position to the top level as main.

In all cases directory creation is never a problem (as expected).

It's almost time for Midday prayer now and TeX Live is still installing on my Windows 10 VM, so I'll do the Windows testing this afternoon.

davidweichiang:
Thanks for your more thorough testing! For you, which of these conditions works under the release-6.1 branch? For me compilation fails in main, `.`, alongside, but not in subdir -- same as what you said above.

Some partial comments:

> It is, however, not under tmp-gre, but alongside it because the parent designator has been used to navigate out of tmp-gre.

This could be resolved by creating the glog and gtex files directly in tmp-gre instead of trying to create subdirectories inside tmp-gre (which, as you've noted, fails if there is `..` in the tex file path). The downside is the possibility of filename collisions.

> Paths are take as relative to working directory (not directory of Main.tex) and thus gabc files cannot be found.

I agree that paths relative to Main.tex makes sense and is consistent with how other commands like `\include` work.

rpspringuel:
After checking, I realize that I was testing against what's installed by the TeX Live 2025 pretest (at the v.6.1.0-beta2 tag).  That's not where I thought I was.  Let me install the tip of this branch and repeat the tests to see if they are any different.

> This could be resolved by creating the glog and gtex files directly in tmp-gre instead of trying to create subdirectories inside tmp-gre (which, as you've noted, fails if there is .. in the tex file path). The downside is the possibility of filename collisions.

Filename collisions are pretty much inevitable under some circumstances unless we switch to hashing gabc files to produce (nearly) guaranteed unique gtex (and glog) names.  While the directory stuff makes filename collisions harder to trigger, I can construct directory structures that still will result in a filename collision.  Further, hashing makes it really hard to tell which gtex file corresponds to which gabc file, unless you copy the gabc file to tmp-gre and give it the appropriate hash name ([jperon/lyluatex](https://github.com/jperon/lyluatex) does something like this).

I might think about trying to strip `..` specifically out of the path before creating directories, which increases file collision possibilities, but perhaps not dramatically so.

>I agree that paths relative to Main.tex makes sense and is consistent with how other commands like \include work. EDITED: Actually it seems that \input and \include also interpret paths relative to the current working directory.

Since `\gresetgregpath` simply changes `\input@path` when we process scores, I actually would have been surprised if it worked differently.  I might like it to be relative to the project (as a personal preference), but that isn't how TeX itself works so I think it's probably something we just have to live with.

davidweichiang:
I just pushed a couple of bug fixes, and a "feature" for replacing `..` with `dotdot`.

rpspringuel:
Testing against [d3e4bd6](https://github.com/gregorio-project/gregorio/pull/1632/commits/d3e4bd6969450b482747483ea3ee49dcea417bcf)

## File locations specified relative to Main.tex
### Compiling from the main directory (the location of the tex file):
Works.  Directory structures are created as expected (alongside is under tmp-gre/dotdot).

### Compiling from Test_gregorio_dev (parent of main):
Scores cannot be found since locations are specified relative to Main.tex and not the current working directory.  Directory structure of tmp-gre is as expected, however.

### Compiling from alongside:
Scores in alongside and main are found (as they are in the same location relative to alongside as they are relative to main).  Scores in main/subdir and main/subdir are not.  Directory structure of tmp-gre is as expected.

### Compiling from subdirectories:
Scores in main and subdir are found (coincidence due to directory names). Scores in subdir/subdir and alongside are not.  Directory structure of tmp-gre is as expected.


## File locations specified with `\gresetgregpath{{../}}`
### Compiling from the main directory (the location of the tex file):
Works.  Directory structure of tmp-gre is as expected (alongside is directly under tmp-gre since we don't have `..` in the path under this condition).

### Compiling from Test_gregorio_dev (parent of main):
Works.

### Compiling from alongside:
Works.

### Compiling from subdirectories:
No scores found.  Directory structure of tmp-gre is as expected.  Changing to `\gresetgregpath{{../../}}` results in file compiling just fine.

My VM is still installing TeX Live (my main system idled the VM earlier, so the install got paused over lunch, grr...).  I'm probably a couple of hours from being able to test on Windows, so that likely means testing tomorrow.

davidweichiang:
That sounds better…some of these conditions are not expected to work, right? Are there remaining bugs?

rpspringuel:
The only ones I would expect to work are compiling from the directory of the file itself and compilations where `\gresetgregpath` has been used to compensate for an out of directory invocation.  Both of those currently work (as well as some others).

We've thought about doing something different in the past ([here's a post from Elie](https://tex.stackexchange.com/questions/232358/get-path-of-source-file-when-compiling-from-another-directory) from 2015 where he appears to have been working on this), but since this would be different from how `\input` and `\include` work, I'm not feeling any pressure to implement it.

Windows testing results:

## File locations specified relative to Main.tex
### Compiling from the main directory (the location of the tex file):
Score in main is found, others are not.  File name normalization doesn't appear to be active.  I'm getting errors involving path names with both `\` and `/` present.  Probably need to apply `lfs.normalize` to filenames before opening them and calling executable (see https://github.com/gregorio-project/gregorio/pull/1622 for example).  We also seem to be missing a path separator when making directories as `tmp-gresubdir` is created rather than `tmp-gre/subdir`.

Testing from other directories not done.  I'm out of time and the above needs to be fixed anyway, so I might as well do those tests after that fix is applied.

src/gregorio-utils.c
```diff
@@ -375,12 +375,20 @@ int main(int argc, char **argv)
     if (error_file_name) {
         gregorio_check_file_access(write, error_file_name, ERROR,
                 gregorio_exit(1));
-        error_file = freopen(error_file_name, "a", stderr);
+        /* First test if we can open error_file_name for writing, so
+           that if there is an error, we can still write the message
+           to stderr. */
+        error_file = fopen(error_file_name, "w");
```
rpspringuel, line 381: Probably need to normalize `error_file_name` before trying to open the error file.

tex/gregoriotex.lua
```diff
@@ -164,15 +163,13 @@ local function gregorio_exe()
 
     -- first look for one with the exact version
     real_gregorio_exe = 'gregorio-6_1_0-beta2' -- FILENAME_VERSION
-    local cmd = string.format([[%s -o "%%s" "%s"]], real_gregorio_exe,
-        test_snippet_filename)
-    exe_version = get_prog_output(cmd, '*line')
+    local cmd = {real_gregorio_exe, '-o', tmpname, test_snippet_filename}
```
rpspringuel, line 166: normalize `tmpname` and `test_snippet_filename` before assembling `cmd`

tex/gregoriotex.lua
```diff
@@ -1088,121 +1085,90 @@ function lfs.dirname(oldpath)
...
-    sep = "/"
-  end
-  dirpath = string.match(file_withdir, "(.*)"..sep)
-  if dirpath then -- dirpath is nil if current directory
-    filename = "^"..file_withdir:match(".*/".."(.*)").."%-%d+_%d+_%d+[-%a%d]*%.gtex$"
-    if lfs.exists(dirpath) then
-      for a in lfs.dir(dirpath) do
+local function delete_versioned_files(dir, base, ext)
+  -- Assume that dir is either empty (current directory) or ends with separator
+  filename = "^"..base.."%-%d+_%d+_%d+[-%a%d]*%."..ext.."$"
```
rpspringuel, line 1090: probably should normalize this `filename` too

tex/gregoriotex.lua
```diff
@@ -1088,121 +1085,90 @@ function lfs.dirname(oldpath)
...
   end
   if not allow_deprecated then
-    extra_args = extra_args..' -D'
+    table.insert(cmd, '-D')
   end
 
-  local cmd = string.format('%s %s -W -o %s -l %s "%s" 2> %s', gregorio_exe(),
-      extra_args, gtex_file, glog_file, gabc_file, glog_file)
-  res = os.execute(cmd)
+  table.extend(cmd, {'-W', '-o', gtex_file, '-l', glog_file, gabc_file})
```
rpspringuel, line 1124: have these filenames been normalized earlier?

tex/gregoriotex.lua
```diff
@@ -1088,121 +1085,90 @@ function lfs.dirname(oldpath)
...
         .."See the documentation of Gregorio or your TeX\n"
         .."distribution to automatize it.",
-        cmd, tex.formatname, tex.jobname)
+        table.concat(cmd, ' '), tex.formatname, tex.jobname)
   elseif res ~= 0 then
-    local glog = io.open(glog_file, 'a+')
+    err("\nSomething went wrong when executing\n    '%s'",
+        table.concat(cmd, ' '))
+  else
+    -- Open glog_file for writing so that the LuaTeX recorder knows that gregorio wrote to it.
```
rpspringuel, line 1139: Are we supposed to be opening glog?  The open statement opens gtex.

tex/gregoriotex.lua
```diff
@@ -1237,119 +1203,107 @@ local function locate_file(filename)
...
+    gabc_dir, base = string.match(gabc_file, "(.-)([^\\/]*)$")
+  end
+  local base_cleaned = base:gsub("[%s%+%&%*%?$@:;!\"\'`]", "-")
+
+  -- Find gabc file
+  gabc_file = string.format("%s%s.gabc", gabc_dir, base)
+  local gabc_found = locate_file(gabc_file)
+  
+  -- Set up output directory
+  local output_dir = base_output_dir..sep..gabc_dir
```
rpspringuel, line 1234: `sep` appears to be blank on Windows here

tex/gregoriotex.lua
```diff
@@ -1237,119 +1203,107 @@ local function locate_file(filename)
...
 end
 
-local function include_score(input_file, force_gabccompile, allow_deprecated)
-  if string.match(input_file, "[#%%]") then
+local function include_score(gabc_file, force_gabccompile, allow_deprecated)
+  local sep = ""
+  local onwindows = (os.type == "windows" or
+                     string.find(os.getenv("PATH"),";",1,true))
+  if onwindows then
+    sep = "\\"
```
rpspringuel, line 1211: Something appears to be going wrong here.  Maybe we should be getting rid of the whole sep distinction here and rely on `lfs.normalize` to make sure our path separators are correct.

davidweichiang:
Quick question about backslashes. The function `lfs.normalize` changes all `\` to `/`, and the comment says that on Windows, `/` is actually okay internally, but users might type `\` so they need to be changed to `/`. That appears to be the opposite of the strategy taken in other parts of the code, where a variable `sep` is set to `\` on Windows (or is supposed to be -- I see your comment above). Are you able to confirm that the comment on `lfs.normalize is correct, and if so, should the same strategy be applied everywhere?

tex/gregoriotex.lua
davidweichiang: Thank you, that's my fault

tex/gregoriotex.lua
davidweichiang: That's very mysterious, but as you suggested, maybe the problem can be circumvented.

davidweichiang:
Does kpse.find_file return paths with `/` or `\`?

rpspringuel:
I don't know.  I'd have to test it (and I'm headed to class right now).  If you can provide a simple test file (one that say, uses `\directlua` to call that function, I'll test it during my lunch break (otherwise it'll have to wait until the school day is over).

rpspringuel:
Windows testing against the new head:

### Compiling from main
Works.  tmp-gre is created and structured as expected (with alongside below (dotdot).

### Compiling from Test_gregorio_dev
No scores found.  tmp-gre is not created.

### Compiling from alongside
Scores in main and alongside are found.  Scores in subdir and subdir/subdir are not.  tmp-gre is created with dotdot/alongside inside, but no subdir.

### Compiling from subdir
Scores in main and subdir are found (coincidence because subdir and subdir/subdir exist and contain the same score).  Scores in subdir/subdir and alongside are not. tmp-gre is created with subdir inside, but not dotdot.

## Changing paths and setting `\gresetgregpath{{../}}`
### Compiling from main
No scores found. tmp-gre not created.  Surprised this one doesn't work.  Looking in the log I see the following error:
```
gregoriotex.lua:1054: attempt to index a nil value (local 'path')
stack traceback:
        ...xlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.lua:1054: in function 'lfs.normalize'
        ...xlive/texmf-local/tex/luatex/gregoriotex/gregoriotex.lua:1227: in field 'include_score'
```
Are we ending up with an empty path?

### Compiling from Test_gregorio_dev
Works.  tmp-gre create as expected.

### Compiling from alongside
No scores found.  tmp-gre not created.

### Compiling from subdir
Scores not found. tmp-gre not created

Changing to `\gresetgregpath{{../../}}` gets the subdir test working with tmp-gre created as expected.

davidweichiang:
> Are we ending up with an empty path?

Oh, this is happening when kpse.find_file can’t find the file, so it returns nil, and lfs.normalize is called on nil. Should be an easy fix. But I am not sure why it’s not finding the file. I will have to understand \gresetgregpath better.

davidweichiang:
Does the last commit change anything? It was certainly a bug, but I don't understand why the bug was activated in that particular case.

rpspringuel:
## Changing paths and setting \gresetgregpath{{../}}

### Compiling from main
Works.  tmp-gre has expected structure.

### Compiling from Test_gregorio_dev
Works.  tmp-gre has expected structure.

### Compiling from alongside
Works.  tmp-gre has expected structure.

### Compiling from subdir
No scores found. tmp-gre created with expected structure (but has no files in it).  Changing to `\gresetgregpath{{../../}}` makes this example work.

Those are all the results I would expect.  Barring any other objections I think this is ready for merge.  That done I'll upload a new version as a release candidate and we'll hopefully be able to wrap the release in the next couple of days.

eschwab:
Looks good to me.

Issue #1636, opened 2025-02-25
Custos before clef change + line break


davidweichiang:
In this example, the custos appears in the wrong place, depending on how much additional bottom space the following line has, compared with the current line. It seems that `\gre@calculate@additionalspaces` is being called too early (after the last note on the line, but before the custos).

```
\gresetinitiallines{0}
\centering
\begin{tabular}{p{1.5in}p{1.5in}p{1.5in}}
No low note:
\gabcsnippet{(c3) h(h) c2Z(c2Z) f(f) f(f)}
&
Low note before break:
\gabcsnippet{(c3) a(a) c2Z(c2Z) f(f) f(f)}
&
Low note after break:
\gabcsnippet{(c3) a(h) c2Z(c2Z) f(f) a(a)}
\end{tabular}
```

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/4c592e2a-d5f4-45ce-8a14-aa9bb89bd4e0" />

Issue #1633, opened 2025-02-21
Collisions with text above and below staff


davidweichiang:
Presumably both of these have similar fixes, but I haven't looked into it:

```
\documentclass{article}
\usepackage{gregoriotex}
\begin{document}

\gresetinitiallines{0}

{\gresetlineheightexpansion{uniform}
  \gabcsnippet{(c3) <alt>abovelines</alt>g(g) c4(c4)}
}

\gabcsnippet{(c2) gz0::c4(gz0::c4) e(e)}

\end{document}
```

<img width="168" alt="Image" src="https://github.com/user-attachments/assets/7b7d2d73-2f79-4d6d-acd5-9fe471736599" />

davidweichiang:
An additional case that is not fixed by #1637:

```
g(g) a+::c3(a+::c3) e(e) (::)
```

<img width="130" alt="Image" src="https://github.com/user-attachments/assets/7ccf4c7b-a5eb-4e5e-b9fd-0cf67c4b7154" />

This custos is tucked inside a `disc` node, so the Lua code that computes the top/bottom pitch does not find it.

PR #1637, opened 2025-02-25
Fix some collisions above/below staff (fixes #1634, half of #1633)

### LINKED ISSUE #1634: Missing abovelinestextheight in nabc
If a score has nabc, the abovelinestextheight is sometimes missing. Adding, for example, a low note causes abovelinestextheight to appear.

```
\documentclass{article}
\usepackage{gregoriotex}
\begin{document}

\long\def\nabcsnippet#1{%
  \directlua{gregoriotex.direct_gabc("\luatexluaescapestring{\unexpanded\expandafter{#1}}", "nabc-lines:1;", false)}%
}%

\gresetinitiallines{0}

{\grechangedim{abovelinestextheight}{1in}{scalable}%

abovelinestextheight=1in missing:
\nabcsnippet{(c3) g|pu(g|pu)}%

abovelinestextheight=1in present:
\nabcsnippet{(c3) a|pu(a|pu)}%
}
\end{document}
```

Image

davidweichiang:
The C code was telling the Lua code when to compute line heights, and it was missing some cases, causing not enough additional space to be added. Now the Lua code always computes line heights (which is simpler).

This breaks several tests: clef_change.gabc and nabc-edge-cases.gabc because the old results had bugs, but fix-569 because it uncovers a new bug (#1636).

The fact that the C code does not always know when additional space is needed points to the fact that uniform line heights probably have several bugs (i.e., whenever `\GreGlyphHeights` is called from inside the TeX code). The first half of #1633 is one of these cases, but I haven't tested the others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Fixed the placement of above-lines text (`<alt>`) relative to a note above the staff or when the number of staff lines is not 4. See [#1613](https://github.com/gregorio-project/gregorio/issues/1613) and [#1614](https://github.com/gregorio-project/gregorio/issues/1614).
 - Fixed a bug where the above-lines text (`<alt>`) could collid with a note above the staff. See [#1613](https://github.com/gregorio-project/gregorio/issues/1613).
 - Fixed a bug that could cause the clef and staff to be printed too high. See [#1503](https://github.com/gregorio-project/gregorio/issues/1503).
+- Fixed a bug that could cause insufficient additional space to be added above or below the staff. See [#1633](https://github.com/gregorio-project/gregorio/issues/1633) and [#1634](https://github.com/gregorio-project/gregorio/issues/1634).
 
 ### Changed
 - Modified gregorio to append to the log file specified as an argument and to send early messages to it.  See [#1541](https://github.com/gregorio-project/gregorio/issues/1541).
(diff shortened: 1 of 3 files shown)


davidweichiang:
I'm not sure if you're still taking bug fixes for 6.1, but this is a fairly small one.

rpspringuel:
At this point I don't want to add anything else to 6.1 unless it's a blocking bug.  We're already past the scheduled code freeze and the final pull from CTAN is Saturday (March 1).  I'm hoping that we can put a bow on 6.1 ASAP.

This PR should be against ctan, but you probably want to wait changing it until after the release as the ctan head is going to move considerably once the release is finished.

davidweichiang:
Okay (what’s the difference between develop, ctan, and release?)

eschwab:
`release-x.y` - frozen code to be tested before final release, bug fixes only.
`develop` - branch for new features, bug fixes, etc, that affect the binary.
`ctan` - same as `develop` but does not affect the binary.

CTAN (the org) only accepts updates that change binaries in preparation for the yearly release of TexLive. All other updates (changes to `*.tex` `*.lua` etc.) are accepted throughout the year. Thus we have to manage two different branches. @rpspringuel can elaborate as needed.

One could make the argument that we should have `ctan` as a stable branch similar to `master` and use a `ctan-dev` for development similar to `develop`, but that may be too redundant.

davidweichiang:
Was this closed automatically because release-6.1 was deleted? I can re-open against ctan.

eschwab:
Yes, and the option to edit the base branch is unavailable.

rpspringuel:
Yeah, I didn't realize that deleting the release-6.1 branch would do that.  Sorry.  Just file a new PR against ctan.

PR #1640, opened 2025-03-02
Windows make

### LINKED ISSUE #1635: configure.ac:36: warning: AC_PROG_LEX without either yywrap or noyywrap is obsolete
This warning shows up when running `./configure` and I don't know enough about lexers to really understand how to fix it.

The documentation for this is [here](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Particular-Programs.html#AC_005fPROG_005fLEX).

Does anyone know if we should be using the `yywrap` option or the `noyywrap` one?

### LINKED ISSUE #1648: Upgrading from 6.0.0 to 6.1.0 RC breaks Scribus (1.4.8) Integration
The newly installed 900.xml file (appearing in the contrib directory after installing 6.1.0) still references "gregorio-scribus.lua". I manually copied over the new 900.xml file from the source.zip, but gregorio still bombed out in Scribus with an unhelpful message:

> The lua interpreter ran into a problem, so the remainder of this lua chunk will be ignored.

I did a fresh install of Tex Live (2025) and that seems to have fixed things.

Long story short:
Upgrading to 6.1 appears to require a TeX Live Upgrade and the GregorioSetup-6_1_0.exe installer appears to be copying older files.

rpspringuel:
So, this has become a bit more than integrating the windows build process into the make file.  Indeed, at this point, that about the only thing I haven't quite finished.

What I'm doing is trying to organize the build process. To that end I've created a number of new targets that hopefully make sense in allowing an individual to create the parts of the build you're looking to make.  I've also started the process of taking scripts and turning them into make recipes.  This should reduce the number of files in the repository and hopefully make things less confusing.  

I know I've always kept around a couple of personal scripts that do build actions I commonly want to do that were previously complicated enough that I couldn't reliably remember all the steps.  Some of my new build targets are designed to replace those scripts.

Most of my targets are what would normally be considered PHONY targets (as they don't correspond to a particular filename) because I've tried to give them easy to remember/understand names.

I'm not done with all these steps yet, but I am far enough along that you should be able to see where I'm going and I'd appreciate any feedback.


rpspringuel:
Now Closes #1648

rpspringuel:
Okay, I've been looking at how to alter the recipes to remove the windows build dependency on `build.sh` and I've learned enough to think that it should be possible, but not enough to actually see it through to completion.  There's a whole bunch more I'd need to learn about autotools to make it happen, so I'm not sure the additional effort is worth it at this point.  What I have right now works and makes the process of building releases a whole lot simpler.  So, I'm going to call this good enough and ask people to check for any major problems before I merge.

PR #1639, opened 2025-02-28
Fix some collisions above/below staff (fixes #1634, half of #1633)

davidweichiang:
The C code was telling the Lua code when to compute line heights, and it was missing some cases, causing not enough additional space to be added. Now the Lua code always computes line heights (which is simpler).

This breaks several tests: clef_change.gabc and nabc-edge-cases.gabc because the old results had bugs, but fix-569 because it uncovers a new bug (https://github.com/gregorio-project/gregorio/issues/1636).

The fact that the C code does not always know when additional space is needed points to the fact that uniform line heights probably have several bugs (i.e., whenever \GreGlyphHeights is called from inside the TeX code). The first half of https://github.com/gregorio-project/gregorio/issues/1633 is one of these cases, but I haven't tested the others.

Issue #1488, opened 2020-02-28
Arbitrary number for \gresetinitiallines


jperon:
Woud it be possible to have more than 2 lines for the initial ?
I tried a workaround with `wrapfigure`, which sounds promising, but the length of the lines doesn't adjust, so they get out of margin :

```latex
\documentclass[12pt]{article}

\usepackage{gregoriotex,libertine,wrapfig}

\begin{document}

\begin{wrapfigure}[3]{l}{0.25\textwidth}
{\fontsize{120pt}{140pt}\selectfont I}
\end{wrapfigure}
\makeatletter\gre@noinitial\makeatother
\gabcsnippet{(c3)
()N(ed~)vo(e)cá(ehgh)bit(fd~) me,(e.) *(,)
et(fde) e(ef/hg)go(h) ex(h)áu(hhh)di(fe)am(e.f!gwh) e(hfge)um :(e.) (:)
e(fde)rí(ehh)pi(f)am(fdf) e(fgFE'f)um,(fe..) (;)
et(ef~) glo(e)ri(ef/hg)fi(g)cá(gi/jhhf)bo(h_g) e(hghf)um :(f.) (:)
lon(hg~)gi(h)tú(i)di(h)ne(hhh) di(f)é(fgFE'f)rum(fe..) (;)
ad(e!f'g)im(gf~)plé(fgE'C)bo(c!e'f) e(fgFE'f)um.(fe..) (::)
}


\end{document}
```

Perhaps it could be related to #1418

rpspringuel:
This is something that's been on my ideal list of features for a while now, but I've not had a chance to work on it.  Right now the readjustment of the line length that is necessary after the initial is handled within the manual line breaks and so it's necessary for the gabc code to contain multiple `z` or `Z` in order for multiline initials.  I'd like to move that into the automatic line breaking mechanisms to remove this dependency and make it (and the placement of the initial) more responsive to setting `\gresetinitiallines` to numbers higher than 2.

There is, however, a work around for now.  If `\gresetinitiallines{2}` is set, then the first "line" of the score can actually be more than 1 line of the score.  The initial's vertical placement has to be adjusted manually using `initialraise` but it's possible to simulate a 3 (or more) line initial in a fragile way:

```
\documentclass[12pt]{article}

\usepackage{gregoriotex,libertine,wrapfig}

\grechangestyle{initial}{\fontsize{200pt}{0pt}\selectfont}
\newdimen\initialwidth
[...]

Issue #1474, opened 2019-08-17
Commentary text overlapping with high staff elements


Nokel81:
Given the following latex:

```
\scriptsize
The mouth of the just man tells of wisdom and his tongue utters what is right. The law of his God is in his heart. Ps. Be not vexed over evildoers, nor jealous of those who do wrong.
\normalsize
\grecommentary{\linespread{1.15}Ps. 91, 13, 14; Ps. 91, 3}
\gregorioscore{Confessor/Mass1/Graduale}
```

It produces the following when compiled with LuaLaTex+se (following the set up on the website).

![image](https://user-images.githubusercontent.com/8225332/63213475-dde9c780-c0da-11e9-9eac-df8bb8862f7e.png)

fiat96:
I usually just add some \vspace .

Nokel81:
That is a potential fix in the interum, but the boundaries of a score should not be less than where items are drawn

Nokel81:
Another problem, which I think is probably worse. I that `grecommentary` sometimes overlaps with notes:

![image](https://user-images.githubusercontent.com/8225332/106397387-0fdc1000-63db-11eb-85a7-751ef4e35741.png)

rpspringuel:
The first problem (commentary overlapping with text above) is because the commentary box contents are not taken into account when computing the line height for the first line.  I'm not immediately sure why, but I'll look into this.

The second problem (commentary overlapping with high notes) is because the commentary box position does not look at what the highest note on the line is.  This information is not available from the current run because the commentary box is printed before any of the notes, and thus must be retrieved from a previous run.

Fixing the second definitely involves messing around with the lua code that stores and retrieves line height information between passes, but I'm not familiar enough with that to say how at this point.  @henryso can you provide some insight here?

Nokel81:
Thanks for the response. For the time being I added `4mm` of height to the `\grecommentary` invocation. I'll open a PR for the website to add information that `\grecommentary` has an optional argument for additional height (since I found that out by reading the release notes).

rpspringuel:
I've just put together that PR: gregorio-project/gregorio-project.github.io#102.  @Nokel81 please look it over.

rpspringuel:
The issue of commentary overlapping with the text above seems to be solved in the most recent develop.  The problem with it overlapping with high score items still exits.

davidweichiang:
Would it be reasonable to place the commentary at the same height as `<alt>` text?

[...]

PR #1641, opened 2025-03-03
One-pass computation of line spacing

### LINKED ISSUE #1633: Collisions with text above and below staff

davidweichiang:
In #1488 the question came up of whether variable line spacing could be done in a single pass. This (sorry, rather large) patch does that. The score lines are printed in TeX with possible collisions, but with different parts of the score distinguished by different attributes, so that they can be put into the proper place in Lua. The advantages of this I see are:

- With two passes, the user might forget to run a second time, or be confused by large changes between the first and second pass.
- The two-pass computation was complicated, with control passing from TeX to Lua to TeX to Lua. It was difficult to inject the `\grechangedim`s at exactly the end of a line, which led to bugs.

The disadvantages are:

- `\grechangenextscorelinedim` and `\grechangenextscorelinecount` now only work for the distances and counts used for line spacing. I guess a helpful error should be printed if the user tries to change something else.
- It might seem conceptually strange to print a bad-looking score at first (which no one would see). In particular, the translation is printed right on top of the lyrics.

All tests pass except for:
gabc-output/glyphs/clef_change.gabc : this is #1633 I think
gabc-output/glyphs/nabc-edge-cases.gabc : this is #1634
gabc-output/bugs/fix-569.gabc : this is #1633
gabc-output/lines/3-lines.gabc : this is #1636
gabc-output/lines/2-lines.gabc : #1636 again
gabc-output/lines/5-lines.gabc : #1636 again
tex-output/dominican-flats/dominican-flats.tex : this one fails for me even in `ctan`, and the "correct" output looks wrong; I must have messed it up somehow
tex-output/bar-spacing/bar-spacing-new.tex : the difference is only in the lines used for debugging, and the new lines look more consistent to me?

Supersedes #1639.
Closes #1633, #1634, #1636.

### DIFF
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -4701,9 +4701,11 @@ void gregoriotex_write_score(FILE *const f, gregorio_score *const score,
     }
     fprintf(f, "\\GreBeginScore{%s}{%d}{%d}{%d}{%d}{%s}{%u}"
             "{\\GreInitialClefPosition{%d}{%d}}%%\n",
-            digest_to_hex(score->digest), status.top_height,
-            status.bottom_height, bool_to_int(status.translation),
-            bool_to_int(status.abovelinestext),
+            digest_to_hex(score->digest),
+            status.top_height, /* DEPRECATED: these arguments can be deleted */
+            status.bottom_height, /* DEPRECATED */
+            bool_to_int(status.translation), /* DEPRECATED */
+            bool_to_int(status.abovelinestext), /* DEPRECATED */
             point_and_click_filename? point_and_click_filename : "",
             score->staff_lines, clef.line, clef.secondary_line);
     if (score->nabc_lines) {
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -4700,19 +4700,20 @@ void gregoriotex_write_score(FILE *const f, gregorio_score *const score,
         clef = score->first_voice_info->initial_clef;
     }
     /* After removing deprecated arguments, this will become:
-    fprintf(f, "\\GreBeginScore{%s}{%s}{%u}{\\GreInitialClefPosition{%d}{%d}}%%\n",
+    fprintf(f, "\\GreBeginScore{%s}{%s}{%u}{}%%\n",
             digest_to_hex(score->digest),
             point_and_click_filename? point_and_click_filename : "",
-            score->staff_lines, clef.line, clef.secondary_line); */
+            score->staff_lines); */
     fprintf(f, "\\GreBeginScore{%s}{%d}{%d}{%d}{%d}{%s}{%u}"
-            "{\\GreInitialClefPosition{%d}{%d}}%%\n",
+            "{\\GreInitialClefPosition{%d}{%d}}%%\n", /* DEPRECATED */
             digest_to_hex(score->digest),
             status.top_height, /* DEPRECATED */
             status.bottom_height, /* DEPRECATED */
             bool_to_int(status.translation), /* DEPRECATED */
             bool_to_int(status.abovelinestext), /* DEPRECATED */
             point_and_click_filename? point_and_click_filename : "",
-            score->staff_lines, clef.line, clef.secondary_line);
+            score->staff_lines,
+            clef.line, clef.secondary_line); /* DEPRECATED */
     if (score->nabc_lines) {
         fprintf(f, "\\GreScoreNABCLines{%d}", (int)score->nabc_lines);
     }
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -56,10 +56,10 @@ typedef struct gregoriotex_status {
     signed char bottom_height;
 
     /* indicates if there is a translation on the line */
-    bool translation;
+    bool translation; /* DEPRECATED */
 
     /* indicates if there is "above lines text" on the line */
-    bool abovelinestext;
+    bool abovelinestext; /* DEPRECATED */
 
     bool suppressed_custos;
 } gregoriotex_status;
@@ -4699,10 +4699,15 @@ void gregoriotex_write_score(FILE *const f, gregorio_score *const score,
     if (score->first_voice_info) {
         clef = score->first_voice_info->initial_clef;
     }
+    /* After removing deprecated arguments, this will become:
+    fprintf(f, "\\GreBeginScore{%s}{%s}{%u}{\\GreInitialClefPosition{%d}{%d}}%%\n",
+            digest_to_hex(score->digest),
+            point_and_click_filename? point_and_click_filename : "",
+            score->staff_lines, clef.line, clef.secondary_line); */
     fprintf(f, "\\GreBeginScore{%s}{%d}{%d}{%d}{%d}{%s}{%u}"
             "{\\GreInitialClefPosition{%d}{%d}}%%\n",
             digest_to_hex(score->digest),
-            status.top_height, /* DEPRECATED: these arguments can be deleted */
+            status.top_height, /* DEPRECATED */
             status.bottom_height, /* DEPRECATED */
             bool_to_int(status.translation), /* DEPRECATED */
             bool_to_int(status.abovelinestext), /* DEPRECATED */
--- a/tex/gregoriotex-main.tex
+++ b/tex/gregoriotex-main.tex
@@ -394,14 +394,12 @@
   \fi%
   \gre@debugmsg{annotation}{Width check completed.}%
   % Now compute the raise
-  \gre@dimen@temp@five = \dimexpr(\gre@dimen@textlower+\gre@space@dimen@initialraise)\relax%
+  \gre@dimen@temp@five = \dimexpr(\gre@space@dimen@spacebeneathtext+\gre@space@dimen@initialraise)\relax%
   \ifnum\gre@count@initialposition=0\relax
     \advance\gre@dimen@temp@five by 
       \dimexpr(\gre@dimen@staffheight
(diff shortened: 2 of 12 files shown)


rpspringuel:
On the tests:

gabc-output/glyphs/clef_change.gabc : The new results look correct, while the old ones have a bad collision.
gabc-output/glyphs/nabc-edge-cases.gabc : I can confirm this fails, but the new results look fine to me.
gabc-output/bugs/fix-569.gabc : Again, new results seem to be better.
gabc-output/lines/3-lines.gabc : Definately a fix for the custos position.
gabc-output/lines/2-lines.gabc : Same as for 3-lines
gabc-output/lines/5-lines.gabc : Same as for 3-lines
tex-output/dominican-flats/dominican-flats.tex : This one doesn't fail for me.
tex-output/bar-spacing/bar-spacing-new.tex : I agree, the new results look like they are placing the debugging lines better.

tex-output/hidelyricsnotes/text-only.tex fails for me.  The document doesn't compile and the log indicates theirs a problem trying to do math that probably shouldn't be done in this case: `gregoriotex.lua:744: attempt to perform arithmetic on a nil value (field 'glyph_top')`

Some comments:
>With two passes, the user might forget to run a second time, or be confused by large changes between the first and second pass.

Multiple passes are a standard part of preparing TeX documents, so this would only be a factor for those who are new to the ecosystem.

That said, getting rid of requiring the second pass ourselves could be a huge time savings.  Loading GregorioTeX takes a substantial amount of time so while we probably can't get rid of the need for a second pass all by ourselves, if we can make it so that we are requiring the second pass, then we at least make the single-pass documents using Gregorio possible.  If we can solve some bugs at the same time, all the better.

> `\grechangenextscorelinedim` and `\grechangenextscorelinecount` now only work for the distances and counts used for line spacing. I guess a helpful error should be printed if the user tries to change something else.

The documentation indicates that `\grechangenextscorelinedim`is only supposed to work with `spaceabovelines`, `spacebeneathtext`, and `spacelinestext`.  That it worked with other distances is therefore, an undocumented feature, so your "restriction" would actually be an expansion of what it is documented to work with.

Besides, in both cases, I think this is what we originally intended both of these macros to be for.

>It might seem conceptually strange to print a bad-looking score at first (which no one would see). In particular, the translation is printed right on top of the lyrics.

This is hardly any more strange than the intermediate document produced in the middle of a multi-pass system.

rpspringuel:
Oh, and if you could create a parallel branch in the test repository with the new test results that would be appreciated (at least for the tests where we know the new result is because we're fixing the bugs).

davidweichiang:
> tex-output/dominican-flats/dominican-flats.tex : This one doesn't fail for me.

It must be because I'm working off of https://github.com/gregorio-project/gregorio-test/pull/360 and it's gotten out of sync.

> tex-output/hidelyricsnotes/text-only.tex fails for me. The document doesn't compile and the log indicates theirs a problem trying to do math that probably shouldn't be done in this case: gregoriotex.lua:744: attempt to perform arithmetic on a nil value (field 'glyph_top')

My fault, that one fails for me too, but it should work now with the last commit.

davidweichiang:
There's still a command `\greskipheightcomputation` that still does something, but not what it says. What should it be called instead?

davidweichiang:
Since the speed difference due to `\greskipheightcomputation` seemed to be negligible (<0.1%), I took the liberty of removing it.

rpspringuel:
I was away for the weekend and have been sick since I got back, so I haven’t had a chance to review anything posted since last Wednesday. I’m hoping to feel better and get back to work by Monday. 

davidweichiang:
With the last commit, three additional tests fail:
tex-output/commentary/commentary.tex: because the commentary height takes note height into account (#1474)
tex-output/PopulusSion/PopulusSion.tex: because a small bug was causing the commentary to be printed loo low
tex-output/per-line-spaces/per-line-spaces.tex: same as previous

davidweichiang:
> I was away for the weekend and have been sick since I got back, so I haven’t had a chance to review anything posted since last Wednesday. I’m hoping to feel better and get back to work by Monday.

No worries and hope you feel better soon!

tex/gregoriotex-spaces.tex
```diff
@@ -1590,6 +1590,9 @@
 % #1 : line number
 % #2-#4 : same as #1-#3 of \grechangedim
 \def\grechangenextscorelinedim#1#2#3#4{%
+  \IfStrEqCase{#2}{{abovelinestextheight}{}{noteadditionalspacelinestext}{}{spaceabovelines}{}{abovelinestextraise}{}{spacelinestext}{}{spacebeneathtext}{}{translationheight}{}}[%
```
rpspringuel, line 1590: Just to make sure I'm reading this right, you're using `\IfStrEqCase` to construct a `not (... and ... and ...)` sequence that will raise an error if this macro is called with an unsupported distance.

tex/gregoriotex-main.tex
```diff
@@ -1117,10 +1067,10 @@
 
 %macro called at the beginning of a score
 % #1 is the gabc score id
-% #2 is the high height
-% #3 is the low height
-% #4 is 1 if there is a translation somewhere
-% #5 is if 1 if we have space above the staff
+% #2 is the high height (currently ignored)
```
rpspringuel: Can you foresee a future use for these arguments?  If not, I'd mark them as deprecated, not just ignored.  Once this PR is approved and merged, then we can work on the appropriate PR for the develop branch which will actually remove them (since they'd have to be removed from the executable, that's not a ctan eligible change).

rpspringuel:
I agree that the newly broken tests are all fixes to something which is currently wrong.

tex/gregoriotex-spaces.tex
davidweichiang: That's right.

davidweichiang: Oh, actually, it's `if not (... or ... or ...) then print error`.

tex/gregoriotex-main.tex
davidweichiang: I don't foresee a future use, no. How do I mark them as deprecated?

rpspringuel: I've updated the Wiki page on the Deprecation process to cover this case: https://github.com/gregorio-project/gregorio/wiki/Deprecation-Process

tex/gregoriotex-spaces.tex
rpspringuel: Yeah, that's what I meant.

davidweichiang:
Done, I think

src/gregoriotex/gregoriotex-write.c
```diff
@@ -4701,9 +4701,11 @@ void gregoriotex_write_score(FILE *const f, gregorio_score *const score,
     }
     fprintf(f, "\\GreBeginScore{%s}{%d}{%d}{%d}{%d}{%s}{%u}"
```
rpspringuel, line 4707: I'd put a deprecation comment here indicating what the string of arguments will look like after the deprecated ones are removed.

davidweichiang: OK (also I wrote comments on two struct fields that can be removed).

davidweichiang:
I won't tinker with this anymore -- hopefully it's ready to go?

rpspringuel:
In reviewing this I've come across two unexplained test failures:

gabc-output/queuelength2_legacy.gabc
gabc-output/queuelength2.gabc

In both cases only the last three staves of the second page are effected.  The third to last stave is pushed down (though its lyrics are pushed up) and the last two staves are pushed up (by the same amount as the lyrics in the third to last stave, from the looks of it).

All the other test changes look good and are taken care of by the PR https://github.com/gregorio-project/gregorio-test/pull/373

davidweichiang:
Thanks for catching this -- I promised I wouldn't tinker any more and then I did! I believe this is a bug fix.

The first note of this score is below the staff (c=3), and I didn't fully look into why, but in `ctan` and in all commits of this PR up to the last one, the attribute `\gre@attr@glyph@bottom`=3 reappears in various places throughout the whole score. Ordinarily this seems to be harmless, but on line 14 (third from last), and again I didn't fully look into why, the 3 actually appears as an attribute on an hbox that matters. The line spacing algorithm thinks this line has a `c` note on it, and the lyrics are typeset too low.

image

I believe the shift in the last two staves is because in the old code, line 14's height was greater than `\baselineskip`, so `\lineskip` (=1pt, I think) was inserted.

In the last commit, `\gre@attr@glyph@top` and `\gre@attr@glyph@bottom` are reset after each glyph, so the problem goes away.

image

rpspringuel:
That explains why the lyrics (and the staves below) move up, but why does the 3rd to last set of staff lines move down?  There doesn't appear to be anything on the 4th to last line that would require extra space below that stave nor anything on the 3rd to last line that would require extra space above that line.

davidweichiang:
That's because the baseline of each staff is the baseline of the lyrics (or translation if any), and TeX tries to ensure that the distance between baselines is `\baselineskip`. So lyrics moving up really means staff moving down.

(I think a case could be made for the baseline to instead be the lowest staff line, but that's a separate issue!)

rpspringuel:
This is why I find `\baselineskip` so confusing.

Okay, so it looks like everything this good at this point, just need to have those altered tests accepted into the test repository.

davidweichiang:
Done, thanks!

Issue #1652, opened 2025-04-05
Error on blank line (ZZ)


davidweichiang:
The following
```
\documentclass{article}
\usepackage{gregoriotex}
\begin{document}
\gabcsnippet{(c3) (g) (Z) (Z) (g)}
\end{document}
```
was meant to test whether a blank line could be inserted in the middle of a score. But it fails rather spectacularly.  This happens with or without #1627. The error message "Can't raise/lower a rule (this shouldn't happen)" was introduced recently in #1641 and is probably my fault.

davidweichiang:
Even under 6.0.0, I get this message:

```
warning  (nodes): attempt to copy free hlist node 2487, ignored

warning  (nodes): attempt to copy free hlist node 2496, ignored

warning  (nodes): attempt to copy free hlist node 2512, ignored

warning  (nodes): attempt to copy free hlist node 2187, ignored

warning  (nodes): attempt to copy free hlist node 636, ignored

warning  (nodes): attempt to copy free hlist node 652, ignored

warning  (nodes): attempt to copy free hlist node 666, ignored

warning  (nodes): attempt to copy free hlist node 2172, ignored

warning  (nodes): attempt to copy free vlist node 1922, ignored

warning  (nodes): attempt to copy free rule node 627, ignored

warning  (nodes): attempt to copy free rule node 680, ignored

warning  (nodes): attempt to copy free rule node 711, ignored

warning  (nodes): attempt to copy free rule node 2149, ignored

warning  (nodes): attempt to copy free hlist node 2059, ignored

warning  (nodes): attempt to copy free vlist node 2024, ignored

warning  (nodes): attempt to copy free rule node 1658, ignored

warning  (nodes): attempt to copy free rule node 1632, ignored

warning  (nodes): attempt to copy free rule node 1996, ignored

warning  (nodes): attempt to copy free rule node 2010, ignored

warning  (nodes): attempt to copy free hlist node 3038, ignored
```

and the PDF looks like this

<img width="375" alt="Image" src="https://github.com/user-attachments/assets/68179e8a-d21c-4338-8a41-7f6a5c5873a1" />

That stray horizontal rule is the one that the "Can't raise/lower a rule" message is complaining about.

davidweichiang:
This was closed by #1653.

PR #1653, opened 2025-04-06
Fix bug causing an error on scores with two line breaks in a row

davidweichiang:
Fixes issue #1652.

The root of the problem seems to be a bug in LuaTeX. The following MWE illustrates the bug:
```
\localrightbox{xyz}%
\hfill\break
\localleftbox{abc}%
\bye
```

This produces the errors
```
warning  (nodes): attempt to copy free hlist node 399, ignored

warning  (nodes): attempt to copy free hlist node 369, ignored
```
and the `\localleftbox` is not updated.

Here, the bug was happening when the `\localleftbox` was updated on an empty line. The simplest fix is just to not update the `\localleftbox` (it appears not to be needed after #1641).


rpspringuel:
This looks good to me.

rpspringuel:

PR #1654, opened 2025-04-06
Change mora_vposition from a bitfield to its own field

### LINKED ISSUE #1642: Punctum mora positioning not operative
This issue concerns the alignment of the punctum mora, implemented in #386.

I previously posted this issue #1534 using version 5.2.1.  I closed the issue, as I could not replicate it after updating to version 6.0.0.  Now, having updated to 6.1.0, it is a problem again.

Typing .1 does still force a punctum mora up, but .0 is also forcing it up, instead of down.

MWE:

\documentclass[latin]{book}
\usepackage{gregoriotex}
\begin{document}
	\gabcsnippet{(c4) Hic(iih.) est(iih.0) tex(iih.1)tus(h.). (:) Tex(efd.)tus(efd.0) e(efd.1)ti(d)am(d) hic.(d.)}
\end{document}

Output:
Image

Here, I would expect the second punctum mora on each half to be aligned down.

davidweichiang:
…because it is signed and on some platforms, casting it to unsigned was causing problems (#1642).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][develop]
-
+### Fixed
+- Fixed a bug that could cause a punctum mora that is supposed to be below the line (`.0`) to appear above the line. This bug was platform-dependent and was observed on a Windows system.
 
 ## [Unreleased][CTAN]
 ### Added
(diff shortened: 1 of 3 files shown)


eschwab:
If it woks with both GNU and non-GNU compilers it is fine with me. Would probably need a changelog entry for the benefit of Windows users.

davidweichiang:
Done. I also noticed that the reference does not document `.0` or `.1`, so I added that, and then I noticed this:

image

That's a bug, right?

eschwab:
`(f..)` also produces similar results which looks like a bug to me as well.

davidweichiang:
I filed the double dot issue as a separate issue #1656, since the fix would be against `ctan` instead of `develop`. I believe this PR suffices for closing #1642.

eschwab:
All seems in order for me.

Issue #1631, opened 2025-02-18
Clarifying line spacing


davidweichiang:
I've been trying to understand the various line spacing options and it seems I'm not the only one to find them challenging (#1375, #1528, https://groups.google.com/g/gregorio-users/c/_K5qfmGVunI). I had to draw a picture:

<img width="610" alt="Image" src="https://github.com/user-attachments/assets/1472d4e0-f0fb-4955-9675-f386be1a641a" />

The height of the line is the maximum of the three heights (1), (2), and (3). Spaces that default to less than 1pt have been increased to 1pt for visibility.

I'd like to suggest adding this picture to the documentation, but want to check that it's correct, and also whether this was really how it was meant to be. Specifically:

1. I'm surprised that `abovelinestextheight` is in (1), where it doesn't actually reach as high as the abovelinestext. Would it make more sense to move it to (2) instead, starting from the baseline of the abovelinestext and creating a little bit of whitespace at the top?

2. The height (2) includes 4 copies of `interstafflinespace`, which looks like an off-by-one error to me (as noted [here](https://github.com/gregorio-project/gregorio/pull/1616/files#r1949391268)). Can the extra `interstafflinespace` be deleted, and the default `abovelinestextraise` be increased to compensate?

3. The documentation for `noteadditionalspacelinestextthreshold` says that it's "the number of low notes which will add space between the lines and the lyrics (`additionalbottomspace`)." To me that sounds like a higher threshold leads to less additional space, but it's actually the opposite (higher threshold leads to more additional space).

4. Because the default `additionaltopspacenabcthreshold` of 4 is so high, high notes collide with nabc neumes.

rpspringuel:
We definitely need some additional clarity here, but it's something that I won't have time to look at until after the release is finished.

davidweichiang:
Understood! As a result of digging into this, I've found a few bugs (besides the one above, which I'm thinking there might be an explanation for?). I'll make separate issues for those.

Issue #1375, opened 2017-09-04
Interaction between `baselineskip` and `spacebeneathtext` not clear


rpspringuel:
This may simply be a defect in my understanding, but the interaction between these two parameters isn't clear to me and leaves me wondering if it's buggy.  If some else understands the behavior, I would appreciate an explanation (which I think should be added to the documentation).

eroux:
baselineskip is a convoluted LaTeX mechanism, you can read about it [here](https://books.google.fr/books?id=72wKBwAAQBAJ&pg=PA93&lpg=PA93&dq=baselineskip+lineskip&source=bl&ots=_nZtjcKeZH&sig=DsvhRoLkW5ICY6VrmpB5pQ3ac54&hl=fr&sa=X&ved=0ahUKEwiU8oyHu43WAhWMmbQKHQFkBDwQ6AEINjAC#v=onepage&q=baselineskip%20lineskip&f=false), spacebeneathtext is just a space that is added under every line of text in Gregorio. The two don't have much in common (if anything at all)... do you have a more precise example of your puzzlement?

rpspringuel:
See the following example in which `baselineskip` is set to a large value (to exaggerate the effect)  and then `spacebeneathtext` is varied:

```
\documentclass[12pt]{article} % use larger type; default would be 10pt

% usual packages loading:
\usepackage{fullpage}
\usepackage{gregoriotex} % for gregorio score inclusion

\begin{document}
\grechangedim{baselineskip}{2cm}{scalable}

\grechangedim{spacebeneathtext}{0cm}{scalable}
\gabcsnippet{(c3)Herr,(e) ich(f) {traue auf}(hr0) || <ul>dich</ul>,(i) { }(fr) { }(f.) *(;) (z) {lass mich nimmermehr zu-}(hr0) || schan(g)den(h) | wer(f) {-}(fr) den.(ee) (::)}
\grechangedim{spacebeneathtext}{0.25cm}{scalable}
\gabcsnippet{(c3)Herr,(e) ich(f) {traue auf}(hr0) || <ul>dich</ul>,(i) { }(fr) { }(f.) *(;) (z) {lass mich nimmermehr zu-}(hr0) || schan(g)den(h) | wer(f) {-}(fr) den.(ee) (::)}\grechangedim{spacebeneathtext}{0.5cm}{scalable}
[...]

PR #1659, opened 2025-04-10
Clarify vertical spacing

### LINKED ISSUE #1374: Strange behavior when `spacebeneathtext` is set to negative value
In the following MWE, the score, as a whole, gets pushed down as a result of the negative value for `spacebeneathtext` but the lines do not get any closer together.

```
\documentclass[12pt]{article}

\usepackage{gregoriotex}

\begin{document}
text needed to show how a negative spacebeneathtext value will move the whole score

\grechangedim{spacebeneathtext}{-2cm}{scalable}
\gabcsnippet{(c3)Herr,(e) ich(f) {traue auf}(hr0) || <ul>dich</ul>,(i) { }(fr) { }(f.) *(;) (z) {lass mich nimmermehr zu-}(hr0) || schan(g)den(h) | wer(f) {-}(fr) den.(ee) (::)}

\end{document}
```

What should the behavior here be?  Should the lines be pushed into overlapping?  Or perhaps we need to simply add a check to make sure that `spacebeneathtext` is not being set to a negative value?

davidweichiang:
This patch addresses #1631 by making two changes to vertical spacing and adding more detailed documentation about vertical spacing. The two changes unfortunately will break some existing scores, but hopefully make spacing less confusing and easier to extend in the future.

A line without above-lines text looks like this. In all these pictures, the following lengths have been changed from their defaults:

```
\grechangedim{spaceabovelines}{0.2cm}{scalable}
\grechangedim{spacebeneathtext}{0.2cm}{scalable}
\grechangedim{abovelinestextraise}{0.2cm}{scalable}
```

image

But with above-lines text, the picture becomes more complicated:

image

The total height is the maximum of the two heights shown.

The first problem is that the docs say that `abovelinestextraise` is the "Height of the text above the note line." To me, that sounds like the distance from the top staff line to the baseline of the above-lines text. But there's actually an extra `interstafflinespace` in there, which I'm guessing was an off-by-one error.

The second problem is that the docs say that `abovelinestextheight` "must be bigger than the text for it to be taken into consideration," but in this example, one would have to increase `abovelinestextheight` by quite a bit more than that before seeing any effect. The reason is that the total height computation leaves out `abovelinestextraise`, and I'm guessing this was just an oversight.

Under this patch, the picture becomes:

image

It makes two changes:

1. `abovelinestextraise` is measured from the top staff line. The default `abovelinestextraise` is changed from -0.1cm to 1.7351cm to produce the same result as before (all tests pass). But any code that sets `abovelinestextraise` explicitly would have to be corrected.

2. `abovelinestextraise` is included in the computation of the total height. Decreasing the default `abovelinestextheight` to compensate (0.12649cm) makes all tests pass, but I think the current default of 0.3 cm is good, and adds a little breathing room above the above-lines text.

The new documentation is on pages 77-82 of GregorioRef.pdf. It goes into more detail about additional space above/below the staff, and about `baselineskip` and `lineskip`.

Closes #1374, #1375, #1631.

### DIFF
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -1776,7 +1776,7 @@ The space beneath the text.
 \end{gdimension}
 
 \begin{gdimension}{abovelinestextraise}
-Height of the text above the note line.
+Distance from the top staff line to the baseline of the above-lines text.
 \end{gdimension}
 
 \begin{gdimension}{abovelinestextheight}
--- a/doc/Command_Index_User.tex
+++ b/doc/Command_Index_User.tex
@@ -1896,6 +1896,10 @@ Space after a clef at the beginning of a line, when the clef and first note are 
 \newcommand{\lab}[1]{\scalebox{0.5}{\tt\small #1}}%
 
 \gresetinitiallines{0}
+\gresetlyriccentering{firstletter}
+\gresetlastline{justified}
+\grechangestyle{abovelinestext}{\rmfamily\itshape}
+\grechangestyle{translation}{\rmfamily\itshape}
 
 A staff has text (lyrics) below it, and optionally another line of translation below the text. Above the staff, there can also be a line of text (called the above-lines text) or NABC neumes. The vertical position and height of all these parts are shown in Figure~\ref{fig:vertical-spacing-within-staff} (with \texttt{spaceabovelines} and \texttt{spacebeneathtext} set to 0.2 cm for clarity). The staff together with all the text and neumes above and below it form one box, whose topline and baseline are drawn in red.
 
@@ -1990,6 +1994,7 @@ Note that if there are high notes, increasing \texttt{spaceabovelines} may not a
 
 \begin{center}
 \rowcolors{1}{white}{white}%
+\ttfamily
 \begin{tabular}{ccccc}
   \lines{\gabcsnippet{i[0 steps](i)}} &
   \lines{\gabcsnippet{j[0 steps](j)}} &
@@ -2004,6 +2009,7 @@ Note that if there are high notes, increasing \texttt{spaceabovelines} may not a
 High notes also cause the above-lines text, if any, to be raised. The above-lines text has its own threshold, \texttt{additionaltopspacealtthreshold} (default 0):
 \begin{center}
 \rowcolors{1}{white}{white}%
+\ttfamily
 \begin{tabular}{ccccc}
   \lines{\gabcsnippet{<alt>alt</alt>i[0 steps](i)}} &
   \lines{\gabcsnippet{<alt>alt</alt>j[0 steps](j)}} &
@@ -2021,6 +2027,7 @@ Because the above-lines text and the topline have different thresholds, they may
   
 \begin{center}
 \rowcolors{1}{white}{white}%
+\ttfamily
 \begin{tabular}{ccccc}
   \lines{\gabcsnippet{<alt>alt</alt>i(i)}} &
   \lines{\gabcsnippet{<alt>alt</alt>j(j)}} &
@@ -2037,14 +2044,14 @@ NABC neumes use a different threshold, \texttt{additionaltopspacenabcthreshold}.
 \verb|\grechangedim{abovelinestextheight}{0.1cm}{scalable}| \\
 \grechangedim{abovelinestextraise}{0.8cm}{scalable}%
 \verb|\grechangedim{abovelinestextraise}{0.8cm}{scalable}|%
-\gresetlyriccentering{syllable}%
 \begin{center}
 \rowcolors{1}{white}{white}%
+\ttfamily
 \begin{tabular}{ccccc}
   \lines{\nabcsnippet{m|vi[0 steps](m|vi)}} &
   \lines{\nabcsnippet{m_1|vi[0 steps](m_1|vi)}} &
-  \lines{\nabcsnippet{m'1|vi[1 step](m'1|vi)}} &
-  \lines{\nabcsnippet{m_1'1|vi[2 steps](m_1'1|vi)}}
+  \lines{\nabcsnippet{m<v>\textquotesingle</v>1|vi[1 step](m'1|vi)}} &
+  \lines{\nabcsnippet{m_1<v>\textquotesingle</v>1|vi[2 steps](m_1'1|vi)}}
 \end{tabular}
 \end{center}
 \end{quote}
@@ -2057,6 +2064,7 @@ The minimum value of \texttt{noteadditionalspacelinestextthreshold} is currently
 
 \begin{center}
 \rowcolors{1}{white}{white}%
+\ttfamily
 \begin{tabular}{ccccc}
   \lines{\gabcsnippet{e[0 steps](e)}} &
   \lines{\gabcsnippet{d[0 steps](d)}} &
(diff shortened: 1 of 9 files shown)


davidweichiang:
Closes #1660.

doc/Command_Index_User.tex
```diff
@@ -1885,25 +1857,301 @@ \subsection{Distances}\label{distances}
...
+\end{gdimension}
+
+\begin{gdimension}{translationheight}
+The space between the baseline of the text and the baseline of the translation.
+\end{gdimension}
+
+\begin{gdimension}{spacebeneathtext}
+The space beneath the text (and translation, if any).
+\end{gdimension}
+
```
rpspringuel, line 1981: When I create the pdf, the entry for `spacebeneathtext` is forced to a new page when there's plenty of room for it on the previous page.  We need to do something to encourage page breaking at this point.

doc/Command_Index_User.tex
```diff
@@ -1885,25 +1857,301 @@ \subsection{Distances}\label{distances}
...
+A legacy option is to change the amount of additional bottom space per step by setting \verb|\gresetnoteadditionalspacelinestext{manual}| and the following:
+
+\begin{gdimension}{noteadditionalspacelinestext}
+The per-step additional space between lines and the bottom of the text, when using \verb|\gresetnoteadditionalspacelinestext{manual}|.
+\end{gdimension}
+
+\paragraph{Vertical spacing between staves}
+
+The standard rules for paragraph and line spacing apply to scores and staves.
```
rpspringuel: While the standard rules apply, the below listed distances affect how the rules are applied inside a score.  I think that distinction should be spelled out for users who might not be familiar with how the TeX rules work.  Maybe something like:

> The standard rules for paragraph and line spacing apply to scores and staves.  However, the distances which control those rules are different inside a score from in the regular text.  The values listed below will be used inside a score and can be changed with `\grechangedim` without affecting how text outside of a score is typeset.

rpspringuel:
These changes affect several tests.  I didn't see anything that stood out to me as problematic, but I'd like you to check that the changes are what you're expecting.

davidweichiang:
These tests fail for me:

gabc-output/glyphs/clear.gabc
gabc-output/glyphs/nabc-edge-cases.gabc
gabc-output/glyphs/above_lines_text.gabc
gabc-output/FactusEst.gabc
gabc-output/bugs/fix-1613.gabc
gabc-output/bugs/fix-1015.gabc
gabc-output/bugs/fix-347.gabc
gabc-output/bugs/test98.gabc
gabc-output/soft-alterations.gabc
gabc-output/heights_legacy.gabc
gabc-output/heights.gabc
gabc-output/high_clef.gabc
tex-output/bugs/fix-1614/fix-1614.tex
tex-output/hideabovelinestext/hideabovelinestext.tex
tex-output/bar-spacing/end-of-line.tex
tex-output/bar-spacing-old/bar-spacing-old.tex
tex-output/bar-spacing/bar-spacing-new.tex

All of these have either above-lines text or NABC, and the differences are all slight increases in the space above-lines text or NABC. As mentioned above, there's a setting for abovelinestextheight that makes all the tests pass. But I think it's better to have that little extra space, as long as you're okay with changing existing scores.

doc/Command_Index_User.tex
davidweichiang: Done

doc/Command_Index_User.tex
davidweichiang: That page break was really weird. I inserted a \goodbreak to fix it and hope that's ok.

rpspringuel: Looks good now.

doc/Command_Index_User.tex
```diff
@@ -1885,25 +1857,316 @@ \subsection{Distances}\label{distances}
...
+\verb|\grechangedim{abovelinestextheight}{0.1cm}{scalable}| \\
+\grechangedim{abovelinestextraise}{0.8cm}{scalable}%
+\verb|\grechangedim{abovelinestextraise}{0.8cm}{scalable}|%
+\gresetlyriccentering{syllable}%
+\begin{center}
+\rowcolors{1}{white}{white}%
+\begin{tabular}{ccccc}
+  \lines{\nabcsnippet{m|vi[0 steps](m|vi)}} &
+  \lines{\nabcsnippet{m_1|vi[0 steps](m_1|vi)}} &
+  \lines{\nabcsnippet{m'1|vi[1 step](m'1|vi)}} &
```
rpspringuel: The straight quote in the lyrics of this and the next snippet are being converted to curly quotes in the output.  Please make use of `\textsinglequote` in the lyrics to get the quote to print correctly.

rpspringuel:
> All of these have either above-lines text or NABC, and the differences are all slight increases in the space above-lines text or NABC. As mentioned above, there's a setting for abovelinestextheight that makes all the tests pass. But I think it's better to have that little extra space, as long as you're okay with changing existing scores.

As long as the changes are expected given the changes you're submitting here, I don't have a problem with them.  I just wanted to make sure the changes in the tests were all in accord with the expected differences these changes would cause.

doc/Command_Index_User.tex
davidweichiang: Done (and changed font to match similar examples elsewhere)

doc/Command_Index_User.tex
```diff
@@ -2093,11 +2099,11 @@ \subsubsection{Vertical spacing}
 \end{gdimension}
 
 \begin{gdimension}{lineskiplimit}
-The effective \verb=\lineskiplimit= inside of a score. If two consecutive staves have baselines \texttt{baselineskip} apart and the distance between the first staff's bottomline and the second staff's topline is less than this, then \texttt{lineskip} is added between them.
+The effective \verb=\lineskiplimit= inside of a score. If two consecutive staves have baselines \texttt{baselineskip} apart and the distance between the first staff's bottome and the second staff's top is less than this, then \texttt{lineskip} is added between them.
```
rpspringuel: s/bottome/bottom/

davidweichiang: Done

(I realized that "topline" means something different than how I was using it, so I changed "topline" and "bottomline" to "top" and "bottom".)

davidweichiang:
Yes, the changed tests are all expected given the changes that are made to the meanings of `abovelinestextraise` and `abovelinestextheight`.

davidweichiang:
@rpspringuel @eschwab Is there anything else you'd like to see fixed in this PR? I will have a second PR that is waiting on this one; maybe I'll go ahead and open that one as a draft.

davidweichiang:
There's a paragraph in the new section that says that if you want to ensure space above the top note, you should set `spaceabovelines` to something greater than 0.41cm (the default is 0cm). Should the default `spaceabovelines` be increased to 0.41cm?

cf. https://groups.google.com/g/gregorio-users/c/_K5qfmGVunI/m/OwOCOC5cDgAJ

rpspringuel:
I'm satisfied with the documentation updates.

On the issue of defaults, I'm all for defaults that "make sense" or "seem natural", but understand that this is hard to define because different people have different opinions on what qualifies.  With all the changes that have been made recently to the spacing (including the PRs not yet merged) it may be time to reconsider our defaults, but I would make that a separate PR.

davidweichiang:
> I'm satisfied with the documentation updates.

Great, and the changes to the meanings of `abovelinestextraise` and `abovelinestextheight` as well?

> I would make [changing defaults] a separate PR.

That makes sense to me.

rpspringuel:
> Great, and the changes to the meanings of abovelinestextraise and abovelinestextheight as well?

I'm good with those redefined meanings.  You've managed to simplify things and explain them in a way that should make things much easier to understand.  The vertical spacings as they existed before this PR were always confusing to me (and plenty of others).

PR #1664, opened 2025-05-06
Speed up \gre@createdim

davidweichiang:
Previously, `\gre@createdim` would determine whether a dim should allow glue (i.e., whether it would be a skip or a dimen) by looking it up in a table, so in some sense it was creating something that already existed. The more practical problem was that this table lookup caused loading to take O(n^2) time in the number of dims. This PR makes the skip/dimen choice into a new argument to `\gre@createdim` and reduces the time to run tests by about 10%.

### DIFF
--- a/doc_check.sh
+++ b/doc_check.sh
@@ -71,7 +71,7 @@ sed -i.temp 's:.*gre@space@count@\([a-z@]*\).*:\1:' $CODEFILE
 
 #distances
 grep -h '\\gre@createdim{.*' gregoriotex-gsp-default.tex >> $CODEFILE
-sed -i.temp 's:\\gre@createdim{\([a-z@]*\)}.*:\1:' $CODEFILE
+sed -i.temp 's:\\gre@createdim{[a-z]*}{\([a-z@]*\)}.*:\1:' $CODEFILE
 sed -i.temp 's:.*gre@space@.*::' $CODEFILE
 
 #styles
--- a/tex/gregoriotex-spaces.tex
+++ b/tex/gregoriotex-spaces.tex
@@ -1448,7 +1448,7 @@
   \fi
 }
 
-% a macro for changing a dimension.  Unlike \grecreatedim, this function won’t create a new distance, just change an existing one.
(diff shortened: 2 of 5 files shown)


davidweichiang:
It looks like a lot of tests are failing (this is dug up from some time ago and I haven't tested in a while), so converting to draft.

davidweichiang:
All tests pass now (after https://github.com/gregorio-project/gregorio-test/pull/383).

doc_check.sh
```diff
@@ -71,7 +71,7 @@ sed -i.temp 's:.*gre@space@count@\([a-z@]*\).*:\1:' $CODEFILE
 
 #distances
 grep -h '\\gre@createdim{.*' gregoriotex-gsp-default.tex >> $CODEFILE
-sed -i.temp 's:\\gre@createdim{\([a-z@]*\)}.*:\1:' $CODEFILE
+sed -i.temp 's:\\gre@createdim{[a-z]*}{\([a-z@]*\)}.*:\1:' $CODEFILE
```
rpspringuel, line 74: Just to make sure I understand, your using `[a-z]*` to match the expected `skip` or `dimen` argument?

tex/gregoriotex-spaces.tex
```diff
@@ -1484,58 +1495,33 @@
 }
 
 % The common internals for \grecreatedim and \grechangedim.
```
rpspringuel: Not your fault (probably mine), but can you fix the typo in this comment and add the `@` that should be in `\gre@createdim`?

rpspringuel:
Assuming I understood the sed modification correctly, this looks good to me aside from the one typo I noticed that wasn't even in your changes, merely adjacent to them.  Fix the typo and I'll merge this.

tex/gregoriotex-spaces.tex
davidweichiang: Done

doc_check.sh
davidweichiang: That's right.

rpspringuel:

PR #1671, opened 2025-09-13
checkSyllabation.py: Handle more tags; handle words with notes only at the end

bhank:


rpspringuel:
I've never used this script, but a basic inspection of the changes seems to pass muster.  Can you provide a simple example and explain what should happen both before and after the changes so that I can do a more thorough check?

PR #1672, opened 2025-10-11
New syntax for separating gabc and nabc voices

### LINKED ISSUE #1667: Propose new delimiter to split NABC sections
Following https://github.com/gregorio-project/gregorio/issues/1662#issuecomment-2854678153, I would like to propose a change in NABC syntax, introducing a new delimiter to split multiple NABC sections, and reserving current delimiter `|` for alternating between GABC and NABC sections.

Current syntax example:
```
Text(gabc1|nabc11|nabc12|gabc2|nabc21|nabc22)
```

Proposed syntax example (using `&` as hypothetical delimiter for NABC sections):
```
Text(gabc1|nabc11&nabc12|gabc2|nabc21&nabc22)
```

This change would make it easier to implement syntax highlighting for GABC+NABC code.

davidweichiang:
Closes #1667.

- `&` advances to the next voice, but doesn't cycle back to the first voice.
- `|` is for cycling back to the first voice, but for backwards compatibility also can be used in place of `&`.

Currently, `nabc-lines:` is still required.

### DIFF
--- a/src/gabc/gabc-score-determination.l
+++ b/src/gabc/gabc-score-determination.l
@@ -383,13 +383,16 @@ semicolon. */
         BEGIN(notes);
         return OPENING_BRACKET;
     }
-<notes>(\$.|[^|\)])+ {
+<notes>(\$.|[^|&\)])+ {
         gabc_score_determination_lval.text =
                 gregorio_strdup(gabc_score_determination_text);
         return NOTES;
     }
 <notes>\| {
-        return NABC_CUT;
+        return NABC_MAJOR_CUT;
+    }
+<notes>& {
+        return NABC_MINOR_CUT;
     }
 <notes>\) {
         BEGIN(score);
--- a/src/gabc/gabc-score-determination.y
+++ b/src/gabc/gabc-score-determination.y
@@ -722,7 +722,7 @@ static char *concatenate(char *first, char *const second) {
 %token ALT_BEGIN ALT_END
 %token NLBA_B NLBA_E
 %token EUOUAE_B EUOUAE_E
-%token NABC_CUT NABC_LINES
+%token NABC_MAJOR_CUT NABC_MINOR_CUT NABC_LINES
 %token CLEAR
 %token PROTRUSION PROTRUSION_VALUE PROTRUSION_END PROTRUDING_PUNCTUATION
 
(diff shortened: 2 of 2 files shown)


davidweichiang:
This works as far as I know, but I just left it as a draft in case there was discussion about what the syntax should look like.

rpspringuel:
What would removing the `nabc-lines` header requirement involve?

davidweichiang:
My recollection is that the nabc voices are stored in an array, so the `nabc-lines:` is needed in order to malloc the array. Removing the `nabc-lines:` header would require adding code to realloc the array whenever a new voice is seen.

Tangentially, this is also related to a question that came up on the mailing list (https://groups.google.com/g/gregorio-users/c/IwslfmnTPDk); currently `\gabcsnippet` does not allow nabc because it doesn't offer a way to pass the `nabc-lines:` header.

rpspringuel:
So which would be easier, in your opinion, removing the `nabc-lines` header or creating a way to pass headers to `\gabcsnippet`?

davidweichiang:
I think they're both doable but the latter would be easier (and orthogonal to this PR).

I actually kind of like the `nabc-lines` header, since it's a global property of a score.

davidweichiang:
This for me went on the back burner when @lbssousa said that "this proposal is not needed anymore." If it is still needed, then let's iron out the syntax. I'm looking at this PR and do not understand why I chose this syntax. I would think it would be better, and still backwards compatible, to make `|` mean "if at last voice, cycle back to first voice, else advance to next voice" and make `&` mean "if at last voice, cycle back to first voice, else error."

lbssousa:
> This for me went on the back burner when @lbssousa said that "this proposal is not needed anymore."

In fact, I've said "this proposal was not needed anymore" for syntax highlighting purposes (since a [LSP server for Gregorio](https://github.com/AISCGre-BR/gregorio-lsp) could be smart enough to identify the GABC/NABC alternations by semantics), but it would still be useful for other purposes (getting rid of `nabc-lines` header[*], better treatment of error cases like the one pointed in #1726, etc.).

> I would think it would be better, and still backwards compatible, to make `|` mean "if at last voice, cycle back to first voice, else advance to next voice" and make `&` mean "if at last voice, cycle back to first voice, else error."

I agree (this syntax would resemble the one adopted in MusixTeX: `|` for intra-instrument staff alternating, and `&` for instrument alternating), and would suggest a deprecation warning about that "if at last voice, cycle back to first voice" for `|` delimiter, leading to a breaking change in Gregorio 7.0.0.

[*] One suggestion is implementing a "multi-stage parsing" of GABC files. In the first stage, the parser would calculate the maximum number of NABC voices within a snippet in the score (which would be easier to do with this new proposed syntax), and set `nabc-lines` to this value.

PR #1679, opened 2026-01-11
Eliminate \gre@luasavedim by using the "new" token library

davidweichiang:
In #1641 there was a hack called `\gre@luasavedim` for sending a dimen/skip from TeX to Lua code. These are defined as macros, and I thought, based on old information, that it was not possible for Lua code to access TeX macros. But it is possible, and this patch takes advantage of that to simplify the vertical spacing code. No behavior should be changed, except that this now requires (I believe) LuaTeX 0.85 or later, which was released in late 2015. Is that a safe assumption to make?

If so, then there's some other modernization that can be done, so this is a draft PR for now.

### DIFF
--- a/doc/Command_Index_internal.tex
+++ b/doc/Command_Index_internal.tex
@@ -1230,9 +1230,6 @@ Alias for \verb=\resizebox=.
 \macroname{\textbackslash gre@dimension}{}{gregoriotex-spaces.tex}
 Workhorse function for setting distances in \verb=\grecreatedim= and \verb=\grechangedim=.
 
-\macroname{\textbackslash gre@luasavedim}{}{gregoriotex-spaces.tex}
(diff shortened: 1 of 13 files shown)


davidweichiang:
Since LuaTeX 1.0 (Sep 2016) is a round number, I went ahead and made that the minimum version. All tests pass, and the code is a bit simpler.

rpspringuel:
I've reviewed the code and everything looks good at a glance.  I just need to check the tests.  As best I can tell, this shouldn't change any tests, but I'm in the middle of checking the tests on another PR so this one will need to wait its turn.  If all the tests pass as expected, I'll merge this.

davidweichiang:
Thanks!

rpspringuel:

Issue #1452, opened 2019-04-04
Morae after clivis not aligned vertically


chrylis:
I have run into a bizarre issue where a certain input causes the duplex morae to be rendered out of alignment vertically. I tried to reduce the file as much as possible, but this is the limit of what I've been able to figure out. Commenting the last line eliminates the problem. Strangest of all, the amount of the horizontal offset varies from clivis to clivis.

I'm running 5.0.1 (Gentoo still only has TeX Live 2017), and the misalignment doesn't require any interesting TeX.

[misalign.gabc.gz](https://github.com/gregorio-project/gregorio/files/3042323/misalign.gabc.gz)

![misaligned-morae](https://user-images.githubusercontent.com/2394512/55539431-3c5f5d80-5686-11e9-84dc-41ae718dc9c4.png)

rpspringuel:
Confirmed on TeX Live 2018 running latest develop branch.  

The higher punctum mora is printed second and it appears to always be shifted right from the lower one.  The lower punctum mora is supposed to be printed with zero width, but there must be something in the kerning calculations that doesn't quite return us to where we started so that the higher one is printed directly above it.

davidweichiang:
@rpspringuel This can be closed, I think

PR #1682, opened 2026-01-19
Change spacebeforesigns from a skip to a dimen (#1656)

### LINKED ISSUE #1656: Double punctum mora (..) in justified lines
`g..` shows up like this:

Image

davidweichiang:
This does change a number of tests. Most of the changes are miniscule; the biggest one was `tex-output/multi-line-initials/multi-line-initials.tex`.

Closes #1656.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Previously, if a score ended with `Z` (ragged line break) or `z` (justified line break), the appearance of the last line would sometimes depend on `Z` versus `z` and sometimes depend on `\gresetlastline{ragged}` versus `\gresetlastline{justified}`. Now, the appearance of the last line always depends on `\gresetlastline`.
 - The meanings of the distances `spaceabovelines` and `abovelinestextheight` were changed to be (hopefully) easier to use and closer to their descriptions in the documentation.
 - All lengths related to vertical spacing are documented in greater detail in a dedicated section in GregorioRef.pdf.
+- `spacebeforesigns` was changed from a skip to a dimen; that is, it can no longer have a stretch or shrink (`plus` or `minus`).
 
 ### Deprecated
 - The count `grefinalpenalty` no longer has any effect and will be removed in a future release.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -18,6 +18,10 @@ Forced line breaks (`Z` or `z`) at the very end of a score are discouraged. Plea
 
 The meanings of these distances were changed to be (hopefully) easier to use and closer to their descriptions in the documentation. Please see the section "Vertical spacing" in GregorioRef.pdf for more information. Scores that use the default settings and have above-lines text (`<alt>`) will now have slightly more whitespace above the above-lines text. Scores that set either of these two distances explicitly may experience larger changes in whitespace between staves, and should be checked.
 
+### `spacebeforesigns`
+
+This space (the space before a punctum mora) was changed from a skip to a dimen; that is, it can no longer have stretch or shrink (`plus` or `minus`). For example, a command `\grechangedim{spacebeforesigns}{0.050 cm plus 0.004 cm minus 0.004 cm}{scalable}` should be changed to `\grechangedim{spacebeforesigns}{0.050 cm}{scalable}`.
+
 ## 6.1
 
 ### Multiline initials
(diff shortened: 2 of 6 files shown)


rpspringuel:
Definitely a bunch of tests changing, but you've only noted the new one in the test PR.  Can you go through the tests which are changed and make sure the changes, as minuscule as they are, are what you'd expect?  If so, add those changes to the test PR.

davidweichiang:
### Large differences
tex-output/multi-line-initials/multi-line-initials-2.tex

image
image
image

That's a big difference, but I can't really say which is better.

**tex-output/gsptest/gsptest.tex: Something bad happened here, but I will have to investigate later.**

### Small differences

tex-output/multi-line-initials/multi-line-initials.tex: I think the new one is better, with the dots less spaced out
tex-output/bugs/fix-1139/fix-1139.tex: Same

### Almost imperceptible differences

tex-output/PopulusSion/PopulusSion.tex
plain-tex/PopulusSion/PopulusSion.tex
plain-tex/luatex-ja/PopulusSion.tex
gabc-output/glyphs/punctum-inclinatum-2.gabc
gabc-output/veni.gabc
gabc-output/FactusEst.gabc
gabc-output/bugs/fix-928.gabc
tex-output/lineshifts/lineshifts.tex
tex-output/euouae/euouae.tex
tex-output/bugs/fix-595/fix-595.tex
tex-output/bugs/fix-845/fix-845.tex
tex-output/bugs/fix-396/fix-396.tex
tex-output/header-hooks/header-hooks.tex
tex-output/hideabovelinestext/hideabovelinestext.tex
tex-output/hidelyricsnotes/hidelyricsnotes.tex
tex-output/per-line-spaces/per-line-spaces.tex
tex-output/last-line/last-line.tex
tex-output/hidetranslation/hidetranslation.tex

rpspringuel:
tex-output/multi-line-initials/multi-line-initials-2.tex:
This looks to me like a justification switch.  The tiny bit less glue that comes with this change means that the line no longer has enough glue to be justified and thus switches to a more compressed unjustified setting.  I'd say accept this change.

tex-output/gsptest/gsptest.tex:
This test is for testing the loading of custom spacing configurations from a file, so check that the gsp being loaded by this test has spacebeforesigns defined as a dimension, not a skip.

For the others, it sounds like they can be accepted as is, but I'll double check.

davidweichiang:
> This test is for testing the loading of custom spacing configurations from a file, so check that the gsp being loaded by this test has spacebeforesigns defined as a dimension, not a skip.

That's exactly what happened, but I'm worried about the fact that it did not lead to an error; instead, the "plus" and "minus" specifications got printed into the score!

rpspringuel:
That doesn't surprise me as we LaTeX automatically detects the end of the dimension and reads in the glue as text under normal circumstances, so the text just gets inserted into the document each time the macro storing the value is called.  Since learning about the [ifis](https://mirrors.mit.edu/CTAN/macros/plain/contrib/ifis-macros/ifis-macros-doc.pdf) package, I've thought we should introduce a check into the macros that define our distances to verify that they are receiving valid input, but I've never had the time to implement it.

davidweichiang:
There was a stub of a check in `\gre@dimension` that maybe used to be a full check but was removed (hopefully not by me); it can be easily restored.

rpspringuel:
Checked all the small changes.  They look like the kind that I've come to expect when there are minor tweaks to the fonts and can be approved in the test PR.

davidweichiang:
OK, I've accepted all the small changes and edited away the big one.

rpspringuel:
I'm still getting two test failures:

gabc-output/glyphs/shapes.gabc : FAIL - [gabc-output/glyphs/shapes.gabc.out/page-0.png] differ from expected
tex-output/gsptest/gsptest.tex : FAIL - [tex-output/gsptest/gsptest.tex.out/page-0.png] differ from expected

However, in both cases they are inconsequential changes.  What's your dpi setting for the test?  Maybe I've got my setting a bit higher than you so I see something small that you don't?

davidweichiang:
I’m guessing that I forgot to accept gsptest, but I’m not sure what happened to shapes. I’ll check it after the weekend.

rpspringuel:
I think this will fix https://github.com/gregorio-project/gregorio/issues/1452 in addition to https://github.com/gregorio-project/gregorio/issues/1656

davidweichiang:
At the default dpi of 300, shapes.gabc passes.

davidweichiang:
But I just updated the result for gsptest.tex.

davidweichiang:
Even at 1200dpi, shapes.gabc passes for me.

PR #1687, opened 2026-01-28
Yearly Maintenance plus

rpspringuel:
New year, new copyright dates.

Also remove a file from the update list that was removed in #1640.


eschwab:
Year is correct.

PR #1691, opened 2026-02-02
Missing CHANGELOG and UPGRADE entries

### LINKED ISSUE #1688: Missing CHANGELOG entries
Seems I got a bit over eager on merges and failed to note PRs that were missing CHANGELOG entries:

#1681 
#1679 
#1666 
#1664 
#1662 
#1653

davidweichiang:
Closes #1688.


davidweichiang:
Oh, sorry, I just saw that this duplicates #1689 and #1690. Mostly I think your entries are better. I think a brief explanation of how to get NABC below the lines could be helpful (or maybe it needs to be spelled out better in the manual?). And the required LuaTeX version should be mentioned in UPGRADE as well, right?

rpspringuel:
The note in UPGRADE about the new minimum version is a good idea.  I'll add it to #1698.

The new NABC line isn't something that goes in UPGRADE, as it doesn't require the user to make changes to compile old documents.  If you think the documentation on how to use it could be improved, that should go in GregorioNABCRef.tex

PR #1692, opened 2026-02-02
Correct recording of files read and written in Lua

### LINKED ISSUE #1685: feature request to allow tmp-gre directories while running latexmk
As I posted on the Gregorio mailng list, something changed when I updated to TeXLive 2025 with MacTeX (I am on a 2018 Macbook Pro running Sequoia 15.7.1).

I run latexmk to automatically typeset a second etc. time. It usually works correctly including for an initial typesetting run, because the spacing, braces, and other featuress depend on this second pass, and it's sometimes necessary for indices, cross-references and so on.

The auxiliary files generated by GregorioTeX are now placed in a directory called tmp-gre; in that directory is a directory called scores (since I place my gabc files in such a directory and then call `\gregorioscore{scores/score.gabc}`

However, because of the way that GregorioTeX does things, a change to a .gabc file (any change: orthographic, notes, a line break…) is not recognized. latexmk will not run, and since I run things from TeXShop which has a shortcut for typesetting, I can't force typesetting, which I could do from the command line; this, however, would break SyncTeX (at least for the TeXShop implementation of it: that is, either the PDF preview in TeXShop becomes black or the PDF appears, but without the ability to click into the PDF or source respectively).

So to get it to work again, I have to delete the auxiliary files, which would be problematic in the case of a document with labels and cross-references, especially if it's larger.

I have attached a minimum working example for someone to run; all that you would need to do to test the problem is make some change to the .gabc file.

John Collins, the maintainer of latexmk, received a message from Herb Schulz (who essentially implements latexmk for the Mac and specifically for TeXShop) sent on my behalf. John wrote back:

[gregorio&latexmk.pdf](https://github.com/user-attachments/files/24863475/gregorio.latexmk.pdf)

[Gregorio with latexmk.zip](https://github.com/user-attachments/files/24863482/Gregorio.with.latexmk.zip)

>  Yes, the problem is that the .gabc is being processed by a lua script, so it doesn't get listed in either of the .fls or .log
> files as an input file. The script does in fact put a relevant message in the .log file, but in a package-specific form that latexmk doesn't recognize. Ideally the gregoriotex package should be modified appropriately. (The author(s) do know about latexmk.) Here are two possibilities, and have attached corresponding files:
>  1. Override the package's \gregorioscore macro to put a
>  suitable message in the .log file. (That's a trick I've
>  recommended in other similar situations.) This is the simplest
>  immediate fix.
>  
>  2. Modify the lua script to write the message. This is more
>  general, because it works after the script has done a search
>  for the file, and takes account of the script's treatment of an
>  extension for the file.
> 
> It would also be possible to configure latexmk to recognize the package's message, and I've done that in other similar situations. But I think the above solutions are simpler. The first attached file is test.tex, which implements the first solution. It's just the MWE you sent with some extra lines. The second file is gregoriotex.lua. It's a direct replacement for the file of the same name in the gregoriotex package, just modified to write an appropriate message. I would suggest putting it in the same directory as the document. Then the modified lua script gets used instead of the distributed version, but the distributed version is intact. I suggest Matthew puts in a feature request for the package. He's welcome to quote my implementation.

I have attached the zip file sent from John. I haven't had a chance to test his local solutions yet. I have no idea what is happening with latexmk for other users; I highly recommend latexmk and ordinarily would say that it should be a part of the introductory document in the documentation on the project website (because it took me way too long to realize that certain things like psalm braces require a second pass), but this is a big problem.
[gabc_mwe.zip](https://github.com/user-attachments/files/24863368/gabc_mwe.zip)

davidweichiang:
Correct recording of files read and written in Lua, so that latexmk recognizes dependencies correctly. 

Closes #1685.


rpspringuel:
Can you think of a way that we could test this bug in the test repository?  I'd like to avoid it creeping back in, especially with the plans to really rework the backend to be pure Lua.

davidweichiang:
I've just added a test that hopefully does the job.

PR #1696, opened 2026-02-06
Make bool.h compatible with C23 (used by default by GCC 15)

### LINKED ISSUE #1650: Compiling using gcc 15 fails
Compiling the project using gcc-15 fails. The following code in `gregorio/gregorio-src/src/bool.h` is not accepted any more. The attached patch solves the issue or me, but makes bool.h basically obsolete...

[gregorio_bool.h.txt](https://github.com/user-attachments/files/19609738/gregorio_bool.h.txt)

davidweichiang:
Passes all tests with -std=gnu89, -std=c99, and -std=c23. (With the latter two, the option -Wno-gnu-statement-expression also has to be removed.)

I did not check whether the change in sizeof(bool) would be a problem.

Credit: @jakubjelinek
Closes #1650


rpspringuel:
Looks fine to me, though I've never seen the cited problem myself so I might still be missing something.

rpspringuel:
Oh, can you add a CHANGELOG entry?

davidweichiang:
Sure, done.

PR #1707, opened 2026-02-20
Add commands to documentation for macros with long arg tables

### LINKED ISSUE #1706: Documentation blow-up
I was starting the process of testing the current development branch against the pretest and ran into a problem with the documentation blowing up.  The page count goes to absurdly high numbers and the compilation eventually fails with a TeX capacity exceeded error.

Thinking it might be a pretest related thing, I switched back to TeX Live 2025 and made sure I'd downloaded and installed all the most recent ctan updates; the problem remained.

At this point I need to start sorting through all the most recent updates but I'm out of time to work on this today.  I'm testing #1683 by bypassing the documentation build, but this will need to be figured out before we can submit to the Pretest.  *If you can look at this, please do so as soon as possible.*

rpspringuel:
`tabularx` does not allow it's table contents to be broken over multiple pages.  Previously having a table bigger than could normally fit on one page just led to an overfull page, but recent changes have caused such a table create a cascade that leads to TeX capacity exceeding errors. Using the `xltabular` package, we define a new version of `argtable`, `longargtable`, which can be split over multiple pages. To prevent issues with the page breaking controls that are built into `\macroname` interfering with the headers and footers of this `longargtable`, we also define `\longmacroname` which forces a pagebreak directly before that macro in the docs.  Paired with `longargtable` this forces the macro with the long argument description to start a new page and use the pagebreaks allowed by the new environment.

Fixes https://github.com/gregorio-project/gregorio/issues/1706

### DIFF
--- a/doc/Command_Index_internal.tex
+++ b/doc/Command_Index_internal.tex
@@ -1082,11 +1082,11 @@ Macro for defining the various types of horizontal episema porrectus.
   \#2 & string & Camel case name of horizontal episema porrectus to be defined\\
 \end{argtable}
 
-\macroname{\textbackslash gre@get@spaceskip}{\#1}{gregoriotex-signs.tex}
+\longmacroname{\textbackslash gre@get@spaceskip}{\#1}{gregoriotex-signs.tex}
 Loads \verb=\gre@skip@temp@four= with the appropriate rubber length given the
 desired case.
 
-\begin{argtable}
+\begin{longargtable}
   \#1 & \texttt{0} & Default space.\\
   & \texttt{1} & Zero-width space.\\
   & \texttt{2} & Space between flat or natural and a note.\\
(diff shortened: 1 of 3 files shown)


eschwab:
Looks good to me. Do we want a comment above the `\newcommand`'s for `{,long}macroname` `{,long}argtable` to give directions for their proper use?

OR: this may be better in a separate issue, perhaps a bit of expansion of `doc_README.md` is in order to better explain these differences and the distinctions between macros found in `Command_Index_{User,internal,gregorio}.tex`

rpspringuel:
@eschwab I've made a couple of attempts to do both of the things you asked about.  Not sure if this is the best way to do it, but perhaps it's good enough for now as I'd like to get a beta into the pretest this weekend, if possible.

eschwab:
That is basically what I had in mind since I did not think we had that type of explanation anywhere.

PR #1713, opened 2026-02-23
Feature/nabc skip accidentals

lbssousa:
Fixes #1712.

**NOTE**: there should be a small conflict when merging `ctan` branch into `develop` after this PR, due to lack of lack of `\gre@nabcvoice@{i,ii}@phantomwrapper` in `ctan`. They were introduced in PR #1683 directly in `develop`. Resolving this conflict is just a matter of replacing `\GreSetNabc{Above,Below}Lines` with `\csname GreSetNabc{Above,Below}Lines\ifgre@nabc@voice@i@skipaccidentals Deferred\fi\endcsname`

PR #1711, opened 2026-02-23
Fix regressions introduced after ctan merge (#1708)

lbssousa:
This PR fixes a regression introduced after resolving the conflicts found on merging the `ctan` branch into `develop` (see issue #1708), making the test `gabc-output/bugs/fix-1700` fail with a fatal LaTeX error.

# Root cause of the regressions

The `ctan` branch refactored the internal NABC rendering pipeline in `gregoriotex-main.tex`, replacing the single-phase pair `\gre@currentnabcvoice@{i,ii}` with a two-phase architecture:

| Phase | Voice i | Voice ii |
|-------|---------|----------|
| Pre-evaluate (phase 1) | `\gre@nabc@pre@voice@i` | `\gre@nabc@pre@voice@ii` |
| Place/emit (phase 2) | `\gre@nabc@emit@voice@i` | `\gre@nabc@emit@voice@ii` |

However, recent features implemented directly in `develop` branch still make use of some pieces defined in old `\gre@currentnabcvoice@{i,ii}` context.

When the merge was performed, some important parts of code didn't have their `\gre@currentnabcvoice@{i,ii}` dependency fixed (e.g. `\GreFlushNABC`), or weren't refactored as needed (e.g. `\gre@nabc@prepare`).

# Solution

- Fix `\GreFlushNABC` implementation, replacing the old `\gre@currentnabcvoice@{i,ii}` macros with new two-phase ones.
- Restore `\gre@dimen@nabc@maxwidth` within `\gre@nabc@prepare`, needed for NABC anti-overlapping mechanism. 
- Remove stray `part_attr=7` from `\gre@nabc@prepare` to fix invisible NABC voice 1 spacing 

Fixes #1708.

Supersedes #1710.

PR #1714, opened 2026-02-24
New feature: \gresetnabcskipalterations

### LINKED ISSUE #1712: [Feature Request] Skip accidentals when aligning NABC snippets
Currently, NABC glyphs are aligned to the leftmost element of a GABC note group, including accidental marks (`(gx)`, `(gy)`, `(g#)` and parenthesized variants), which may lead to undesired results.

I would like to propose a new option `\gresetnabcskipaccidentals`, which may be configured independently for each NABC voice, to make NABC glyphs skip those accidental marks and anchor to the first GABC note of the group. This option should be `false` by default, in order to maintain compatibility with previous versions of Gregorio.

lbssousa:
Fixes #1712.
Supersedes #1713.
Depends on #1711.


lbssousa:
Now that PR #1711 is merged, I need to add a fix for a particular case when we have both `\gresetnabcalignment{neume}` and `\gresetnabcskipalterations{true}` set, and the score has both note alterations and NABC glyph with significant letters at left.

rpspringuel:
Since you merged develop in here in the main repository, you need to do the same on the test repository side to keep the things consistent.

PR #1716, opened 2026-03-01
Fix NABC inter-syllable overflow spacing and add nabcintersyllablemingap

### LINKED ISSUE #1715: NABC extrakern causes excessive inter-syllable spacing when natural gap is sufficient
Updated after https://github.com/gregorio-project/gregorio/issues/1715#issuecomment-3976224837

# Description

The NABC extrakern mechanism, introduced in commit e644a34d to prevent overlap between adjacent syllables' NABC neumes, unconditionally adds horizontal space at the end of a syllable whenever its NABC content is wider than the underlying notes. However, it does not account for the natural inter-syllable spacing that GregorioTeX already inserts between syllables (syllable-final skip, text alignment adjustments, etc.).

In many common cases, the natural gap between the right edge of one syllable's notes and the left edge of the next syllable's notes is already large enough to accommodate the NABC overshoot without any overlap. In these situations, the extrakern adds space that is entirely unnecessary, resulting in visibly excessive gaps between syllables.

# Examples

## Current behaviour

Image

## Proposed new behaviour

Image

# Root cause

The extrakern is computed as `maxwidth_NABC − notes_width`, which is the amount by which the widest NABC neume extends past the right edge of the notes. This kern is then appended to the syllable's notes box, unconditionally widening it.

The problem is that this calculation ignores all other sources of horizontal separation between syllables:

1. **Syllable-final skip**: the base inter-syllable spacing that GregorioTeX inserts between every pair of adjacent syllables.
2. **End difference**: when the syllable text extends past the notes' right edge, that extra text width already provides natural separation.
3. **Next-begin difference**: when the next syllable's text extends before its notes' left edge, that also contributes to the effective gap.

Because the extrakern does not subtract these existing contributions, it adds the full NABC overshoot as extra space even when the natural gap already exceeds it. The result is a gap that is roughly double what would be needed — the natural spacing plus the redundant extrakern.

lbssousa:
Fixes #1715.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Fixed a few bugs related to horizontal spacing around bars and clef changes. See issues [#1191](https://github.com/gregorio-project/gregorio/issues/1191), case 3 of [#1724](https://github.com/gregorio-project/gregorio/issues/1724), [PR #1743](https://github.com/gregorio-project/gregorio/pull/1743), and [#1745](https://github.com/gregorio-project/gregorio/issues/1745).
 - Fixed a bug in horizontal spacing when the first syllable consists of only a bar. See [PR #1741](https://github.com/gregorio-project/gregorio/pull/1741).
 - Fixed a bug in horizontal spacing before a soft alteration. See [PR #1761](https://github.com/gregorio-project/gregorio/pull/1761).
+- Fixed excessive inter-syllable spacing when NABC neumes extend past the notes of their syllable: extra space is now added only when the neumes of consecutive syllables would actually come closer than `nabcintersyllablemingap`.  See [#1715](https://github.com/gregorio-project/gregorio/issues/1715).
 
 ### Added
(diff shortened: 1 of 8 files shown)


davidweichiang:
In #1720 I'm working on moving everything related to syllablefinalskip into Lua, while this PR adds a dependency on syllablefinalskip. So merging this PR first would create some extra work needed on #1720.

On the other hand, if #1720 were merged first, then managing the space between NABC syllables would become much easier. It would just involve adding a pass to the Lua `pre_linebreak` filter that looks for consecutive nodes with part=7, and if the distance between them is less than 0, increase the `syllablefinalskip` node by that amount. Similarly for part=8.

Maybe #1720 can reach a point where it's mergable in a few days. @lbssousa, what do you think about trying this in Lua instead?

lbssousa:
Sure, @davidweichiang . I'll try it when your work is ready. We can hold this PR for now.

davidweichiang:
@lbssousa Now that #1720 has been merged, I think it will be a lot cleaner to implement this.

lbssousa:
@davidweichiang Just refactored here PR after #1720 is merged, but I also depend on PR #1764 do be merged in order to complete this one.

davidweichiang:
What do you need from #1764? Maybe we can cherry-pick those into a separate smaller PR.

lbssousa:
Essentially, I need the implementation around `bar_syllable_spacing` / `old_bar_syllable_spacing` functions. The `end_shift` constraint set in `bar_syllable_spacing` (and `old_bar_syllable_spacing`) can be extended by this PR with one term per NABC voice.

Issue #1673, opened 2025-11-11
Moving C/TeX code into Lua?


davidweichiang:
This might be a crazy idea, but it seems to me that a great deal of complexity in managing this project stems from the fact that it is written in three languages:

- The C code is long, difficult to maintain, can only be pushed to TeXLive twice a year, and is subject to additional security restrictions.
- The TeX code is slow, difficult to maintain, and has severe limitations like not being able to look ahead.

But Lua doesn't have any of these problems, as far as I can tell. I can imagine that by moving all of the C and most of the TeX into Lua, the result could be shorter, simpler, and faster.

Rewriting the C into Lua:
- It depends on lex/yacc for parsing, but it's not hard to write a PEG parser in Lua. (I've got a prototype of this working.) 
- There would be a small speed penalty, which _might_ be offset by not having to call an external program.
- The data structures would become far simpler, maybe even from the user point of view. (For example, in #1667, it was suggested to eliminate the `nabc-lines:` header field. One of the reasons this option exists is that the C code needs to allocate a fixed-size array to store the lines.)

Rewriting TeX into Lua:
- There's a lot of repetition in the .gtex file to get around the fact that TeX cannot look ahead. Moving this logic into Lua would cut down on this complexity.
- Any processing moved from TeX to Lua should go much faster.
- In the most extreme case, Lua code could directly add glyphs and boxes without going through TeX code at all (https://wiki.luatex.org/index.php/TeX_without_TeX).

rpspringuel:
I'm not adverse to the idea in principle, but I'm also not in a position to contribute to it either.  I simply have too much on my plate at the moment.

This leads me to worry about the scope of the changes.  I don't see an easy way to step through this kind of change piece by piece, which means any code review is going to be looking at a whole lot of changes.  Now, with our extensive test suite, we should be able to catch anything that deviates too wildly, but for anyone who is not involved in making the changes (i.e. me) wrapping their head around the diff is going to be a challenge.  Is there some way that you can think of that would reasonably break this kind of rewrite down into steps?  It would be ideal if those steps resulted in working code milestones, but even if that's not possible, even being able to stagger the review into more manageable pieces would help a set of outside eyes follow the work and not simply take it on faith because the tests work.

davidweichiang:
Thanks for entertaining a crazy idea. I tried writing the bare minimum to generate the image below:

[...]

PR #1720, opened 2026-03-01
Syllable spacing in Lua

### LINKED ISSUE #1719: Syllable rewriting bugs
Syllable rewriting works by moving the last part of a syllable to the first part of the next syllable. This happens as long as there's no dash after the current syllable.

1) If `maximumspacewithoutdash` > 0, then there may be a space between the two syllables, and the last part "jumps" to the next syllable. Here, the "nc" jumps to the next syllable, which is probably not desired:

Image

2) If there is a ligature between the middle and last part of the syllable, then rewriting breaks the ligature. Here, the font has an "an" ligature which is broken:

Image

### LINKED ISSUE #1725: <clear> undoes wrong syllablefinalskip
In `\gre@clearsyllable`, there is a line that undoes `syllablefinalskip`, but it's the syllablefinalskip after the _second_ (cleared) syllable when (I think) it should be the first.

This can make the cleared syllable appear too far to the left or right. In the example below, "b" is the cleared syllable:

```
\gresetinitiallines{0}
\gabcsnippet{a(gg)<clear>b(g) c(g)}
\verb|\grechangedim{interwordspacenotes}{1in}{scalable}|
\grechangedim{interwordspacenotes}{1in}{scalable}
\gabcsnippet{a(gg)<clear>b(g) c(g)}
```

Image

### LINKED ISSUE #1756: Mid-score \gresetsyllablerewriting{off}  loses characters
There should be three `f`s, but the middle one gets dropped:

```
\gabcsnippet{iiiif(g)iiiif(g[ev:\gresetsyllablerewriting{off}])iiiif(g)}
```

Image

This is because the `f` gets saved for the next syllable, but in the meantime rewriting is turned off, so the `f` is not printed in the next syllable.

davidweichiang:
This is aiming towards #1673, but pursuing a different strategy than what was proposed there.

The first commit (https://github.com/gregorio-project/gregorio/pull/1720/commits/02ec95278623f615b5effda672bb2704bc8061aa) moves syllable rewriting into Lua, which has the following benefits:

- The old algorithm worked by moving the last part of a syllable to the first part of the next syllable. It could therefore break ligatures between the middle and last part of a syllable (#1719). The new algorithm works by joining the text boxes together; it does not break ligatures.
- If maximumspacewithoutdash > 0, the old algorithm could make the last part of a syllable "jump" across the space between syllables (#1719 again); the new one only rewrites syllables if they are actually touching.

ETA: These two cases are demonstrated in https://github.com/gregorio-project/gregorio-test/pull/420.

And there are still the following problems:

- [x] The old algorithm computed syllable spacing after rewriting; the new algorithm computes them in the opposite order. This means that some syllable spacing will be slightly too big or small.
- [ ] When the width of a syllable changes due to rewriting, the new syllable is *left*-aligned with the old syllable, but it might make more sense to center it with the old syllable. 

The first problem can be seen in the test
tex-output/syllable-rewriting/syllable-rewriting.tex

ETA: The existing code was already making an approximation, in that it was aligning the first note and vowel without taking ligaturing and kerning into account. The new code is just extending this approximation to possibly a whole word. In my opinion, the approximation is visible only in the test `syllable-rewriting.tex` (because of the unusual font) and causes only a miniscule difference elsewhere.

Further tests that have miniscule differences:
gabc-output/glyphs/torculus_resupinus_quilisma.gabc
gabc-output/glyphs/torculus_resupinus.gabc
gabc-output/slurs.gabc
gabc-output/FactusEst.gabc
tex-output/dominican-flats/dominican-flats.tex
tex-output/SalveReginaOP/SalveReginaOP.tex
tex-output/bugs/fix-595/fix-595.tex
tex-output/bugs/fix-1110/fix-1110.tex
tex-output/lyric-centering/lyric-centering.tex

Tests that have big changes:

- [x] tex-output/multi-line-initials/multi-line-initials-2.tex: fixed in #1743 
- [x] tex-output/bugs/fix-984/fix-984.tex: this is some complicated interaction with eolshift and doesn't seem better or worse than before

Closes #1719.


davidweichiang:
Commit https://github.com/gregorio-project/gregorio/pull/1720/commits/4fa508d2a27c519f07bc00ff0f5a756de970e140 starts to move the calculation of `syllablefinalskip` into Lua. The TeX code computes the minimum distance between text and notes (between `\GreSyllable`s only, not `\GreBarSyllable`s), and emits a zero-width skip instead of syllablefinalskip. The Lua code computes the syllablefinalskip and edits the skip to have the right width.

Because it uses the actual `begindifference` of the next syllable, not `nextbegindifference`, and there are some existing bugs in how `nextbegindifference` is calculated, there are quite a few new broken tests.

Because of #1723:
gabc-output/glyphs/accidentals.gabc (case 1)
gabc-output/glyphs/brackets.gabc (case 4)
gabc-output/bugs/fix-1700.gabc (case 2)
gabc-output/soft-alterations.gabc (case 3)
tex-output/nabc-alignment/nabc-alignment-two-voices.tex (case 5)
tex-output/nabc-alignment/nabc-alignment.tex (case 5)
gabc-output/lines/2-lines.gabc (case 6)
gabc-output/lines/3-lines.gabc (case 6)
gabc-output/lines/5-lines.gabc (case 6)

Because of #1724:
gabc-output/glyphs/clef_change.gabc
gabc-output/glyphs/only-special.gabc
~~gabc-output/bugs/fix-1285.gabc~~
gabc-output/edge_cases.gabc

Only miniscule changes:
gabc-output/glyphs/punctum-inclinatum-2.gabc

davidweichiang:
Commit https://github.com/gregorio-project/gregorio/pull/1720/commits/ec4e22c9bb2d3b90b96b656f90c2b254eb283a10 moves syllable clearing into Lua.

The amount to shift a syllable in order to clear it is now measured exactly, whereas previously it depended on predicting the begindifference of the next syllable, as well as issue #1725. So cleared syllables are tighter than before.

Breaks two new tests:
gabc-output/glyphs/clear.gabc: because of the changes to clearing described above
tex-output/bar-spacing-old/bar-spacing-old.tex: because the actual begindifference of a bar is used, whereas before it was 0

davidweichiang:
Commit https://github.com/gregorio-project/gregorio/pull/1720/commits/a48c861348f42654c17ce27d0b0a6cbca281282d moves hyphenation into Lua. This was the last thing that depended on syllablefinalskip, so syllablefinalskip can be deleted from the TeX code, and perhaps it would make sense to merge (so that @lbssousa can work on #1715), after 6.2.0 of course.

Some very minor benefits:
- A couple of cases in `\gre@calculate@eolshift` appeared to be incorrect or missing; these are now fixed.
- End-of-line hyphens used to not be kerned, but now they are.
- When lyrics are hidden, the old code inserted extra stretch after hyphenated syllables. The new code does not. This affects one test:
  - tex-output/hidelyricsnotes/hidelyricsnotes.tex

Several more tests are broken.

If a hyphenated word comes before a bar, the space between them is now incorrect until bar spacing is moved into Lua. This happens in the following tests:

- [ ] gabc-output/glyphs/custos_linebreak.gabc
- [x] tex-output/bar-spacing/bar-spacing-new.tex
- tex-output/bar-spacing/lyric-centering.tex: only the lines printed for debugging are affected

Miniscule differences:

- ~~gabc-output/two-nabc.gabc~~
- ~~gabc-output/bugs/fix-1424.gabc~~
- ~~gabc-output/bugs/fix-804.gabc~~
- ~~gabc-output/bugs/fix-347.gabc~~
- ~~tex-output/illumination/illumination.tex~~
- tex-output/bugs/fix-1110/fix-1110.tex
- tex-output/bugs/fix-1319/fix-1319.tex
- ~~tex-output/snippet/snippet.tex~~
- ~~tex-output/hidelyricsnotes/text-only.tex~~
                                                                                             
More than miniscule but not worrisome in my opinion: 
- tex-output/visibility-controls/visibility-controls-main.tex

davidweichiang:
This code defers some computations to the pre-linebreak filter, which can depend on user settings. If the user changes settings in the middle of the score (using `[ev:...]` or the like) then the change will affect the entire score, which may be surprising. In the tests, this only happens in `bar-mora.gabc`, which is not affected here. But in principle, are there settings that need to be settable in the middle of the score?

@rpspringuel

doc/Command_Index_internal.tex
rpspringuel: The table below seems to indicate that you're changing the number of arguments.  If that's true it should be reflected up here too.

rpspringuel: The table below seems to indicate that you're changing the number of arguments.  If that's true it should be reflected up here too.

rpspringuel:
I'd definitely like to see some documentation on the new Lua code before merging.  We've been very poor about documenting our Lua code thus far and I'd like to start changing that.

I'm thinking that our policy for new Lua code should be as follows:

Every new function (or non-trivial change to an existing function) gets a documentation comment in EmmyLua-style:
```
--- Calculates the area of a rectangle
---@param width number The width of the rectangle
---@param height number The height of the rectangle
---@return number area The computed area
local function area(width, height)
    return width * height
end
```

Do you think that's reasonable?  The goal is to improve our Lua documentation over time without there being an overwhelming amount of work. (Though if someone gets ambitious and wants to document all current Lua functions, I won't stop them.  That's how I really got into the project: creating the TeX documentation.)

eschwab:
The code documentation suggestion seems good. I started a discussion on the gregorio-devel mailing list in an attempt to standardize our lua style.

davidweichiang:
The last commit takes a first stab at adding the Lua comments.

davidweichiang:
To fix:
- [x] Position hyphen correctly in a rewritten syllable (visible in test fix-396.tex)
- [x] Improve positioning of clef changes without bars

tex/gregoriotex-main.tex
```diff
@@ -249,6 +263,18 @@
         \hfill
       \fi
       \gre@penalty{-10001}%
+      % Above, we emitted kerns of -enddifference and -eolshift early,
+      % and they will be emitted again later in \GreSyllable. The
+      % second set of kerns has no effect because they are at the
```
rpspringuel: If the second set of kerns have no effect because they are at the beginning of a line, can we just not emit them rather than emitting their opposite here to cancel them out?

tex/gregoriotex-syllable.lua
```diff
@@ -0,0 +1,494 @@
...
+--but WITHOUT ANY WARRANTY; without even the implied warranty of
+--MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--GNU General Public License for more details.
+--
+--You should have received a copy of the GNU General Public License
+--along with Gregorio.  If not, see <http://www.gnu.org/licenses/>.
+
+-- This file contains Lua functions to support spacing of syllables.
+
+-- GREGORIO_VERSION 6.1.0
```
rpspringuel: That version number should be 6.2.0.  You also need to add this file to the list of files that VersionManager.py maintains.

rpspringuel (commented):
The addition of the Lua comments makes it much easier to read through the new Lua file.  Thank you.

Still need CHANGELOG entry.  Based on what I saw, I don't think any UPGRADE entry is needed.

davidweichiang:
I believe merging #1743 first will make this one easier, as it will fix a few failing tests.

tex/gregoriotex-main.tex
davidweichiang: Actually I think my comment may not be entirely accurate. If the break occurs mid syllable, then the second set of kerns will not be discarded. (But both kerns will be wrong, #1738.)

I agree that this third set of kerns is ugly, but I’m hoping that the resolution of #1738 will make it go away.

davidweichiang:
To fix:
- [x] When a hyphen is added, kerning is not happening when it should

tex/gregoriotex-main.tex
davidweichiang: I tried to improve the commenting in this macro. In principle, the enddifference and eolshift kerns should be emitted wherever there is a potential break. Since this syllable has two potential line breaks, it makes sense for it to have two sets of kerns.

davidweichiang:
The last outstanding issue with this is that there are several settings that are read at pre-linebreak time. If the user changes any of these mid-score, the change will affect the entire score, which is probably not what they want.

Currently, I believe it's just:
- maximumspacewithoutdash
- intersyllablespacestretchhyphen
- \ifgre@showlyrics
- \ifgre@rewritesyllables
This list will grow as more code gets moved into Lua. 

I think it will be better to save these settings per-syllable. Should it just save all of them, or are there any settings that are inconceivable that anyone would want to change them mid-score?

rpspringuel:
I would normally have said all those were score-wide settings, but as I think about it, I can conceive of situations where someone might want to change them mid-score, at least for a few syllables.

My question now is whether we should try to implement this as a defaults-exceptions scheme (where the changes inside a score create exceptions for some number of syllables, up to the rest of the score, but leave the original settings intact), or simply allow the user to change the settings mid-score and rely on them to change things back for the next score if that is what they want>

davidweichiang:
It looks like mid-score changes persist to the next score, so I think that should stay the same. It also looks like:
- `[ev:\gresetsyllablerewriting{off}]` drops characters: #1756
- `<v>\gresetsyllablerewriting{off}</v>` (or any other setting) inside text has no effect, because text is expanded inside a group
- `[ev:\grechangedim{maximumspacewithoutdash}{1cm}{scalable}])` takes effect starting from the following syllable, which I find mildly surprising but I guess it's ok:
```\gabcsnippet{a(g)a(g[ev:\grechangedim{maximumspacewithoutdash}{1cm}{scalable}])a(g)a(g)}```
image

**Comment by davidweichiang** [comment]
The above commit was the last thing I wanted to fix, so I think this could be really ready now.

**Comment by davidweichiang** [comment]
@rpspringuel This PR has a sequel which moves bar syllable spacing into Lua as well. Do you prefer it to be a separate PR, or do you prefer combining them into one giant one?

**Comment by rpspringuel** [comment]
I'm nearly done reviewing this PR, so please do a new one.  Make that one a draft until this one is merged so that I don't start looking at it prematurely.

**Comment by davidweichiang** [comment]
There are still 5 tests failing when this is tested against https://github.com/gregorio-project/gregorio-test/pull/420. I didn't go ahead and accept these because the changes were not small and it wasn't clear whether the new results are better or worse. (My opinion: neither better nor worse.)

These four are related to clef changes, where the old code uses the wrong nextbegindifference (issue #1724 case 1, and issue #1191):
- gabc-output/glyphs/clef_change.gabc
- gabc-output/glyphs/only-special.gabc
- gabc-output/edge_cases.gabc
- gabc-output/glyphs/custos_linebreak.gabc

For the fifth one, my notes say that "this is some complicated interaction between syllable rewriting and eolshift" but I don't remember the details:
- tex-output/bugs/fix-984/fix-984.tex

**Comment by rpspringuel** [comment]
Looking at those tests, gabc-output/glyphs/custos_linebreak.gabc is definitely a better result as we see proper protrusion of the hyphen in the margin on line 2 and the note glyph being right up against the end of the line when there is not custos or syllable text to force it inwards on line 4.  Both of those are more in line with how I think those mechanisms were supposed to work.

For gabc-output/glyphs/clef_change.gabc, gabc-output/glyphs/only-special.gabc, and gabc-output/edge_cases.gabc, I agree this is a six-of-one/half-dozen-of-another situation.

For tex-output/bugs/fix-984/fix-984.tex, I think what is happening in the old text is that the "l" in "Sal" is getting pushed into "va" in case there's an "lv" ligature (which most fonts won't have) and this has the side-effect of pushing the "Sal" syllable down to the next line so that "Salva" appears all at once (i.e. ligature formation is favored over the line break).  Your code either recognizes that "lv" is not a ligature, or favors line breaking over ligatures.  Does that sound like something your code might be doing?

**Comment by davidweichiang** [comment]
> For tex-output/bugs/fix-984/fix-984.tex, I think what is happening in the old text is that the "l" in "Sal" is getting pushed into "va" in case there's an "lv" ligature (which most fonts won't have) and this has the side-effect of pushing the "Sal" syllable down to the next line so that "Salva" appears all at once (i.e. ligature formation is favored over the line break). Your code either recognizes that "lv" is not a ligature, or favors line breaking over ligatures. Does that sound like something your code might be doing?

I don't think either version of the code makes decisions based on whether ligatures are possible or not. Both versions are supposed to suppress syllable-rewriting across line breaks. So the difference is elsewhere, and I apologize that I can't remember what the "complicated interaction with eolshift" was.

**Comment by rpspringuel** [comment]
Well, just looking at the results, the change isn't anything that concerns me based on our existing test.  I'd accept all 5 of those new results and call this good.

**Comment by davidweichiang** [comment]
OK, done.

File: CHANGELOG.md [hunk 3]
```diff
@@ -10,6 +10,8 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed a bug in horizontal spacing when the first syllable consists of only a bar. See [PR #1741](https://github.com/gregorio-project/gregorio/pull/1741).
 - Fixed a bug in horizontal spacing before a soft alteration. See [PR #1761](https://github.com/gregorio-project/gregorio/pull/1761).
 
+### Added
+- Code for several features related to horizontal spacing (spacing between non-bar syllables, syllable rewriting, clearing, and hyphenation) was moved into Lua. This results in some changes in horizontal spacing, which, if perceptible, should be improvements. See [#1720](https://github.com/gregorio-project/gregorio/issues/1720).
 ### Changed
```
rpspringuel, line 16: There should be a blank line between the sections here.

PR #1727, opened 2026-03-12
Fix segfault on empty GRE_ELEMENT when nabc_state wraps mid-syllable (#1726)

### LINKED ISSUE #1726: Segfault when nabc_state wraps to 0 mid-syllable creating empty GRE_ELEMENT
## Description

When `nabc-lines` is set to 2, the NABC state cycles modulo 3 (0 = GABC notes, 1 = voice 1, 2 = voice 2). If a syllable contains a GABC syntax error with an extra `|` separator — e.g. `||` where only `|` was intended — `nabc_state` can wrap back to 0 mid-syllable. A standalone `NABC_CUT` at state 0 then creates an empty `GRE_ELEMENT` with `first_glyph == NULL`.

Later, `write_default_end_of_element()` in `gregoriotex-write.c` dereferences `element->u.first_glyph->next` without a NULL check, causing a segmentation fault.

## Context

In the GABC notation for two NABC voices, a single `|` advances to the next NABC state while `||` advances by two. With `nabc-lines: 2`, the correct way to start a new GABC element within the same syllable after voice 2 is a single `|` (which wraps the state from 2 back to 0). Writing `||` instead is a syntax error that wraps to 0 and then immediately advances to 1, leaving an empty GABC element in between.

For example, the correct notation is:

```
ma(f!hhf||ta//tghh|gf||cl)
```

The erroneous notation (extra `|` before `gf`) triggers the crash:

```
ma(f!hhf||ta//tghh||gf||cl)
```

## Minimal reproducer

```gabc
name:t;
nabc-lines:2;
%%
(f3) ma(f!hhf||ta//tghh||gf||cl) (::)
```

Running `gregorio` on the above file produces a segfault (core dump).

## Expected behavior

Gregorio should not crash on malformed input. It should ideally report the syntax error (unexpected state wrap) and handle the empty element gracefully.

lbssousa:
Fixes #1726.

When `nabc_state` wraps back to 0 mid-syllable (e.g. with `nabc-lines: 2` and more `||`-groups than one cycle), the parser creates an empty `GRE_ELEMENT` with `first_glyph == NULL`. `write_default_end_of_element()` then dereferences `first_glyph->next` unconditionally, causing a segfault.

This adds a NULL check for `element->u.first_glyph` before iterating glyphs to find the last pitch for unison detection.

### Reproducer

```gabc
name:t;
nabc-lines:2;
%%
(f3) ma(f!hhf||ta//tghh||gf||cl) (::)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Fixed a bug that could cause a punctum mora that is supposed to be below the line (`.0`) to appear above the line. This bug was platform-dependent and was observed on a Windows system.  See [#1642](https://github.com/gregorio-project/gregorio/issues/1642).
 - Error messages from executable have been cleaned up to be more uniform.  See [#1644](https://github.com/gregorio-project/gregorio/issues/1644).
 - NABC neumes are now rendered for syllables with empty GABC/NABC snippets when NABC content is present (e.g. `(|vi|ta)`, `(|vi)`, `(||ta)`, `(g||ta)`).  See [#1700](https://github.com/gregorio-project/gregorio/issues/1700).
+- Fixed a segfault related to NABC state wrapping.  See [#1726](https://github.com/gregorio-project/gregorio/issues/1726).
 
 ## [Unreleased][CTAN]
 ### Fixed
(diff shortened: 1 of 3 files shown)


rpspringuel:
I'd like to see a scripted test that triggers the warning message; though actually, I think there's a hole in our tests right now in testing the various messages.  Maybe a gabc-gtex test that would segfault before the fix and doesn't after would be more in line with our current test set?

rpspringuel:
I can't merge this without the conflicts being resolved.

PR #1730, opened 2026-03-14
Add shim for lfs.mkdirp for compatibility with LuaTeX < 1.18

### LINKED ISSUE #1728: Lua lfs.mkdirp requires LuaTeX 1.18
A call to lfs.mkdirp was added in commit 240c6d413 (sorry). As this was released only in 2024, a shim should probably be written for compatibility with older versions.

davidweichiang:
Fixes #1728.


rpspringuel:
Going back to TeX Live 2023 (the last version where LuaTeX < 1.18) I can compile most of the test repository.  The tests that break seem to be ones related to other issues (like fancyhdr-hooks).

PR #1736, opened 2026-03-26
Fix and clarify messages related to selection of vowel rules

rpspringuel:
As I was looking at the output messages of various tests, I noticed that the messages were regularly saying that Latin vowel rules weren't found in the vowel rules files and that gregorio was falling back to Latin instead.  I found this kind of confusing, so I went looking at the code related to this and discovered that the rule appeared to be backwards: triggering the message required "latin" (in various forms) to be requested and not found.  Really, this message should be used when any non-"latin" language is requested and not found.

This fixes that, and also introduces a distinct message to be used when the default Latin rules are used.  I originally was going to skip the look-up loop if the language was Latin (as the rules for Latin are built-in to the executable), but I realized that users might want to customize the Latin rules themselves, in which case we should still be checking the vowel rule files for those customizations.

I have not, at this point, looked at how this affects the tests nor updated the documentation, so for the moment I'm simply looking for comments on this change.


davidweichiang:
Do "default" and "internal" mean the same thing? If so, I think the same word should be used. I can't really think of a clear way to distinguish the error messages and wonder if they really need to be distinguished.

davidweichiang:
Maybe “Selecting Latin instead of specified language” and “Using default rules for Latin instead of specified rules”?

rpspringuel:
The executable has a set of rules for Latin built into it and these are the rules that we fall back on when we can't find rules for the selected language in gregorio-vowels.dat.  Additionally, if the selected language is Latin, we don't just jump straight to these rules, but still check gregorio-vowels.dat for Latin rules in case the user wants to customize them.  If there are no custom Latin rules, then we use the builtin rules.  So, while "default" and "internal" are referring to the same set of rules, when I say "default" I'm referring to the first process, while when I say "internal" I'm referring to the second.  The difference in vocabulary is meant to distinguish not so much the rules being used, the reason for using those rules.  And as I think about it, the first one should probably be a warning (because we're ignoring user input) while in the second case, we're just informing the user of what we're doing (which is most likely, exactly what they want us to do).

davidweichiang (approved):

src/characters.c
```diff
@@ -108,7 +129,9 @@ static bool read_vowel_rules(char *const lang) {
 void gregorio_set_centering_language(char *const language)
 {
     if (!read_vowel_rules(language)) {
-        if (strcmp(language, "Latin") != 0 && strcmp(language, "latin") != 0 && strcmp(language, "la") != 0 && strcmp(language, "lat") != 0) {
+        if (strcasecmp(language, "latin") != 0 ||
+            strcasecmp(language, "la") != 0 ||
+            strcasecmp(language, "lat") != 0) {
             gregorio_messagef("gregorio_set_centering_language",
```
rpspringuel: I think this warning may be redundant to the messages that `read_vowel_rules()` already generates (and those are more detailed, with the changes above).  Can anyone think of a situation in which this message would occur and `read_vowel_rules()` would not have already provided more information about the failure?

src/characters.c
davidweichiang: If `"latina"` is aliased to `"latin"`, then here `language` gets set to `"latin"` but `is_latin` does not get updated.

src/characters.c
davidweichiang: Agreed

src/characters.c
```diff
@@ -85,8 +91,23 @@ static bool read_vowel_rules(char *const lang) {
...
+            gregorio_messagef(
+                "read_rules",
+                VERBOSITY_WARNING,
+                0,
+                "Selecting Latin instead of %s",
+                language
+            );
+        }
     }
     if (status == RFPS_ALIASED) {
```
davidweichiang: Could change to `else if` to make structure more clear

src/characters.c
davidweichiang: This message is not as informative as the others; maybe add `"Selecting Latin instead of %s", lang`.

rpspringuel: I don't think this is important.  If you set an new alias for Latin but don't provide custom Latin rules, then the "unresolved alias" messages get triggered, not the "not found" messages, and it's only in the "not found" messages that `is_latin` gets used.  If you do provide custom Latin rules, then neither set of messages would be triggered.

rpspringuel: I've updated the message to both indicated the unresolved alias, and the fallback to Latin.

rpspringuel:
I've added some tests to look at the new messages.  Looking at the messages produced, I think there's some further refinement needed here as the messages get really repetitive and potentially deceptive (especially with aliases).  Please take a look at https://github.com/gregorio-project/gregorio-test/pull/424 and tell me what you think.

src/characters.c
davidweichiang: I see, but it would also be confusing to get the error "Unable to resolve alias for Latina. Selecting Latin instead", wouldn't it?

rpspringuel: Definitely.  I'm experimenting with an alternative strategy that will fix this.

davidweichiang:
The "Aliased to language in file" messages are wrong, and the fix depends on what the search order is supposed to be (both untested):

- Change line https://github.com/gregorio-project/gregorio/blob/a702023893bca361029b55654ed248181ed150bf/src/characters.c#L64 so that the loop condition is `status == RFPS_NOT_FOUND`. 
- At the beginning of function https://github.com/gregorio-project/gregorio/blob/a702023893bca361029b55654ed248181ed150bf/src/vowel/vowel.c#L237, insert `*status = RFPS_NOT_FOUND;`.

The expected output of the first fix would be:
```
Looking for first in ./gregorio-vowels.dat
Aliasing first to second
Aliasing second to first
Aliased to first in ./gregorio-vowels.dat
Looking for first in ./gregorio-vowels.dat
Aliasing first to second
Aliasing second to first
Aliased to first in ./gregorio-vowels.dat
warning:Unable to resolve alias for first. Selecting Latin instead
```

The expected output of the second fix would be:
```
Looking for first in ./gregorio-vowels.dat
Aliasing first to second
Aliasing second to first
Aliased to first in ./gregorio-vowels.dat
Looking for first in /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregorio-vowels.dat
Could not find first in /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregorio-vowels.dat
etc.
Looking for first in ./gregorio-vowels.dat
Aliasing first to second
Aliasing second to first
Aliased to first in ./gregorio-vowels.dat
Looking for first in /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregorio-vowels.dat
Could not find first in /usr/local/texlive/texmf-local/tex/luatex/gregoriotex/gregorio-vowels.dat
etc.
warning:Unable to resolve alias for first. Selecting Latin instead
```

The second fix preserves the existing logic, but I think the first way makes more sense. For example, suppose:
- the local gregorio-vowels.dat has "alias b to c" to "alias a to b", in that order
- the global gregorio-vowels.dat has rules for both b and c
Then under the current code, if the user selects language a, they'll get the rules for b, with confusing messages.
Under the first fix, they'll get the rules for c. Under the second fix, they'll get the rules for b.

Also note that if a chain of aliases is too long, then this algorithm would falsely detect a cycle.

rpspringuel:
I've tried a new strategy which I think should make things clearer.  Please look at the tests: https://github.com/gregorio-project/gregorio-test/pull/424/changes/0d210b9becaeb7cb7d08a6a2662ad02100704224

I still need to make the expectations independent of my computer, so unless your installation pattern is identical to mine the tests will likely fail on your machine, but let me know if the expectations make sense.

src/characters.c
```diff
@@ -64,60 +81,169 @@ static bool read_vowel_rules(char *const lang) {
...
+                
+                /* Recheck to see if the alias points to latin
+                 * that way if there are no custom latin rules we won't trigger
+                 * a warning when we fall back to the internal latin rules */
+                is_latin =
+                    (strcasecmp(language, "latin") == 0 ||
+                     strcasecmp(language, "la") == 0 ||
+                     strcasecmp(language, "lat") == 0);
+
+                alias_found = true;
```
davidweichiang: I think this variable might not be needed, since you can just check status == RFPS_ALIASED below?

davidweichiang:
That looks good to me (and makes me wish this were in Lua).

src/characters.c
rpspringuel: I was able to remove that variable and found some other things that hopefully clear up the loop logic.

rpspringuel:
Okay, the tests should now pass on any machine and not just one which is setup exactly like mine.  I'll confirm that later this week when I have time to boot up one of my VMs.  Right now I'm going for lunch and then a bike ride.

rpspringuel:
Okay.  I've tested this on my Windows VM and things work there too.  I'm ready to merge this if there are no further comments.

Issue #1732, opened 2026-03-20
Windows Test Failures


rpspringuel:
I'm slowly getting Windows to a state where it can run all the tests against TeX Live 2026 and Gregorio 6.2.0-beta1 where both have been installed via their respective installers.  I'm using Cygwin to provide the testing tools (bash, ImageMagick, etc).  The whole test suite doesn't run yet (there are still some setup and compatibility issues to be resolved), but I'll note the tests that fail here when I'm reasonably certain that the test failure is not due to those issues.  If anyone has any insight into the a particular test failure, please chime in.

# 6.2.0-beta3 updates
## gabc-dump
gabc-dump/grow_buffers/long_path_01/long_path_02/long_path_03/long_path_04/long_path_05/long_path_06/long_path_07/long_path_08/long_path_09/long_path_10/long_path_12/long_filename_long_filename_long_filename_long_filename_long_filename_long_filename_long_filename_long_filename_long_filename_long_filename.gabc: Complains about being unable to open the log file for writing.  This is a 310 (314 for the log) character path.  So this is almost certainly a problem with Windows long-path limit (260 characters).  Fixing requires reworking source to support long-paths on Windows. See #1740.  Will not fix before 6.2.0 release.
gabc-dump/should-fail/bad-vowel-file/bad-vowel-file.gabc: Fixed by https://github.com/gregorio-project/gregorio-test/pull/411

## gabc-gabc
All Pass

## gabc-gtex
gabc-gtex/english.gabc : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
gabc-gtex/hospodine.gabc : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411

## scripted
scripted/cli/bad_informat.sh : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
scripted/cli/bad_outformat.sh : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
scripted/cli/debug-bad_err.sh : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
scripted/cli/debug-bad_out.sh : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
scripted/cli/debug.sh : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
scripted/cli/deprecation.sh : Fixed by https://github.com/gregorio-project/gregorio-test/pull/411
[...]

PR #1739, opened 2026-04-02
Handle option errors internally

rpspringuel:
While working on #1732, I noticed that the `invalid option` error had some quirks to it that made it inconsistent with our other errors.  After some investigation, I discovered this was because `getopt_long()` was producing part of the error message and it has its own sense of how to format the message (in a way that wasn't consistent with ours).  I thus disabled `getopt_long()`'s error handling and modified our option handling to account for an invalid option (formatting the message the way we want it to be).

There will be some test implications from this (scripted/cli/invalid_option.sh and related), but I don't have time to make the corresponding PR on the test repository tonight, hence I'm marking this as DRAFT for now.  Once I have the chance to futz with the tests, I'll mark this ready for review.

### DIFF
--- a/src/gregorio-utils.c
+++ b/src/gregorio-utils.c
@@ -549,7 +549,7 @@ int main(int argc, char **argv)
             break;
         case '?':
             if (optopt) {
-                fprintf(stderr, "%s: invalid option -- '%c'\n", argv[0], optopt);
+                fprintf(stderr, "%s: invalid option '%c'\n", argv[0], optopt);
             } else {
                 fprintf(stderr, "%s: invalid option\n", argv[0]);
             }
@@ -558,7 +558,7 @@ int main(int argc, char **argv)
             break;
 
         case ':':
-            fprintf(stderr, "%s: option requires an argument -- '%c'\n",
+            fprintf(stderr, "%s: option '%c' requires an argument \n",
                     argv[0], optopt);
             print_short_usage(argv[0]);
             gregorio_exit(1);
--- a/src/gregorio-utils.c
+++ b/src/gregorio-utils.c
@@ -331,7 +331,7 @@ int main(int argc, char **argv)
     bool debug = false;
     bool must_print_short_usage = false;
     int option_index = 0;
-    static const char *const options = "o:SF:l:f:shOLVvWDpd";
+    static const char *const options = ":o:SF:l:f:shOLVvWDpd";
     static const struct option long_options[] = {
         {"output-file", 1, 0, 'o'},
         {"stdout", 0, 0, 'S'},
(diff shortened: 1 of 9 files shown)


rpspringuel:
Okay, I've done my testing and this is ready for review.

src/gregorio-utils.c
```diff
@@ -541,12 +548,25 @@ int main(int argc, char **argv)
             debug = true;
             break;
         case '?':
-            must_print_short_usage = true;
+            if (optopt) {
+                fprintf(stderr, "%s: invalid option -- '%c'\n", argv[0], optopt);
```
davidweichiang: I'm wondering if someone might think that the `--` was part of the invalid option. I think it's more usual just to write `invalid option '%c'`.

src/gregorio-utils.c
```diff
@@ -541,12 +548,25 @@ int main(int argc, char **argv)
...
+            } else {
+                fprintf(stderr, "%s: invalid option\n", argv[0]);
+            }
+            print_short_usage(argv[0]);
+            gregorio_exit(1);
+            break;
+
+        case ':':
+            fprintf(stderr, "%s: option requires an argument -- '%c'\n",
```
davidweichiang: Same as above

davidweichiang: I would expect to see `option '%c' requires an argument`

src/messages.h
```diff
@@ -80,10 +80,7 @@ int gregorio_get_return_value(void);
 #define gregorio_not_null_ptr(VARIABLE,FUNCTION,ON_FALSE) \
     gregorio_assert(VARIABLE && *VARIABLE, FUNCTION, #VARIABLE " may not be null", ON_FALSE)
 
-#define gregorio_fail(FUNCTION,MESSAGE) \
-    gregorio_message(_(MESSAGE), #FUNCTION, VERBOSITY_ASSERTION, __LINE__)
-
-#define gregorio_fail2(FUNCTION,FORMAT,ARG) \
-    gregorio_messagef(#FUNCTION, VERBOSITY_ASSERTION, __LINE__, FORMAT, ARG)
+#define gregorio_fail(FUNCTION, FORMAT, args...) \
```
davidweichiang, line 83: Variadic macros were introduced in C99. I think that is sufficiently old to be fine (although the fact that variadic macros are new to me says something about when I learned C). But the current build system compiles with `-std=gnu89`, so something would have to change to be made consistent. Cf. #1650.

davidweichiang (changes requested):

src/gregorio-utils.c
rpspringuel: Makes sense.  I'll fix that.

src/gregorio-utils.c
rpspringuel: Now that you say that, it totally makes more sense.

src/messages.h
rpspringuel: This trick is technically making use of the GNU extensions rather than C99 variadic macros.

I agree in principle with C99 being sufficiently old to be fine.  I just didn't want to undertake the work to ensure the entire program and build system was compatible with that standard in this PR and was looking for the minimal change.  It does raise a whole bunch of warnings, which can be ignored by adding the flag `-Wgnu-zero-variadic-macro-arguments` when compiling, but again I was looking for the minimal change and we're already ignoring some warnings.

I think the logical long-term question is whether it's worth the squeeze to shift the whole codebase to C99 standards.  If you're able to port everything over to Lua, then the C standards become moot, but if we're going to stick to the C code for a while, that might be a reasonable goal.

davidweichiang (approved):
Looks good to me.

PR #1741, opened 2026-04-06
Unset \gre@ifbeginningofscore in \GreBarSyllables

davidweichiang:
Previously, beginningofscore would remain set if the first syllables were \GreBarSyllables. This would cause the first \GreSyllable to be set as if it were the first syllable of the line (with bolshift). Now, the first \GreSyllable should be set correctly.


davidweichiang:
This changes a few tests, but the one that seems iffy to me is tex-output/fancyhdr-hooks/gregoriotest.tex. This one starts with a no-note syllable, which now has the effect of cancelling out the bolshift of the second syllable.

image

PR #1748, opened 2026-04-19
Deprecation path advancement

rpspringuel:
Remove OBSOLETE code
Code which was DEPRECATED in the background (i.e. pending modification of executable) is modified.  This really should have been done for 6.2 (since we modified the executable in that release), but was overlooked.

### DIFF
--- a/src/gregoriotex/gregoriotex-write.c
+++ b/src/gregoriotex/gregoriotex-write.c
@@ -4693,7 +4693,7 @@ void gregoriotex_write_score(FILE *const f, gregorio_score *const score,
     if (score->first_voice_info) {
         clef = score->first_voice_info->initial_clef;
     }
-    fprintf(f, "\\GreBeginScore{%s}{%s}{%u}{}%%\n",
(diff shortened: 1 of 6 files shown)


src/gregoriotex/gregoriotex-write.c
```diff
@@ -4712,21 +4693,10 @@ void gregoriotex_write_score(FILE *const f, gregorio_score *const score,
     if (score->first_voice_info) {
         clef = score->first_voice_info->initial_clef;
     }
-    /* After removing deprecated arguments, this will become:
     fprintf(f, "\\GreBeginScore{%s}{%s}{%u}{}%%\n",
```
davidweichiang: If the fourth argument is always empty, why not remove it as well? (I know I wrote this line, but I don't remember the reason.)

rpspringuel: We probably should remove it.  I can't find anything in the codebase that indicates that the empty argument is still used.

davidweichiang (approved):

PR #1753, opened 2026-04-30
Generalize mechanism for turning off callbacks during headers

### LINKED ISSUE #1752: Fatal error: "Can't raise/lower a rule (this shouldn't happen)" when a score is split across pages in v6.2.0
**Gregorio version:** 6.2.0
**LuaTeX version:** LuaHBTeX 1.22.0 (TeX Live 2025)
**OS:** Linux (NixOS)

### Description

In testing, a multiline score compiles successfully in some reduced contexts,
but fails in Gregorio 6.2.0 when the score is allowed to break across pages.
When the same score is forced to stay on a single page by increasing
`\textheight`, the error disappears.

### Error message

```text
warning  (node filter): error: ...ltluatex.lua:109:
Module gregoriotex Error: Can't raise/lower a rule (this shouldn't happen)

l.<varies> \GreEndScore
==> Fatal error occurred, no output PDF file produced!
```

The error is raised in `gregoriotex.lua` at the `shift_up` function (line 933),
which is called from `adjust_additional_spaces`, which in turn is called from
the `post_linebreak` callback triggered by `\GreEndScore` -> `\par`.

### Reproducer

**Minimal LaTeX file** (all that is needed to trigger the error):

```latex
\documentclass{memoir}

\setlength{\textheight}{220pt}

\begin{document}
\chapter{Test}
\section{Score}
\gregorioscore{path/to/multiline/score}
\end{document}
```

**Key requirements:**
- Document class: `memoir`
- Package: `gregoriotex` with `[autocompile]` option
- `\setlength{\textheight}{220pt}` — Forces page break during score rendering (essential to trigger the bug)
- A multiline score

**Compilation result:**

```text
Underfull \vbox (badness 10000) has occurred while \output is active
warning  (node filter): error: ...ltluatex.lua:109:
Module gregoriotex Error: Can't raise/lower a rule (this shouldn't happen)
l.<varies> \GreEndScore
==> Fatal error occurred, no output PDF file produced!
```

The fatal error is reproducible with a public Kyriale score from GregoBase
(`Gloria XV`, id `2975`), providing a standalone public reproducer.
The exact line number at `\GreEndScore` may vary across runs.

In this environment, an attempted further reduction to `\documentclass{article}`
without `\chapter` did **not** reproduce the crash (same score, forced small
`\textheight`), which suggests the trigger may depend on class/output-routine
behavior in addition to score page breaking.

### Root cause analysis

`adjust_additional_spaces` (introduced by PR #1641 - one-pass line spacing)
contains a `shift_up` function that iterates over all top-level nodes in
`line.head`. It handles `hlist`, `vlist`, and `disc`, but raises a fatal error
on `rule` nodes:

```lua
local function shift_up(n)
    if n.id == rule then
        err("Can't raise/lower a rule (this shouldn't happen)")
    elseif n.id == hlist or n.id == vlist then
        n.shift = n.shift - everything_raise
    ...
end
```

The current evidence suggests that the problematic `rule` node is exposed at the
top level of `line.head` specifically when the score is being broken by the page
builder / output routine. In the failing reduced reproducer, the log contains:

```text
Underfull \vbox (badness 5288) has occurred while \output is active
```

immediately before the Gregorio fatal error. That strongly suggests interaction
between Gregorio's one-pass line-spacing logic and page breaking during score
output.

The exact mechanism is still not fully understood, but the likely failure mode is:

1. `\GreEndScore` ends the paragraph and triggers `post_linebreak_filter`.
2. `adjust_additional_spaces()` later calls `shift_up()` over top-level nodes.
3. During a page-breaking scenario, a `rule` node appears directly at top level.
4. `shift_up()` treats that as impossible and aborts.

The problematic assumption is that node shapes remain stable when the score is split across pages.
This appears to be violated specifically in page-breaking scenarios with multiline scores.

### Workaround

Keeping the score from breaking across pages avoids the error in our tests. For
example, increasing `\textheight` enough for the whole score to fit on one page
makes the reduced reproducer compile successfully.

```latex
\setlength{\textheight}{2000pt}
```

That is only a diagnostic workaround, not a practical fix for real books.

### Expected behavior

Either:

1. `shift_up` should handle `rule` nodes gracefully during score/page breaking,
    or
2. the page-break path should ensure such `rule` nodes are wrapped or filtered
    before `shift_up` runs.

### Additional context

- Likely regression in the 6.2.0 line-spacing/page-break path.
- Public standalone test score (GregoBase, Gloria XV):
    `https://gregobase.selapa.net/download.php?id=2975&format=gabc&elem=1`

davidweichiang:
Previously, hooks provided by fancyhdr were used to turn off callbacks to keep the headers/footers from being processed as if they were score lines. This uses LuaTeX hooks instead, so it works with memoir as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 
 ## [Unreleased][develop]
 ### Fixed
-- Fixed a bug that could cause problems with headers/footers, and in particular an error with the `memoir` class. See [#1753](https://github.com/gregorio-project/gregorio/pull/1753).
+- Fixed a bug that could cause problems with headers/footers by generalizing the fix for fancyhdr (see [#1610](https://github.com/gregorio-project/gregorio/pull/1610) to work with any headers/footers. In particular, it fixes a similar problem with the `memoir` class. See [#1753](https://github.com/gregorio-project/gregorio/pull/1753).
 
 ## [Unreleased][CTAN]
 *Note:* 6.2.0 was not released to CTAN and is not compatible with 6.1.0 which is on CTAN.  Please make all changes against develop until this is resolved.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][develop]
-
+### Fixed
+- Fixed a bug that could cause problems with headers/footers, and in particular an error with the `memoir` class. See [#1753](https://github.com/gregorio-project/gregorio/pull/1753).
 
 ## [Unreleased][CTAN]
 *Note:* 6.2.0 was not released to CTAN and is not compatible with 6.1.0 which is on CTAN.  Please make all changes against develop until this is resolved.
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -951,8 +951,6 @@ local function adjust_additional_spaces(line, info, linenum)
   end
 end
 
--- in each function we check if we really are inside a score,
--- which we can see with the dash_attr being set or not
 local function post_linebreak(h, groupcode, glyphes)
   --dump_nodes(h)
   -- TODO: to be changed according to the font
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). It follows [some conventions](http://keepachangelog.com/).
 
 ## [Unreleased][develop]
-
+### Fixed
+- Fixed a bug that could cause problems with headers/footers, and in particular an error with the `memoir` class. See [#1753](https://github.com/gregorio-project/gregorio/pull/1753).
```
rpspringuel: I'd phrase this differently:
```
Fixed a bug that could cause problems with headers/footers by generalizing the fix for fancyhdr (see [#1610](https://github.com/gregorio-project/gregorio/pull/1610).  Resolves similar error with the `memoir` class.  See [#1753](https://github.com/gregorio-project/gregorio/pull/1753).
```

That way it's clearer what's happening.

PR #1743, opened 2026-04-08
Fix previousenddifference and bar@lastskip in discretionaries (clef changes)

davidweichiang:
The bar spacing algorithm uses the previous syllable's enddifference, but in a discretionary (clef change), the "replace" syllable uses the "pre" syllable's enddifference, but it should be the syllable before the discretionary's enddifference. This commit saves and restores the enddifference correctly.

Fixes case 3 of https://github.com/gregorio-project/gregorio/issues/1724
Fixes https://github.com/gregorio-project/gregorio/issues/1191

PR #1754, opened 2026-05-03
Add missing \GreNoBreak

davidweichiang:
This adds a `\GreNoBreak` at a place where a line break could unintentionally occur (at a kern followed by a glue; putting the penalty in between makes it impossible to break at the kern).


davidweichiang:
Sorry, duplicate of #1743.

PR #1759, opened 2026-05-28
fix(lua): replace os.spawn with io.popen to prevent fd-inheritance corruption

### LINKED ISSUE #1757: os.spawn fd-inheritance corrupts LuaTeX \input state in \gregorioscore and \gabcsnippet (triggered on clean builds)
## Summary

`tex/gregoriotex.lua` calls `os.spawn()` in **two** functions:

| Lua function              | Entry TeX macro                      | Line (6.2.0) | Line (`develop`) |
| ------------------------- | ------------------------------------ | ------------ | ---------------- |
| `get_prog_output(cmd, …)` | `\gabcsnippet` via `direct_gabc`     | 141          | 218              |
| `compile_gabc(…)`         | `\gregorioscore` via `include_score` | 1330         | 829              |

On Unix, `os.spawn()` is implemented as `fork()+execve()` **without** setting `O_CLOEXEC` on any file descriptor. The child process (`gregorio`) therefore inherits every fd open in the LuaTeX parent — including the fd of the `.tex` file currently being `\input`-ted. Because parent and child share the same kernel-level `struct file` (which holds the seek position `f_pos`), any fd activity in the child can silently advance the read position of files the parent is still reading. LuaTeX then processes tokens from the wrong position, producing hard-to-diagnose errors.

Both call sites are affected: `\gabcsnippet` reaches `os.spawn` through `get_prog_output`; `\gregorioscore` — the primary macro of every real-world Gregorio project — reaches it through `compile_gabc`.

The fix requires replacing `os.spawn` with `io.popen` at both sites using POSIX single-quote argument quoting, so that paths containing spaces or shell metacharacters remain safe.

## Conditions

The bug may manifest when **both** of the following are true:

1. `\gregorioscore` or `\gabcsnippet` (or any macro that calls them) is used inside a file loaded via `\input` — i.e., not in the main `.tex` job file itself.
2. The build is running from a **clean state** where no `.gtex` files are cached for the scores being compiled.

The bug was observed in at least two scenarios:

- **NixOS** build sandboxes and `nix develop` shells
- **GitHub Actions** with `xu-cheng/latex-action`

## Observed symptoms

The corruption of the shared `f_pos` can produce any of the following, depending on where the corrupted byte lands inside a TeX token:

| Corruption position                   | Observed symptom                                                          |
| ------------------------------------- | ------------------------------------------------------------------------- |
| Inside a control-sequence name        | `! Undefined control sequence. \gre` (i.e. `\gregorioscore` is truncated) |
| At a structural delimiter (`\\`, `&`) | `! LaTeX Error: There's no line here to end.` / `! Misplaced \noalign.`   |
| Inside paragraph text                 | Paragraph appears duplicated, or a text fragment is repeated in the PDF   |

### Case A — `\gabcsnippet`

**`main.tex`**

```latex
\documentclass{article}
\begin{document}
\input{fragment}
\end{document}
```

**`fragment.tex`**

```latex
\begin{tabular}{|c|c|}
  \hline
  Row 1 & \gabcsnippet{(e)} \\
  \hline
  Row 2 & \gabcsnippet{(f)} \\
  \hline
  Row 3 & \gabcsnippet{(g)} \\
  \hline
  Row 4 & \gabcsnippet{(a)} \\
  \hline
  Row 5 & \gabcsnippet{(b)} \\
  \hline
\end{tabular}
```

Compile with `lualatex --shell-escape main.tex` in a directory containing no cached `.gtex` files.

Expected error:

```
! Misplaced \noalign.
\hline -> ...
l.17   \hline
```

Note: `fragment.tex` has only 16 lines. The error at the non-existent line 17
is itself a symptom: LuaTeX skipped content due to the corrupted read position.

Moving the table directly into `main.tex` (without `\input`) compiles successfully — this is the control case.

### Case B — `\gregorioscore`

**`main.tex`**

```latex
\documentclass{article}
\begin{document}
\input{fragment}
\end{document}
```

**`fragment.tex`**

```latex
Pre-score paragraph.  This must appear exactly once in the output.

\gregorioscore{sample-1}

Mid-paragraph.  Must appear exactly once.

\gregorioscore{sample-2}

Post-score paragraph.  Must appear exactly once.
```

Compile with `lualatex --shell-escape main.tex` in a directory where `sample-1.gtex`, `sample-2.gtex`, and `sample-3.gtex` do not yet exist.

Expected symptom: the PDF is produced without any compile error, but
the log may show `sample-2.gtex` and `sample-3.gtex` included **twice**:

```
(./tmp-gre/.../sample-1-6_2_0.gtex)
(./tmp-gre/.../sample-2-6_2_0.gtex)
(./tmp-gre/.../sample-3-6_2_0.gtex)
(./tmp-gre/.../sample-2-6_2_0.gtex)   ← second time: content is duplicated in PDF
(./tmp-gre/.../sample-3-6_2_0.gtex)   ← second time
```

The corruption from the first `compile_gabc` call (for `sample-1`) causes LuaTeX
to re-read a portion of `fragment.tex`, processing `\gregorioscore{sample-2}` and
`\gregorioscore{sample-3}` a second time. The second invocations find the `.gtex`
already cached and include it again silently.

### `get_prog_output` (Case A)

```lua
local function get_prog_output(cmd, tmpname, fmt)
  local rc = os.spawn(cmd)  -- fork()+exec() inherits all open fds
  ...
end
```

### `compile_gabc` (Case B)

```lua
-- inside compile_gabc(), immediately before the error-handling branch:
kpse.record_output_file(gtex_file)
local res = os.spawn(cmd)  -- same problem
```

In both cases, `os.spawn` is implemented in `texk/web2c/luatexdir/lua/loslibext.c` as:

1. `fork()`
2. in the child: flush stdio buffers — does **not** close or set `FD_CLOEXEC` on any fd
3. `execve()` in the child
4. `waitpid()` in the parent

The child inherits every fd LuaTeX has open, including the fd of the currently `\input`-ted `.tex` file. Routine operations in the child — locale initialisation, dynamic-linker probing, internal `dup2` calls — can shift `f_pos`. Because the parent reads the same `struct file`, it resumes from the wrong position.

## Proposed fix

Replace `os.spawn` with `io.popen` at **both** call sites. `io.popen` invokes `/bin/sh -c`, which manages its own fd space before the final `exec`, eliminating the inheritance problem. On Windows, `os.spawn` maps to `CreateProcess`, which does not suffer from this issue (fds are not inherited by default), so the original path is preserved there.

Arguments must be individually POSIX-quoted before concatenation: passing an unquoted command string breaks on any path containing a space and allows shell injection if a filename contains metacharacters.

### Helper functions (add once, near the top of the Lua module)

```lua
local function gre_shellquote(s)
  -- POSIX single-quote wrapping; safe for any argument on Unix.
  return "'" .. s:gsub("'", "'\\''") .. "'"
end
local function gre_build_cmdstr(cmd)
  local quoted = {}
  for i, a in ipairs(cmd) do quoted[i] = gre_shellquote(a) end
  return table.concat(quoted, ' ')
end
-- os.spawn on Windows uses CreateProcess and is not affected; cmd.exe
-- quoting is entirely different, so keep the original path there.
local gre_use_popen = (os.type ~= 'windows')
```

### Patch for `get_prog_output`

```diff
 local function get_prog_output(cmd, tmpname, fmt)
-  local rc = os.spawn(cmd)
+  local rc
+  if gre_use_popen then
+    local handle = io.popen(gre_build_cmdstr(cmd), 'r')
+    rc = 1
+    if handle then
+      handle:read('*all')
+      local ok, _, code = handle:close()
+      if ok then rc = 0 else rc = code or 1 end
+    end
+  else
+    rc = os.spawn(cmd)
+  end
```

### Patch for `compile_gabc`

```diff
   kpse.record_output_file(gtex_file)
-  local res = os.spawn(cmd)
+  local res
+  if gre_use_popen then
+    local handle = io.popen(gre_build_cmdstr(cmd), 'r')
+    res = nil  -- nil = "exec failed → shell-escape probably off" (see branch below)
+    if handle then
+      handle:read('*all')
+      local ok, _, code = handle:close()
+      if ok then res = 0 else res = code or 1 end
+    end
+  else
+    res = os.spawn(cmd)
+  end
```

> **Note on `res = nil`:** the code immediately after this call tests `if res == nil then err("…shell-escape mode may not be activated…")`. Preserving `nil` when `io.popen` cannot open a handle keeps that diagnostic message intact.

### Regression risk

- **macOS / non-Nix Linux:** functionally neutral. `popen(3)` is POSIX standard on all platforms where LuaTeX runs. `gregorio`'s stdout is now captured in a pipe and discarded (`:read('*all')`), but important diagnostic output goes to the `.glog` file and to `stderr`, both of which are unaffected.
- **Windows:** `os.spawn` is preserved via the `gre_use_popen` flag; no change for Windows users.
- **Paths with spaces or special characters:** `gre_shellquote` handles these safely on all POSIX systems.

## Workaround

Pre-compile all `.gabc` files to `.gtex` **before** invoking `lualatex`:

```sh
for f in path/to/*.gabc; do
  gregorio -W -o "${f%.gabc}.gtex" -l "${f%.gabc}.glog" "$f"
done
lualatex --shell-escape main.tex
```

When `.gtex` files are already present, `include_score()` takes the cache branch and never calls `compile_gabc()`.

**Limitations:**

- Does **not** cover `\gabcsnippet`, which always calls `get_prog_output()` — there is no `.gtex` cache for snippets.
- The output filename must exactly match what `include_score()` constructs (including the internal-version suffix and the character-substitution applied to the basename). A mismatch causes a cache miss and re-triggers `compile_gabc()`.
- Fragile with `latexmk`/Makefile: any inadvertent `touch` of a `.gabc` file or a gregorio version bump invalidates the cache.

The workaround is useful as an immediate mitigation for CI pipelines but does not replace a code fix.

## Regression test for `gregorio-test`

The following three files can be dropped into `gregorio-project/gregorio-test` at `tests/tex-output/bugs/fix-1757/`. The `harness.sh` function `tex_output_find` already excludes `gsp-*.tex` files from independent test discovery; `run-lualatex.sh` compiles each test in a fresh `{filename}.out/` subdirectory (no `.gtex` is ever cached between runs), guaranteeing the bug path is exercised on every CI run.

**`fix-1757.tex`**

```latex
% !TEX TS-program = lualatex
% issue: https://github.com/gregorio-project/gregorio/issues/1757
% notes: Compiling \gregorioscore (or \gabcsnippet) from within an
%        \input-ted file corrupts LuaTeX's read position because
%        os.spawn() inherits LuaTeX's open file descriptors without
%        FD_CLOEXEC.
\documentclass{article}
\begin{document}
\input{gsp-fragment}
\end{document}
```

**`gsp-fragment.tex`**

```latex
% Exercises both affected os.spawn call sites in one compile:
%   * \gregorioscore -> compile_gabc -> os.spawn
%   * \gabcsnippet in a tabular -> get_prog_output -> os.spawn
% Without the fix, this fails or produces a duplicated paragraph on any
% clean-state CI run.

Pre-score paragraph.  Must appear exactly once.

\gregorioscore{fix-1757}

Mid-paragraph.  Must appear exactly once.

\begin{tabular}{|c|c|}
  \hline
  Row 1 & \gabcsnippet{(e)} \\
  \hline
  Row 2 & \gabcsnippet{(f)} \\
  \hline
  Row 3 & \gabcsnippet{(g)} \\
  \hline
  Row 4 & \gabcsnippet{(a)} \\
  \hline
  Row 5 & \gabcsnippet{(b)} \\
  \hline
\end{tabular}

Post-tabular paragraph.  Must appear exactly once.
```

**`fix-1757.gabc`**

```
name: fix-1757;
%%
(c4) Sa(g)lus(h) po(j)pu(h)li(g) e(h)go(g) sum.(g) (::)
```

After applying the fix and rebuilding gregorio:

```sh
./gregorio-test.sh -n tex-output/bugs/fix-1757   # generate output
./gregorio-test.sh -a tex-output/bugs/fix-1757   # accept (creates fix-1757.pdf)
```

Without the fix, the `-n` step fails consistently on containerised CI.

## Long-term alternative

`gregoriotex.lua` is one of many LuaTeX clients that call `os.spawn`. Other packages (`pythontex`, `sagetex`, `hyperxmp`, …) share the same underlying vulnerability. The correct permanent fix is in LuaTeX upstream:

- Mark all open file descriptors `O_CLOEXEC` before `fork()`, **or**
- Replace `fork()+execve()` with `posix_spawn()` using `POSIX_SPAWN_SETFD` attribute actions to explicitly close inherited descriptors.

Until an upstream fix lands, each client package must apply its own `io.popen` workaround.

lbssousa:
## Summary

- Replaces `os.spawn()` with `io.popen()` at both call sites in `tex/gregoriotex.lua` where a `gregorio` child process is launched.
- Adds three helpers (`gre_shellquote`, `gre_build_cmdstr`, `gre_use_popen`) defined once, before `get_prog_output`.
- Windows is unaffected: `gre_use_popen` is `false` when `os.type == 'windows'`, preserving the original `os.spawn()` path.

## Problem

`os.spawn()` on Unix is implemented as `fork()+execve()` without setting `FD_CLOEXEC` on any file descriptor.  The `gregorio` child process therefore inherits every fd open in the LuaTeX parent — including the fd of any `.tex` file currently being `\input`-ted.  Parent and child share the same kernel-level `struct file` (which holds the seek position `f_pos`); any fd activity in the child can silently advance the read offset LuaTeX still relies on, corrupting TeX's input state.

Both call sites are affected:

| Call site | Reached via |
|---|---|
| `get_prog_output()` | `\gabcsnippet` → `direct_gabc()` |
| `compile_gabc()` | `\gregorioscore` → `include_score()` |

The bug manifests only when `\gregorioscore` or `\gabcsnippet` is used inside an `\input`-ted file **and** no cached `.gtex` exists (clean-state build).  Observed on NixOS sandboxes and GitHub Actions.

Typical symptoms:

- `! Misplaced \noalign.` / `! Undefined control sequence. \gre` (hard error)
- Silent content duplication in the PDF (scores or paragraphs appear twice, no error)

## Fix

`io.popen()` invokes `/bin/sh -c`, which manages its own fd space before the final `exec`, eliminating the inheritance.  Arguments are individually POSIX-single-quoted by `gre_shellquote()` so paths with spaces or shell metacharacters remain safe.

Closes #1757

### DIFF
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -142,8 +142,44 @@ local catcode_at_letter = luatexbase.catcodetables['gre@atletter']
 
 local first_line_prevdepth = 0
 
+-- Single-quotes s for safe use as one word in a shell command line.
+local function gre_shellquote(s)
+  return "'" .. s:gsub("'", "'\\''") .. "'"
+end
+
+-- Unlike os.spawn(), which execs cmd's arguments directly, io.popen() only
+-- takes a single command line run through /bin/sh -c. Quote each argument so
+-- that paths with spaces stay intact and shell metacharacters in filenames
+-- (e.g. from the .tex source) aren't interpreted as shell syntax.
+local function gre_build_cmdstr(cmd)
+  local quoted = {}
+  for i, a in ipairs(cmd) do
+    quoted[i] = gre_shellquote(a)
+  end
+  return table.concat(quoted, ' ')
+end
+
+-- Returns exit code (0 = success, positive = failure, nil = could not
+-- launch — caller interprets nil as shell-escape disabled).
+local function gre_exec(cmd)
+  -- Workaround for a LuaTeX bug: os.spawn() leaks open fds to the child on
+  -- Unix, corrupting \input state (see #1757). Use io.popen() there instead;
+  -- Windows is unaffected and keeps os.spawn().
+  if os.type ~= 'windows' then
+    local handle = io.popen(gre_build_cmdstr(cmd), 'r')
+    if handle then
+      handle:read('*all')
+      local ok, _, code = handle:close()
+      return ok and 0 or (code or 1)
+    end
+    return nil
+  else
+    return os.spawn(cmd)
+  end
+end
+
 local function get_prog_output(cmd, tmpname, fmt)
-  local rc = os.spawn(cmd)
+  local rc = gre_exec(cmd) or 1
   local content = nil
   if rc == 0 then
     local f = io.open(tmpname, 'r');
@@ -1350,7 +1386,7 @@ local function compile_gabc(gabc_file, gtex_file, glog_file, allow_deprecated)
   kpse.record_input_file(gabc_file)
   kpse.record_output_file(glog_file)
   kpse.record_output_file(gtex_file)
-  local res = os.spawn(cmd)
+  local res = gre_exec(cmd)
 
   if res == nil then
     err("\nSomething went wrong when executing\n    '%s'.\n"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 - Fixed a few bugs related to horizontal spacing around bars and clef changes. See issues [#1191](https://github.com/gregorio-project/gregorio/issues/1191), case 3 of [#1724](https://github.com/gregorio-project/gregorio/issues/1724), [PR #1743](https://github.com/gregorio-project/gregorio/pull/1743), and [#1745](https://github.com/gregorio-project/gregorio/issues/1745).
 - Fixed a bug in horizontal spacing when the first syllable consists of only a bar. See [PR #1741](https://github.com/gregorio-project/gregorio/pull/1741).
 - Fixed a bug in horizontal spacing before a soft alteration. See [PR #1761](https://github.com/gregorio-project/gregorio/pull/1761).
+- Worked around a LuaTeX bug (file descriptor inheritance in `os.spawn()`) that could corrupt LuaTeX's input state when `\gregorioscore` or `\gabcsnippet` was used inside an `\input`-ted file on a clean build. See [#1757](https://github.com/gregorio-project/gregorio/issues/1757).
 
 ### Added
 - Code for several features related to horizontal spacing (spacing between non-bar syllables, syllable rewriting, clearing, and hyphenation) was moved into Lua. This results in some changes in horizontal spacing, which, if perceptible, should be improvements. See [#1720](https://github.com/gregorio-project/gregorio/issues/1720).


tex/gregoriotex.lua
```diff
@@ -137,8 +137,37 @@ local catcode_at_letter = luatexbase.catcodetables['gre@atletter']
 
 local first_line_prevdepth = 0
 
+-- On Unix, os.spawn() uses fork()+execve() and the child inherits every fd
+-- open in the LuaTeX parent (including any \input-ted .tex file).  Parent and
+-- child share the same kernel file description (f_pos), so any fd activity in
+-- the child can silently advance the read offset of files LuaTeX is still
```
davidweichiang: minor comment: it's actually silently rewinding the offset

tex/gregoriotex.lua
```diff
@@ -137,8 +137,37 @@ local catcode_at_letter = luatexbase.catcodetables['gre@atletter']
...
+local function gre_build_cmdstr(cmd)
+  local quoted = {}
+  for i, a in ipairs(cmd) do quoted[i] = gre_shellquote(a) end
+  return table.concat(quoted, ' ')
+end
+local gre_use_popen = (os.type ~= 'windows')
+
 local function get_prog_output(cmd, tmpname, fmt)
-  local rc = os.spawn(cmd)
+  local rc
```
davidweichiang: What do you think about making this a function to avoid code duplication with 1348-1359? (Or, even better, let `compile_gabc` call `get_prog_output`.)

davidweichiang (approved):

tex/gregoriotex.lua
lbssousa: Updated in commit [71ffc9c](https://github.com/gregorio-project/gregorio/pull/1759/commits/71ffc9c440357f2ea57fb81580d8ca2cd5be209a). Thanks!

tex/gregoriotex.lua
lbssousa: Updated in commit [71ffc9c](https://github.com/gregorio-project/gregorio/pull/1759/commits/71ffc9c440357f2ea57fb81580d8ca2cd5be209a). Thanks!

davidweichiang:
I think this just needs an entry in CHANGELOG.md and then it looks good to me.

rpspringuel:
It looks to me like it should be possible to generate a test file that follows your test plan and add it to the test repository.  @lbssousa, can you please do that?

davidweichiang:
@lbssousa Did you mean to close the other one (#1763) instead? I thought this one was still needed.

lbssousa:
I'm reviewing the code for this PR. I thought about opening a new one, but on second thought, I decided to just force push to this one instead.

lbssousa:
Just smoothed out the rough edges here. Could you please review it once again?

davidweichiang:
What changed from before? (I can't tell because of the force-push.)

I haven't tested, but it looks reasonable to me.

lbssousa:
> What changed from before? (I can't tell because of the force-push.)
> 
> I haven't tested, but it looks reasonable to me.

Just removed a local variable in favour of inline if-condition in `gre_exec`, and added better comments for auxiliary functions `gre_shellquote` and `gre_build_cmdstr`. This PR core remains unchanged.

I've also refactored the corredponding test https://github.com/gregorio-project/gregorio-test/pull/426 from scripted to tex-output, as suggested.

davidweichiang:
I've run the tests on a Mac and they pass. @rpspringuel, are you okay with me merging?

rpspringuel:
Go ahead.  I keep getting interrupted every time I go to look at this.  The result is that while my test repository is indicating test failures, I'm also not sure when I ran those tests and against what build they were run. If you've got a clean test sheet, then I'm willing to accept that and not try to resolve whatever confusing muddle I've currently got on my machine.

davidweichiang:
Most likely, it's because https://github.com/gregorio-project/gregorio-test/pull/426 has not yet be merged with develop. I had to do that in order to make the tests pass.

PR #1763, opened 2026-06-27
fix: crash in post_linebreak when lastglyph is nil (issue #1762)

### LINKED ISSUE #1762: `post_linebreak()` crashes with "nil lastglyph" when a mid-word `<v>…\hbox{…}…</v>` syllable falls at a line break
## Summary

GregorioTeX crashes with:

```
gregoriotex.lua:1101: attempt to index a nil value (local 'lastglyph')
```

when a GABC score uses a `<v>…</v>` verbatim element whose LaTeX content wraps its
output in a nested box (e.g. `\hbox{…}`, or `\stackon{…}{…}` from the `stackengine`
package), **and** the typeset line breaks at that syllable.

## Environment

| Software | Version |
|---|---|
| gregoriotex | 6.2.0 |
| LuaHBTeX | 1.24.0 (TeX Live 2026) |

## Minimal Reproducible Example

Two files are needed.

### `score.gabc`

```
name:MWE gregoriotex nil lastglyph;
%%
(c3)
a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)
<v>\hbox{so-}</v>(g)<v>\hbox{le-}</v>(g)<v>\hbox{ni-}</v>(g)<v>\hbox{da-}</v>(g)<v>\hbox{de}</v>(g)
a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g)a(g),(g) (::)
```

The five consecutive `<v>\hbox{…}</v>` syllables form a single word
(`so-le-ni-da-de`).  The four with a trailing hyphen inside the box are
mid-word (`end_of_word = 0` in the generated `.gtex`).  The narrow `\textwidth`
below forces a line break inside this word.

### `document.tex`

```latex
\documentclass[12pt]{article}
\usepackage{gregoriotex}
\setlength{\textwidth}{50mm}
\setlength{\oddsidemargin}{0pt}
\begin{document}
\gregorioscore[n]{score.gtex}
\end{document}
```

### Steps to reproduce

```
gregorio score.gabc
lualatex document.tex
```

No special font or third-party package is required.

### Error output

```
! LuaTeX error .../gregoriotex/gregoriotex.lua:1101:
   attempt to index a nil value (local 'lastglyph')
stack traceback:
    .../gregoriotex/gregoriotex.lua:1101: in local 'post_linebreak'
    ...
!  ==> Fatal error occurred, no output PDF file produced!
```

## Root Cause

`post_linebreak()` is defined at line 987 of `gregoriotex.lua`.  It loops over
every typeset line and, when the last syllable on a line is mid-word
(`potentialdashvalue = 1`), inserts a hyphen after it.  The relevant section
(lines 1094–1106) is:

```lua
if adddash then
  local lastglyph
  for g in node.traverse_id(glyph, lastseennode.head) do
    lastglyph = g          -- (1) find the last glyph to copy its font
  end
  if not (lastglyph.char == hyphen or lastglyph.char == 45) then  -- (2) line 1101
    local dashnode, hyphnode = getdashnnode()
    hyphnode.font = lastglyph.font
    insert_after(lastseennode.head, lastglyph, dashnode)
  end
end
```

`lastseennode` is the outer syllable-text hbox created by
`gregoriotex-syllable.tex:1254`:

```tex
\hbox attr \gre@attrid@part=4 {\unhcopy\gre@box@syllabletext}
```

When the GABC syllable text is ordinary characters, `lastseennode.head`
contains individual **glyph nodes** at the top level and `traverse_id(glyph, …)`
finds them normally.

However, when the GABC syllable text is a LaTeX command that produces a
**nested box** — such as `\hbox{so-}` or `\stackon[0pt]{so-}{ve-}` — executing
that command inside `\gre@box@syllabletext` places an inner **hlist node** as a
direct child of the outer hbox, with no glyph nodes at the top level.
`node.traverse_id` does **not recurse into sub-lists**, so the loop at (1)
finds nothing, `lastglyph` stays `nil`, and the dereference at (2) crashes.

The variable `lastseennode` is declared at function scope (line 990) and is
**not reset between lines**.  This means `lastseennode` can refer to a
nested-box syllable from one line when `adddash` becomes true on a later line,
compounding the risk.

## Why Forced Line Breaks (`(z)`, `(Z)`) Do Not Trigger the Crash

GABC provides two forced line break notations in the note field: `(z)` (justified
break, generates `\GreNewLine`) and `(Z)` (ragged break, generates `\GreNewParLine`).
One might expect that placing such a break immediately before a `<v>\hbox{…}</v>`
mid-word syllable would make the crash deterministic and easy to reproduce:

```
name:test;
%%
(c3)
a(g)…a(g)
<v>\hbox{so-}</v>(g)<v>\hbox{le-}</v>(g)<v>\hbox{ni-}</v>(z)<v>\hbox{da-}</v>(g)<v>\hbox{de}</v>(g)
a(g)…(::)
```

**This does not crash.**  Forced breaks are immune to the bug because they
activate a completely different code path in the TeX layer, which handles the
hyphen inline — before `post_linebreak()` ever runs.

### How forced breaks avoid the bug

When `(z)` is the note group of a syllable (here `\hbox{ni-}`), the gregorio
compiler inserts `\GreLastOfLine` inside the `\GreSetNextSyllable` argument of
the *preceding* syllable (`\hbox{le-}`):

```tex
\GreSyllable{…\GreSetThisSyllable{}{\hbox{le-}}…}{}{\GreUnstyled}{0}{
  \GreSetNextSyllable{}{\hbox{ni-}}{}{\hbox{ni-}}{}\GreLastOfLine
```

`\GreLastOfLine` sets the counter `\gre@lastoflinecount` to 1.  Later, in
`\gre@syllable@end` (`gregoriotex-syllable.tex`, lines 1170–1181), the end-of-syllable
processing checks this counter and, when the spacing between the current syllable
and the next would require a hyphen, inserts `\GreHyph` **directly into the
syllable text box** at TeX processing time:

```tex
\ifgre@showhyphenafterthissyllable
  \ifnum\gre@lastoflinecount=1\relax
    \setbox\gre@box@syllabletext=\hbox{%
      …\gre@emit@syllabletext{…\GreHyph…}%   ← hyphen placed inline here
    }%
```

Because the hyphen is inserted inline by this TeX path, the attribute
`\gre@attr@dash` is never set to `potentialdashvalue` (1); it stays at
`nopotentialdashvalue` (2).  When `post_linebreak()` iterates over the forced-break
line it finds `adddash = false` and skips the dash-insertion block entirely.
The crashing code at line 1101 is never reached.

### Why `(::z)` also avoids the bug

The `(::z)` notation (double barline followed by a forced break) has an
additional blocking mechanism: the barline promotes the preceding syllable to
`end_of_word = 1`.  In `post_linebreak()`, an `end_of_word = 1` node sets
`adddash = false`, so the code never enters the hyphen-insertion block regardless
of the syllable text structure.

### The fundamental asymmetry

The crash is therefore exclusive to **natural TeX line breaks**: situations where
TeX's own line-breaking algorithm — operating on the fully built paragraph, after
all TeX-level processing is complete — decides to break at a mid-word
nested-box syllable.  In these cases `\gre@lastoflinecount` is 0, no inline
hyphen is inserted, `\gre@attr@dash = 1` (potentialdashvalue), and
`post_linebreak()` is the sole mechanism responsible for the hyphen.  It is that
Lua-level mechanism that lacks the nil guard for nested-box syllables.

In other words: gregoriotex handles forced breaks correctly via the TeX-layer
`\gre@lastoflinecount` mechanism, but has no equivalent protection in the
Lua-layer `post_linebreak()` fallback that natural breaks rely on.

### Minimal fix — nil guard (safe, conservative)

Add a nil check before the dereference on line 1101:

```lua
-- BEFORE (line 1101):
if not (lastglyph.char == hyphen or lastglyph.char == 45) then

-- AFTER:
if lastglyph and not (lastglyph.char == hyphen or lastglyph.char == 45) then
```

When `lastglyph` is nil the syllable text contains no top-level glyphs
(only sub-boxes), meaning the box content already renders its own visible text
(possibly including a hyphen character typed inside it).  Skipping the extra
inserted hyphen is therefore semantically correct, not just a crash workaround.

### Complete fix — recursive glyph search

To also handle the font correctly when the glyph is inside a sub-box, replace
the `traverse_id` loop with a recursive helper:

```lua
local function find_last_glyph(head)
  local result = nil
  for n in node.traverse(head) do
    if n.id == glyph then
      result = n
    elseif n.head then
      local inner = find_last_glyph(n.head)
      if inner then result = inner end
    end
  end
  return result
end
```

Then replace lines 1095–1099:

```lua
-- BEFORE:
local lastglyph
for g in node.traverse_id(glyph, lastseennode.head) do
  lastglyph = g
end

-- AFTER:
local lastglyph = find_last_glyph(lastseennode.head)
```

This version finds the correct font even when the text is inside a nested box,
which allows an explicit hyphen to be inserted when one is semantically needed.

## Real-world Context

This crash was found in a LuaLaTeX liturgical book that uses
`\stackon[0pt]{#1}{#2}` (from the `stackengine` package) to overlay two
chant-tone text variants inside `<v>…</v>` elements:

```gabc
e(g) na(g)
<v>\stackon[0pt]{so-}{ve-}</v>(g)<v>\stackon[0pt]{le-}{ne-}</v>(g)
<v>\stackon[0pt]{ni-}{ra-}</v>(g)<v>\stackon[0pt]{da-}{ção\hphantom{-}}</v>(g)
<v>\stackon[0pt]{de}{}</v>(g)
```

`\stackon` creates a `\vtop` with two stacked `\hbox`es, which is a nested-box
structure.  With a `memoir` document class, the `ebgaramond` font, and a
190 × 275 mm page, the line-breaking placed one of these syllables at a line
end, triggering the crash.  The MWE above isolates the condition to plain
`\hbox{…}`, confirming the root cause is in `post_linebreak()` and not
in `stackengine` or any other package.

lbssousa:
## Summary

- `node.traverse_id` does not recurse into sub-lists, so when the text of a syllable is inside a nested box (e.g. `\stackon{…}{…}` within `<v>…</v>` in GABC), the loop that searches for the last glyph ends with `lastglyph = nil`, and the following line causes a crash by dereferencing `nil`.
- Introduces `find_last_glyph`, a recursive helper that walks the entire node tree (descending into `hlist`, `vlist`, etc.) to find the last glyph, even when it is inside nested hboxes.
- Replaces the flat loop `for g in node.traverse_id(glyph, lastseennode.head)` with a call to `find_last_glyph`.
- Adds a `nil` guard in the hyphen insertion condition, for the (rare) case in which no glyph is found at any level of the tree.

Closes #1762.

## Test plan

- [ ] Compile a GABC document containing `<v>…\stackon{…}{…}…</v>` in a mid-word syllable, with a natural line break (without `(z)`/`(Z)`), and verify that the crash no longer occurs.
- [ ] Verify that documents without nested boxes continue to produce hyphens correctly at line breaks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/). 
 
 ## [Unreleased][develop]
 ### Fixed
+- Fixed a crash in `post_linebreak` when a syllable's text is produced by a nested hbox (e.g. `\stackon{…}{…}` inside `<v>…</v>` in GABC), which left `lastglyph` as `nil` and caused a nil-dereference on natural line breaks. See [#1762](https://github.com/gregorio-project/gregorio/issues/1762).
 - Fixed a bug that could cause problems with headers/footers by generalizing the fix for fancyhdr (see [#1610](https://github.com/gregorio-project/gregorio/pull/1610) to work with any headers/footers. In particular, it fixes a similar problem with the `memoir` class. See [#1753](https://github.com/gregorio-project/gregorio/pull/1753).
 - Fixed some typos that would cause a handful of spaces to be scaled incorrectly. See [PR #1746](https://github.com/gregorio-project/gregorio/pull/1746).
 - Fixed a few bugs related to horizontal spacing around bars and clef changes. See issues [#1191](https://github.com/gregorio-project/gregorio/issues/1191), case 3 of [#1724](https://github.com/gregorio-project/gregorio/issues/1724), [PR #1743](https://github.com/gregorio-project/gregorio/pull/1743), and [#1745](https://github.com/gregorio-project/gregorio/issues/1745).
--- a/tex/gregoriotex.lua
+++ b/tex/gregoriotex.lua
@@ -951,6 +951,22 @@ local function adjust_additional_spaces(line, info, linenum)
   end
 end
 
+-- node.traverse_id does not recurse into sub-lists; this helper walks the
+-- full node tree so that the last glyph is found even inside nested hboxes
+-- produced by <v>…\hbox{…}…</v> or <v>…\stackon{…}{…}…</v> in GABC.
+local function find_last_glyph(head)
+  local result = nil
+  for n in node.traverse(head) do
+    if n.id == glyph then
+      result = n
+    elseif n.head then
+      local inner = find_last_glyph(n.head)
+      if inner then result = inner end
+    end
+  end
+  return result
+end
+
 local function post_linebreak(h, groupcode, glyphes)
   --dump_nodes(h)
   -- TODO: to be changed according to the font
@@ -1059,13 +1075,10 @@ local function post_linebreak(h, groupcode, glyphes)
 
     -- If the last syllable needed a dash, add it
     if adddash then
-      local lastglyph
-      -- we traverse the list, to detect the font to use,
-      -- and also not to add an hyphen if there is already one
-      for g in node.traverse_id(glyph, lastseennode.head) do
-        lastglyph = g
-      end
-      if not (lastglyph.char == hyphen or lastglyph.char == 45) then
+      -- use find_last_glyph so nested hboxes (e.g. from \stackon) are searched
+      local lastglyph = find_last_glyph(lastseennode.head)
+      -- guard nil: if no glyph is found at any level, skip hyphen insertion
+      if lastglyph and not (lastglyph.char == hyphen or lastglyph.char == 45) then
         local dashnode, hyphnode = getdashnnode()
         hyphnode.font = lastglyph.font
         insert_after(lastseennode.head, lastglyph, dashnode)


rpspringuel:
Please submit in English.

lbssousa:
Sorry, text translated to English.

rpspringuel:
Again, please submit a new test in the test repository which tests this fix.

davidweichiang:
I haven't tested this, but the code looks reasonable to me.

davidweichiang:
@lbssousa Sorry that after merging #1720 the conflicts may be tricky to resolve.

davidweichiang:
Actually, could you first check whether #1720 already fixed the issue? The old logic was to add a hyphen that has the same font as the last glyph, but the new logic is just to use the same font as the whole last syllable. So I don't think there is any longer any code that searches for the last glyph.

lbssousa:
Not needed anymore, since PR #1720 already fixes issue #1762.

Issue #1724, opened 2026-03-03
Spacing around clef changes has some inaccuracies


davidweichiang:
These are more issues that don't necessarily need to be fixed individually.

1. When a clef change occurs without a preceding bar (as happens in the tests, but maybe not in real scores), gregorio generates an alignment/alteration type of `{0}{0}` (punctum with no alteration) for it, which is not right.

2. Ordinarily `\GreBarSyllable` is not preceded by a `syllablefinalskip`, but a clef change without bar is a `\GreBarSyllable` and is preceded by a `syllablefinalskip` (because of the wrong alignment type mentioned above).

3. A clef change uses two syllables, one for when there's a line break and one for when there's no line break. Both should have the same `previousenddifference`, which is the `enddifference` of the syllable before the clef change. But in fact the `previousenddifference` of the no-break case is the `enddifference` of the at-break case.

davidweichiang:
Case (3) is #1191.

davidweichiang:
I think case (2) is a little bit weird, but not a bug. Closing.

Issue #1191, opened 2016-07-21
clef change strange spacings


eroux:
In the following minimal example:

```tex
\documentclass[12pt]{article}
\usepackage{libertine}
\usepackage[autocompile]{gregoriotex}
\begin{document}
\gabcsnippet{(c3) {u}uuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu(g) (z0;c3) ut(h)}
\end{document}
```

the spacing is completely odd, the custo starts at the end of the text, while the spacing algorithm should be similar to the new bar spacing algorithm

PR #1764, opened 2026-07-10
Bar spacing in Lua

### LINKED ISSUE #1723: nextbegindifference should equal begindifference of next syllable, but sometimes doesn't
To compute the space between the current syllable and the next, the syllable spacing code has to predict the `begindifference` of the next syllable (how much the notes extend to the left of the text). There are a number of cases where this prediction is not correct, which leads to the syllable spacing being slightly off. In extreme cases, this can lead to collisions.

I don't think these are worth fixing, since I'm hoping to move the syllable spacing code into Lua. But I'm documenting them here because they will be test failures.

1. When the next syllable is just an alteration with no note, e.g., `(gx)`. Gregorio emits an alignment and alteration type of `{0}{0}` (punctum and no alteration) but it should probably be `{18}{1}` (no note and flat).
2. When the next syllable has NABC but no note, e.g., `(|pu)`. Gregorio emits `{0}{0}` (punctum and no alteration) but it should probably be `{18}{0}` (no note and no alteration)
3. When the next syllable has a soft alteration, e.g., `(gXg)`. There are missing cases to check for an alteration type of `7` (soft flat) so it is treated as no alteration.

### LINKED ISSUE #1734: Incorrect nextbegindifference for \GreBarSyllables
1. If a word is interrupted by a bar syllable, it seems that the bar and its surrounding space are not counted when deciding whether to hyphenate the word. I think the first example should have a hyphen, but doesn't:

```
\gresetinitiallines{0}
\gabcsnippet{a(g)(,)a(g)}
\gabcsnippet{a(gg)(,)a(g)}
```

Output (with barspacing lines turned on):
Image

### LINKED ISSUE #1744: Break at vs. before clef change
A clef change appears differently depending on whether the line break occurs at the clef change itself or before it:
```
\documentclass{article}
\begin{document}
Clef change occurring at line break:
\gabcsnippet{(c3) aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(g) a(g) (f3) a(g)}
Adding one more `a' should not change the second line clef:
\gabcsnippet{(c3) aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(g) a(g) (f3) a(g)}
\end{document}
```
Image

I think only the first example is right.

davidweichiang:
This follow-up to #1720 moves both the old and new bar spacing algorithms to Lua, as well as a few other computations. The result is that much of the redundant information in .gtex files can be eliminated.

Remaining to-do items:
- [x] Quite a few tests are changed, which I will catalogue once the dust is settled on #1720.
- [x] Mid-score changes in bar spacing settings are not yet respected.


davidweichiang:
There are currently 37 tests failing.

## Consecutive bar syllables

This is the biggest category.  If there are two bar syllables in a row, the first syllable's nextbegindifference did not equal the second syllable's begindifference (issue #1734, case 2).

However, it's not really clear what the right behavior is. A bar syllable depends on both the previous and next syllable, so when there are two bar syllables in a row, what should be done? The intent of the old code seemed to be to initialize both bar syllables so that the bar is aligned above the text, then adjust the first bar syllable, then adjust the second bar syllable. That's what the new code does.

Better in my opinion
- gabc-output/glyphs/divisio_finalis.gabc
- gabc-output/glyphs/bars.gabc
- gabc-output/bar-substitution.gabc
- tex-output/bugs/fix-1110/fix-1110.tex

Don't know if better or worse
- gabc-output/glyphs/high-bars-2-lines.gabc
- gabc-output/glyphs/high-bars-3-lines.gabc
- gabc-output/glyphs/high-bars.gabc
- gabc-output/glyphs/high-bars-5-lines.gabc
- gabc-output/empty-syllables.gabc
- gabc-output/lines/2-lines.gabc
- gabc-output/lines/3-lines.gabc
- gabc-output/lines/5-lines.gabc
- tex-output/bugs/fix-1205/fix-1205.tex
- tex-output/bugs/fix-1416/fix'1416.tex

## Punctum mora adjustment

In my opinion these are all better.

New code uses actual width of punctum mora and space before it
- tex-output/bugs/fix-396/fix-396.tex

When punctum mora occurs before a clef change, and there is no break, the old code didn't use correct punctum mora adjustment.
- gabc-output/bugs/fix-1192.gabc

A syllable that has nabc that extends beyond a punctum mora doesn't need a punctum mora adjustment.
- gabc-output/FactusEst.gabc: ve-
- gabc-output/veni.gabc

## Bars with no text (added after 9d45861)

When a bar syllable has no text, there's an empty text box whose position sometimes affects spacing before and after the syllable. Now the text box is centered directly under the bar.

In my opinion better:
- gabc-output/bugs/fix-928-unison-no-break.gabc
- gabc-output/glyphs/flexus-stroke.gabc

## Other features

In my opinion these are all better.

Syllable clearing for bar syllables is more accurate
- gabc-output/glyphs/clear.gabc

Old code computed eolshift using the width of a custos, even when the custos was disabled or flatted.
- fix-1146.tex

Old code adjusted for alteration in the next syllable, even if it's on the next line
- gabc-output/bugs/fix-243.gabc
- gabc-output/bugs/fix-243-old.gabc

In clef changes, old calculation uses wrong nextbegindifference (issue #1724 case 1 and #1191)
- gabc-output/flatted-custos-clef-change.gabc: alteration adjustment is not supposed to happen after clef, but previously it did
- gabc-output/glyphs/clef_change.gabc

## Tests that revealed new bugs

- tex-output/bugs/fix-1503/fix-1503.tex: issue #1744

## Miniscule or no difference, except in debugging marks

- gabc-output/edge_cases.gabc
- gabc-output/bugs/fix-1424.gabc
- tex-output/bar-spacing/last-line.tex
- tex-output/bar-spacing-old/bar-spacing-old.tex
- tex-output/hidelyricsnotes/text-only.tex
- gabc-output/glyphs/punctum-inclinatum-2.gabc
- tex-output/bar-spacing/lyric-centering.tex (new after commit 9d45861)

## Tests that need a closer look

- [x] fix-845.tex: The new result is clearly worse. My notes say that this is actually an existing bug, but I'm not so sure.
- [ ] tex-output/nabc-skip-alterations/nabc-skip-alterations.tex: I haven't paid very close attention to tests involving NABC, since NABC spacing is in flux. However, these changes are large.
- tex-output/nabc-skip-alterations/nabc-skip-alterations-two-voices.tex

rpspringuel:
Just doing a first pass review to categorize this by milestone.  If I've seen things correctly, this doesn't change any user facing code and so could be put into a 6.3.0 release.  @davidweichiang can you confirm that's the intention?

davidweichiang:
Correct, this shouldn’t have any user facing changes except very slight changes in spacing.

doc/Command_Index_internal.tex
```diff
@@ -595,24 +577,12 @@ \section{Gregorio\TeX{} Controls}
...
-\macroname{\textbackslash gre@endafterbar}{\#1}{gregoriotex-main.tex}
-Macro to call after ending a bar.
-
-\begin{argtable}
-  \#1 & \texttt{0} & We are at the end of a line\\
-  & \texttt{1} & We not at the end of a line\\
-\end{argtable}
-
-\macroname{\textbackslash gre@endofsyllable}{\#1\#2\#3}{gregoriotex-syllable.tex}
+\macroname{\textbackslash gre@endofsyllable}{\#1\}{gregoriotex-syllable.tex}
```
rpspringuel: There's an extra backslash in this line.  It should be `{\#1}` in the middle.

davidweichiang: Sorry, fixed.

rpspringuel:
## Changes I'm willing to accept:
gabc-output/glyphs/bars.gabc
gabc-output/glyphs/high-bars-3-lines.gabc
gabc-output/glyphs/high-bars-2-lines.gabc
gabc-output/glyphs/high-bars-5-lines.gabc
gabc-output/glyphs/high-bars.gabc
gabc-output/glyphs/divisio_finalis.gabc
gabc-output/lines/2-lines.gabc
gabc-output/lines/3-lines.gabc
gabc-output/lines/5-lines.gabc
tex-output/bugs/fix-1110/fix-1110.tex
tex-output/bugs/fix-1205/fix-1205.tex
tex-output/bugs/fix-1416/fix'1416.tex
tex-output/nabc-skip-alterations/nabc-skip-alterations-two-voices.tex
tex-output/bar-spacing/bar-spacing-new.tex

## Changes that look problematic to me
### gabc-output/empty-syllables.gabc
I don't like how close the 3 gets to the 2 in this one.  They seem to combine into a single word when they are clearly separate words in the source.  The 10/11 space works better.

## Other notes
### gabc-output/bugs/fix-928-unison-no-break.gabc
This is definitely better, but there are still custos (and a barline) outside of the staff, so this test is still revealing a bug that needs fixing.  Do we have an issue for that?

### gabc-output/glyphs/bars.gabc<br> gabc-output/bar-substitution.gabc
It looks to me like the biggest change is at the beginning of the line and the rest of the changes are mostly knock-on effects from that until things start matching near the end (so long as changes in the line-breaking don't happen.

davidweichiang:
> ### gabc-output/empty-syllables.gabc
> I don't like how close the 3 gets to the 2 in this one.

How about 6-7? I'll dig into this a little bit. I thought there should be interwordspacetext and it doesn't look like there is.

davidweichiang:
A brief update on empty-syllables.gabc:

(a) As noted above, there are some oddities in how spacing between two consecutive bar syllables is computed (including issue #1734 case 2, but maybe others). In the Lua implementation, I didn't try to replicate these.

(b) Syllables ordinarily have a space after (syllablefinalskip), which is often negative. In the new bar spacing algorithm, a bar overrides the previous syllable's syllablefinalskip. However, there are some cases where the previous syllablefinalskip is retained, but the bar spacing is still computed as if the syllablefinalskip were not there. This includes when the previous syllable is a bar. In the Lua implementation, I did try to replicate these.

In the case of empty-syllables.gabc, these two choices combine in a bad way, and I agree that it must be fixed. Since both (a) and (b) both look like bugs to me, I would rather fix it by reducing the cases where (b) occurs than by replicating (a). If you (@rpspringuel) agree, I'll look into how to do that.

davidweichiang:
The above commit suppresses (b) in the case of two consecutive bar syllables. empty-syllables.gabc now looks like this, which is unexpectedly even:

image

No new tests fail, but it's possible some of the failing tests got better or worse.

davidweichiang:
> ### gabc-output/bugs/fix-928-unison-no-break.gabc
> This is definitely better, but there are still custos (and a barline) outside of the staff, so this test is still revealing a bug that needs fixing. Do we have an issue for that?

I think these are just overfull hboxes, right?

davidweichiang:
I went ahead and accepted empty-syllables.gabc as well as some of the ones that we agree are ok. I believe you're okay with the others, too, but will leave them for final discussion.

gabc-output/glyphs/high-bars-2-lines.gabc: consecutive bar syllables are now closer together, but not too close
gabc-output/glyphs/high-bars-3-lines.gabc: same
gabc-output/glyphs/high-bars.gabc: same
gabc-output/glyphs/high-bars-5-lines.gabc: same

gabc-output/glyphs/bars.gabc: The comma under the first quarter bar looks worse to me. The 2nd line and following lines start with a bar (which I didn't realize was possible) and the bar sits considerably closer to the clef. It used to be that the 2nd line's first bar was almost (but not quite) aligned with the 1st line's first bar, but now they look very different.
gabc-output/bar-substitution.gabc: same

tex-output/bugs/fix-1503/fix-1503.tex: clef change sits closer to the clef

tex-output/nabc-skip-alterations/nabc-skip-alterations.tex: note with nabc sits farther from bar or clef change, better?
tex-output/nabc-skip-alterations/nabc-skip-alterations-two-voices.tex: same

davidweichiang:
For the test `gabc-output/glyphs/bars.gabc`, the first difference ("The comma under the first quarter bar looks worse to me") is fixed by the above commit. It seems to reduce some differences in other tests as well.

The second difference ("the bar sits considerably closer to the clef") is because of a change in how no-text bar syllables are handled. In both implementations, there is a zero-width text box. In the old TeX implementation, this box was placed like any other text box (midway between the neighboring text boxes). In the new Lua implementation, this box is placed directly under the bar (where it will not impact neighboring syllables). If I recall correctly, there was a case where the old placement was creating needless space and the new placement was better. (If you want, I can try to remember what case that was.) This difference is another such example; the old implementation shifts the zero-width text box to the left, creating extra space at the beginning of the line. I believe the new way makes more sense and propose not to change it.

davidweichiang:
I went ahead and accepted the changed results of the `high-bars` tests, which now have pretty minor differences.

rpspringuel:
I'm a bit busy with the start of School this week and next.  I'll get back to looking at this stuff in a couple of weeks.

PR #1765, opened 2026-07-30
Support stacked (multi-level) lyric lines in GABC syllables

### LINKED ISSUE #374: Multiple lines of text
This long-awaited feature will add the possibility for users to associate multiple lines of text (with all their syllables aligned) with one line of notes.

It would contain the following features:
- possibility to have syllables centered, left-aligned or right-aligned (different for each syllable)
- possibility to insert small horizontal lines between text lines
- as many text lines as user wants
- possibility of line numbering
- possibility of curly brace at the end of the lines
- possibility of vertical rules between syllables

example-lilypond

multiplelineswithrules

This is quite a lot of work...

lbssousa:
Closes #374.

Adds support for multiple stacked lines of lyrics under a single line of
notes, written by separating the lines with `|` inside a syllable's text:
```
(c4) <sp>V/</sp>. Sanc|Sanc|Sanc(h)|ta|ta(h)ta| De| Vir(g) Ma|i|go(h)rí| Gé| Vír(ixir1)|ni|gi(gr)a,|trix,|num,(g.) (:) <sp>R/</sp>. o(g)ra(f) pro(g) no(h)bis.(h) (::)
```

The text before the first `|` is the main lyric line. Each subsequent `|`
starts a new line further below the staff. This is meant for cases like litanies,
psalm tones, or any score where several verses/invocations share the same
melody and are traditionally printed stacked under one staff instead of
repeating the notes for each one.

This proposed syntax is heavily based on the one introduced in
[Source & Summit online GABC editor](https://www.sourceandsummit.com/editor/alpha/).

## Syntax rules

- Each line of the stack is centered under the notes independently,
  following the normal centering rules (including forced centering with
  `{}`, usable separately per line).
- Word boundaries and hyphenation are computed independently per line,
  following the same rule as the main line: a leading space starts a new
  word. Without one, two **consecutive** syllables' lines are the same word,
  even where the main line's word already ended. A word only ends where
  the following syllable's stack doesn't reach that deep at all (a plain
  syllable, or a shallower stack). A line that's present but
  deliberately left empty (`ca||ci(h)`, or a trailing empty line at a
  stack's end) does **not** end the word, it continues on the next syllable
  that carries text there.
- `|` is now reserved in syllable text; a literal pipe must be escaped as
  `$|`. GABC files written by Gregorio itself already escape `|`, so
  regenerated files are unaffected.


tex/gregoriotex-syllable.lua
```diff
@@ -33,12 +33,16 @@ local temp = node.id('temp')
...
+local hlist = node.id('hlist')
 
 local syllable_id_attr = luatexbase.attributes['gre@attr@syllable@id']
 
 local part_attr = luatexbase.attributes['gre@attr@part']
 local part_lyrics = 4
 local part_notes = 10
+-- additional lyric lines (stacked lyrics, level 2+) use
+-- part_lyric_line_base + level, so they sort after every fixed part
+local part_lyric_line_base = 9
```
davidweichiang: I thought that 9 was used in another PR, which is why I skipped 9 to use 10 for notes.

davidweichiang:
Are the 2nd and beyond lines of lyrics treated differently from the 1st line, with respect to syllable centering, space between syllables, etc.? If so, it makes sense that they are a separate "part" from the 1st line. But if not, why not generalize the existing lyrics to be multi-line?

lbssousa:
> Are the 2nd and beyond lines of lyrics treated differently from the 1st line, with respect to syllable centering, space between syllables, etc.? If so, it makes sense that they are a separate "part" from the 1st line. But if not, why not generalize the existing lyrics to be multi-line?

I worked on the second approach, removing most of the assymetries between the first lyrics line and the others. Some of them must remain for obvious reasons, e.g. only the first lyrics line is eligible for big initial:
image

tex/gregoriotex-syllable.lua
lbssousa: Changed in recent commits

PR #1767, opened 2026-08-03
Reserve vertical spacing for rare signs and below-notes NABC lines for collision avoidance

### LINKED ISSUE #1766: ALT-text/commentary/NABC colliding with note accents r1-r8 and pitch extremes
Gregorio version: 6.2.0-develop (built from `develop` at commit 7c806e08)

# Summary

The extra vertical space GregorioTeX reserves above/below the staff (to avoid collisions with above-lines-text, gregorian commentary, the NABC lines, and the lyrics) is driven entirely by each glyph's highest/lowest *note pitch* (`compute_height_extrema()` in `gregoriotex-write.c`, propagated to Lua via `\GreGlyphHeights`, consumed by `adjust_additional_spaces()` in `gregoriotex.lua`). Two glyphs that are drawn beyond the note's own pitch aren't accounted for by that computation, so they can collide with neighboring elements.

## `r1`-`r8`

accentus, accentus reversus, circulus, semicirculus, semicirculus reversus, musica ficta flat/natural/sharp can collide with above-lines-text/commentary/the upper NABC line

These are always rendered above the note (at least at pitch `highest_pitch - 2`, see `\gre@vepisemaorrare` in `gregoriotex-signs.tex`), but `compute_height_extrema()` never looks at `note->special_sign`, so this extra height never reaches the space-reservation logic.

MWE (dimensions reduced only to make the collision visible without a huge score; the underlying gap exists at any setting):
```latex
\documentclass{article}
\usepackage{gregoriotex}
\grechangedim{abovelinestextheight}{0cm}{scalable}
\grechangedim{abovelinestextraise}{0cm}{scalable}
\begin{document}
\gabcsnippet{<alt>al</alt>ta(gr1)}
\end{document}
```
Screenshot: 

Image

## Low notes (pitch a/b) can collide with the lower NABC line

```
name: low-pitch-blnabc-spacing;
nabc-lines: 2;
%%
(c4)
gra(a||ta)
```
Screenshot (the `ta` neume sits immediately against the low note, while the lyrics below have a generous gap):

Image

`blnabc_lower` in `adjust_additional_spaces()` (gregoriotex.lua) uses a fixed dimension
(`belowlinesnabcheight`) and never grows with `additional_bottom_space`, unlike the lyrics position,
which already does grow with it. So as a note gets lower, the note-to-blnabc gap shrinks while the
blnabc-to-lyrics gap grows.

lbssousa:
Fixes #1766.


src/gregoriotex/gregoriotex-write.c
```diff
@@ -2990,6 +2990,20 @@ static void compute_height_extrema(const gregorio_glyph *const glyph,
                 ++height;
             }
         }
+
+        if (note->special_sign) {
+            /* accentus, circulus, semicirculus and musica ficta always
+             * render above the note, at the same height
+             * gre@vepisemaorrare draws them at */
+            signed char sign_height = (signed char)(note->u.note.pitch + 1);
+            if (score->highest_pitch - 2 > sign_height) {
```
davidweichiang: Why does this need to refer to `score` instead of just using the `note` pitch, as is done for episemas?

lbssousa: Good question! It's not about this note's pitch, it's that rare signs have a floor episemas don't: they never sit below the top staff line even on a low note.

`score->highest_pitch` is just `LOWEST_PITCH + 4 + 2*staff_lines`, so `highest_pitch - 2 == \gre@pitch@raresign` for every `staff-lines` setting. That's the score dependency.

I've just pushed commit 4a2d41ed805618ebcf9dc899ef3051beec18fc97 naming it `score->above_staff_pitch` instead of the bare -2, for clarity.

davidweichiang:
Since the height of a rare sign is computed in TeX, wouldn't it be better to compute the extremum in TeX as well?

https://github.com/gregorio-project/gregorio/blob/5981f89b752acd09c6e91ebf47c36276d6976056/tex/gregoriotex-signs.tex#L1372-L1380

Namely, inserting a call to `\GreGlyphHeights` here ought to do it.

UPGRADE.md
```diff
@@ -8,6 +8,10 @@ This file contains instructions to upgrade to a new release of Gregorio.  These
...
 
+`additionaltopspacenabcthreshold` default changed from `4` to `0`: above-lines nabc used to only grow taller for unusually tall content, leaving a gap right where a rare sign or a plain high note near the top of the staff would touch the nabc line without any extra space being reserved. Above-lines nabc now reacts as soon as content reaches that height, closing the gap. This slightly increases vertical spacing on lines where content reaches close to the top of the staff and nabc is present. If you'd rather keep the old behavior, use `\grechangecount{additionaltopspacenabcthreshold}{4}`.
+
+`abovelinesnabcraise` default changed from `0.17351cm` to `0.25cm`: even with the threshold fix above, the accentus glyph's tip extends a bit higher than the pitch-step reservation assumes, so above-lines nabc could still sit close enough to touch it. This adds a small, fixed amount of headroom to close that gap. This affects every line with above-lines nabc, not just ones with a rare sign, though the shift is small (about 0.03cm/0.8pt more than before). If you'd rather keep the old, tighter spacing, use `\grechangedim{abovelinesnabcraise}{0.17351cm}{scalable}`.
```
davidweichiang: I think the mention of "the accentus glyph" is going to be confusing for most users, since this is a rare use-case.

tex/gregoriotex-gsp-default.tex
```diff
@@ -362,7 +362,7 @@
 % same, for notes taken into account for alt text vertical position
 \grechangecount{additionaltopspacealtthreshold}{0}%
 % same, for notes taken into account for nabc vertical position
-\grechangecount{additionaltopspacenabcthreshold}{4}%
+\grechangecount{additionaltopspacenabcthreshold}{0}%
```
davidweichiang: I think this change makes a lot of sense, but in #1659 I didn't change it because I didn't want to break existing scores. @rpspringuel are you okay with changing the default?

tex/gregoriotex-gsp-default.tex
```diff
@@ -385,7 +385,7 @@
 % height that is added at the top of the lines if there is text above the lines (it must be bigger than the text for it to be taken into consideration)
 \gre@createdim{dimen}{abovelinestextheight}{0.3 cm}{scalable}%
 % distance of the above-lines nabc above the top of the staff
-\gre@createdim{dimen}{abovelinesnabcraise}{0.17351 cm}{scalable}%
+\gre@createdim{dimen}{abovelinesnabcraise}{0.25 cm}{scalable}%
```
davidweichiang: I think this used to be the same as abovelinestextraise, so we gave it the same default. @rpspringuel are you okay with changing the default?

lbssousa:
> Since the height of a rare sign is computed in TeX, wouldn't it be better to compute the extremum in TeX as well?
> 
> https://github.com/gregorio-project/gregorio/blob/5981f89b752acd09c6e91ebf47c36276d6976056/tex/gregoriotex-signs.tex#L1372-L1380
> 
> Namely, inserting a call to `\GreGlyphHeights` here ought to do it.

Agreed in principle, and it'd kill exactly the kind of duplication that let this bug hide for years. The catch here is that `\GreGlyphHeights` overwrites, it doesn't take a max/min.

`write_signs` calls it once per glyph with the max/min already folded in over every note+sign. `\gre@vepisemaorrare` fires per-note inside that same loop, so a bare call there would just clobber the glyph-wide value with whatever sign rendered last (a tall note followed by a shorter signed one would quietly lose its height).

Maybe we can implement a max/min-accumulating variant, but it would be a real refactor, not a one-liner. Would it be better to postpone this to another PR?

davidweichiang:
> The catch here is that \GreGlyphHeights overwrites, it doesn't take a max/min.

I don't think that's right. It sets the attributes `\gre@attr@glyph@top` and `\gre@attr@glyph@bottom`, so if there's both a note and a rare sign, they'd each get their own attributes. The Lua code that reads these attributes takes the max/min over all the boxes, if I'm not mistaken.

lbssousa:
> I don't think that's right. It sets the attributes `\gre@attr@glyph@top` and `\gre@attr@glyph@bottom`, so if there's both a note and a rare sign, they'd each get their own attributes. The Lua code that reads these attributes takes the max/min over all the boxes, if I'm not mistaken.

Sorry, my bad. I'll take a closer look and return with a possible refactor tomorrow.

src/gregoriotex/gregoriotex-write.c
```diff
@@ -2969,8 +2969,8 @@ static __inline bool _found(FILE *const f, const bool found)
 }
 
 static void compute_height_extrema(const gregorio_glyph *const glyph,
-        const gregorio_note *note, signed char *const top_height,
-        signed char *const bottom_height)
+        const gregorio_note *note,
```
davidweichiang: Since there are now only formatting changes on this file, could you revert these remaining changes?

lbssousa: Sure! Reverted in commit d6c6b6f8bef667ceb562fb5d021bc09cce602936.

tex/gregoriotex-gsp-default.tex
lbssousa: Maybe we can postpone this potential breaking change to another PR.

tex/gregoriotex-gsp-default.tex
lbssousa: If you prefer to deliberate about this later, we can move this change to another PR.

lbssousa:
Regarding all those `*raise` parameters whose defaults I'm cahnging in this PR, I'm trying to find another solution to make the necessary increment dinamically if rare signs are present, so we don't need to change the default values.

davidweichiang:
The trouble with raising the commentary based on high notes or signs is that the commentary is always on the right side, so we might not want high notes or signs on the left side of the staff to raise the commentary. I think it might be best to do it manually.

tex/gregoriotex-gsp-default.tex
```diff
@@ -328,7 +328,7 @@
 % Space between lines in the commentary
 \gre@createdim{dimen}{commentaryseparation}{0.05 cm}{scalable}%
 % Amount to raise (positive) or lower (negative) the commentary from the default position (base line of bottom commentary aligned with top line of staff)
-\gre@createdim{dimen}{commentaryraise}{0.17351 cm}{scalable}%
+\gre@createdim{dimen}{commentaryraise}{0.35 cm}{scalable}%
```
davidweichiang: Same comment as the other defaults changes; moreover, in this case, it's a common element (commentary) whose default is being changed for a rare one (rare signs).

davidweichiang:
There are currently five changes in this PR, and my opinions are:
- Rare signs affect height extrema: definitely
- belowlinesnabcthreshold: definitely
- Change abovelinesnabcthreshold: In favor
- Change abovelinesnabcraise: defer to judgement of @lbssousa or anyone else who uses nabc
- Change commentaryraise: not in favor

lbssousa:
Makes sense. I'm reverting the `commentaryraise` change in commit 63813796542c3c08dfb0654d556e29650992723e.

lbssousa:
I also gave up touching `abovelinesnabcraise` after visually checking the tests. Reverted in commit 8b9103cf3a4314ebe61d9ef5148dab8dd805283a.

Now only the following remain in this PR:
- rare signs affect height extrema
- `belowlinesnabcthreshold`
- `abovelinesnabcthreshold` default value change from 4 to 0

davidweichiang:
I'm seeing some test failures:

gabc-output/glyphs/nabc-edge-cases.gabc : FAIL - [gabc-output/glyphs/nabc-edge-cases.gabc.out/page-0.png] differ from expected
gabc-output/two-nabc.gabc : FAIL - [gabc-output/two-nabc.gabc.out/page-0.png] differ from expected
gabc-output/FactusEst.gabc : FAIL - [gabc-output/FactusEst.gabc.out/page-0.png] differ from expected
gabc-output/bugs/fix-1700.gabc : FAIL - [gabc-output/bugs/fix-1700.gabc.out/page-0.png] differ from expected
gabc-output/veni.gabc : FAIL - [gabc-output/veni.gabc.out/page-0.png] differ from expected
tex-output/nabc-alignment/nabc-alignment-two-voices.tex : FAIL - [tex-output/nabc-alignment/nabc-alignment-two-voices.tex.out/page-0.png] differ from expected
tex-output/nabc-alignment/nabc-alignment.tex : FAIL - [tex-output/nabc-alignment/nabc-alignment.tex.out/page-0.png] differ from expected

davidweichiang:
The test failures all look to me like they resulted from the change in the default abovelinesnabcthreshold. I'm unsure whether they're better or worse and/or whether the test cases need to be updated.

lbssousa:
> The test failures all look to me like they resulted from the change in the default abovelinesnabcthreshold. I'm unsure whether they're better or worse and/or whether the test cases need to be updated.

I've just updated these expected results in PR https://github.com/gregorio-project/gregorio-test/pull/430. The visual changes are perfectly acceptable IMHO.

lbssousa:
Now that this PR has gone through so many back-and-forths, should I squash the commits before merging when it's ready, to not pollute Gregorio's Git history?

davidweichiang:
I think that would be a good idea but maybe @rpspringuel can say whether
there’s a policy.

davidweichiang (approved):

lbssousa:
I've included a real-world example in one of my tests:
image

Afterwards, I've modified this example, forcing a lowering of the virga over the note with highest pitch, and noticed a still unresolved collision:
image

This seems an artificial modification, but it's relevant when we need to respect the "relative diastema" between neumes in a NABC line (specially for Laon neumes).

I'll keep working on a solution to avoid the collision in the second example above.

lbssousa:
image
Some progress here. I'm now able to raise the NABC line to avoid collisions even in edge cases like this one.

I also mirrored the solution for the below-lines NABC case, ensuring no collisions occur even in edge cases:
image

tex/gregoriotex-gsp-default.tex
rpspringuel: I'm okay with changing the default, but think that would be clearer as a separate PR that only changes default spacings/counts.  That way it'll be easier to see what changes are the result of the defaults changing, and which are due to the changes in how thresholds are calculated and used.

lbssousa: Just reverted `additionaltopspacenabcthreshold` default value change here, and moved it to the follow-up PR https://github.com/gregorio-project/gregorio/pull/1770.

tex/gregoriotex.lua
```diff
@@ -836,15 +861,22 @@ local function adjust_additional_spaces(line, info, linenum)
   local cur = 0 -- vertical position without any additional space
   local add = 0 -- with additional space
 
+  -- Extra room for nabc glyphs shifted by an explicit hX pitch code:
+  -- extra_bottom compensates a glyph pulled toward the staff (the collision
```
davidweichiang: The comment `(the collision fix)` may not make sense for a reader who is looking only at the code and not at the issue

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ As of v3.0.0 this project adheres to [Semantic Versioning](http://semver.org/).
...
 - Fixed a bug in horizontal spacing when the first syllable consists of only a bar. See [PR #1741](https://github.com/gregorio-project/gregorio/pull/1741).
 - Fixed a bug in horizontal spacing before a soft alteration. See [PR #1761](https://github.com/gregorio-project/gregorio/pull/1761).
+- Fixed vertical spacing so that rare signs (accentus, circulus, semicirculus, musica ficta) no longer collide with above-lines text, and so that below-lines nabc no longer collides with the low note it sits under; the new count `additionalbottomspacenabcthreshold` gives it the same reactive behavior that the lyrics already had, independently of the lyrics threshold. Also fixed two related gaps that could leave this space uncollapsed or unreserved: hiding staff lines alone (`\gresetlines{invisible}`) no longer collapses the pitch grid unless notes and the clef are hidden too (which was misplacing the notes, nabc, and above-lines text still drawn on it), and hiding notes while nabc stays visible (`\gresetnotes{invisible}`) no longer discards the pitch data — including rare signs — that this collision avoidance depends on. See [#1766](https://github.com/gregorio-project/gregorio/issues/1766).
```
davidweichiang: Should this be multiple items?

I don't know what "this space" refers to.

It seems to me there's an additional change (nabc_baseraise_max/min) which is not explained here.

tex/gregoriotex.lua
lbssousa: Clarified comments in commit 379298bb9c9f72f7d2015db94ff3864d3e9dd759. Thanks!

CHANGELOG.md
lbssousa: Clarified in commit 325eced1ef578c7cf73d160ae39f4de02ff34ae6. Thanks!

davidweichiang:
I'm seeing three failed tests:

```
gabc-output/two-nabc.gabc : FAIL - [gabc-output/two-nabc.gabc.out/page-0.png] differ from expected
gabc-output/FactusEst.gabc : FAIL - [gabc-output/FactusEst.gabc.out/page-0.png] differ from expected
gabc-output/veni.gabc : FAIL - [gabc-output/veni.gabc.out/page-0.png] differ from expected
```

They all involve increased vertical space with nabc, which is as expected. But at least for two-nabc.gabc, it seems like too much space to my eye.

lbssousa:
> They all involve increased vertical space with nabc, which is as expected. But at least for two-nabc.gabc, it seems like too much space to my eye.

I tried to find some room for reducing those extra spacing without breaking the anti-overlap logic. Take a look at the new results.

davidweichiang:
I think that looks better. One test still fails for me, tex-output/nabc-alignment/nabc-alignment.tex, with very small differences.

davidweichiang:
@lbssousa even if that test fails for you, if you can accept the result, hopefully it will then pass for everyone. Then I think this will be ready to merge.

lbssousa:
I've already [updated my gregorio-test PR](https://github.com/gregorio-project/gregorio-test/pull/430) accepting this result. Could you please check it once again?

PR #1769, opened 2026-08-09
Add center/right alignment for above-lines text

### LINKED ISSUE #1768: Feature request: add center/right alignment options for above-lines text
# Motivation

Above-lines text (`<alt>...</alt>` and the note-level `[alt:text]` field) is
always left-aligned to its anchor note/glyph: the text's left edge lines up
with the anchor's left edge, and the text extends rightward from there.
There's no way to center it over the anchor or right-align it against it.

This makes it awkward to annotate anything that should read as *ending* at a
note rather than starting there (editorial marks placed at the close of a
melisma or phrase, footnote-style callouts anchored to a cadence,
rubrics anchored to a note near the ending of the line, or simply
centering a short label over a single note instead of having it drift to the
right).

# Proposed syntax

Add optional per-occurrence horizontal alignment to the existing syntax,
plus a document-wide default:

- Extend syllable-level tag `<alt>` with alignment modifiers for that occurrence:
`<alt:l>` (left), `<alt:c>` (center), and `<alt:r>` (right)
- Extend note-level field: `[alt:`*text*`;l]` (left), `[alt:`*text*`;c]` (center),
  and `[alt:`*text*`;r]` (right)
- New macro `\gresetabovelinestextalignment{left|center|right}` to set the
  document-wide default for `<alt>...</alt>` and `[alt:`*text*`]`
(defaults to `left`); per-occurrence markers override the default.

Example:
```
A<alt:c>Alleluia</alt>(fg) B[alt:cadence;r](g!hwi) (::)
```

# Backward compatibility

`<alt>` keeps working exactly as it does today: `<alt:l>`, `<alt:c>`, and
`<alt:r>` are new tag names that were not valid gabc before, so no existing
score using `<alt>` changes its rendered output.

The note-level `[alt:`*text*`]` field is *almost* purely additive: the
document-wide default (`\gresetabovelinestextalignment`) defaults to `left`,
matching the old behavior, and the new `;l`/`;c`/`;r` suffix only takes
effect when it appears literally, unescaped, at the very end of *text*. The
one edge case is a score where *text* already ended in a literal, unescaped
`;l`, `;c`, or `;r`: that suffix is now stripped and treated as an
alignment marker instead of literal text, changing the rendered output. See
UPGRADE.md for the workaround (escaping the semicolon with `$`).

lbssousa:
Fixes #1768.

## Summary

Above-lines text (`<alt>...</alt>` and `[alt:text]`) has always been
left-aligned to its anchoring note/glyph. This adds center and right
alignment, both per-occurrence and as a document-wide default.

## Usage

- Syllable-level tag: `<alt:l>`, `<alt:c>`, `<alt:r>` instead of `<alt>`
  (closing tag is always `</alt>`, regardless of which was opened).
- Note-level field: `[alt:`*text*`;l]`, `[alt:`*text*`;c]`,
  `[alt:`*text*`;r]` instead of `[alt:`*text*`]`.
- Document-wide default: `\gresetabovelinestextalignment{left|center|right}`
  (defaults to `left`, i.e. unchanged prior behavior); per-occurrence
  markers override it.

Additive for the `<alt>` tag (`<alt:l>`/`<alt:c>`/`<alt:r>` are new tag
names, so no pre-existing gabc used them). The note-level
`[alt:`*text*`]` field is additive except for one narrow edge case: if
*text* already ended in a literal, unescaped `;l`, `;c`, or `;r`, that
suffix is now consumed as an alignment marker instead of being displayed.
UPGRADE.md documents the escape workaround (`` $; ``); see CHANGELOG.md
for the corresponding note.

## Implementation

- `gregorio_alt_alignment` enum + a shared char↔alignment mapping
  (`struct.h`), new fields on the note/syllable structs, gabc
  lexer/parser recognition of the new syntax.
- gabc round-trip (`gabc-write.c`) and debug dump (`dump.c`) support.
- `\GreSetTextAboveLines{text}` → `\GreSetTextAboveLines{alignment}{text}`
  in the generated GregorioTeX, and the macro implementation itself:
  for center/right, the text is measured once against the anchor
  glyph's already-known width and kerned into place; left keeps its
  original zero-measurement path.
- Documentation (`Gabc.tex`, `Command_Index_User.tex`) and CHANGELOG.

## Pottentialy breaking change

- Pre-existing `[alt:`*text*`]` fields where *text* already ended in a
  literal `;l`, `;c`, or `;r` will now have that suffix reinterpreted as
  an alignment marker.


davidweichiang:
I didn't understand these two points:
```
- When the anchor note is immediately followed by a bar line,
center/right alignment falls back to the previous note's width
instead of the bar's.
- On a score's literal first syllable, the drop-initial spacing
(beforeinitialshift/afterinitialshift) is applied after the
alignment kern is already computed, which can desync alignment for
that syllable specifically.
```
Can you give examples?

davidweichiang:
For the syllable-level tag, since `<alt>` is HTML-inspired, what would you think about `<alt align="center">text</alt>` instead?

For the note-level tag, what would you think about `[alt align="center":text]` which would avoid the issue with `text` ending in `;c`?

davidweichiang:
What are the use cases for this? Do you have examples?

lbssousa:
> For the syllable-level tag, since `<alt>` is HTML-inspired, what would you think about `<alt align="center">text</alt>` instead?
> 
> For the note-level tag, what would you think about `[alt align="center":text]` which would avoid the issue with `text` ending in `;c`?

In principle I'm fine with the idea of changing the syntax, but it would be unprecedented in Gregorio's development history.

The syntax chosen for the `<alt>` tag was based on the notation of the `<pr>` one, which already accepts a numeric parameter separated by `:` (e.g. `<pr:.5>` for protrusion amount). The syntax chosen for the note-level `[alt:...]` field, in turn, was based on the syntax already supported by other fields of the same nature, which accept additional parameters separated by `;`, such as `[ob:n;size]`, `[oll:left;right]`, `[oslur:shift;width,height]`, etc.

lbssousa:
> I didn't understand these two points:
> 
> ```
> - When the anchor note is immediately followed by a bar line,
> center/right alignment falls back to the previous note's width
> instead of the bar's.
> - On a score's literal first syllable, the drop-initial spacing
> (beforeinitialshift/afterinitialshift) is applied after the
> alignment kern is already computed, which can desync alignment for
> that syllable specifically.
> ```
> 
> Can you give examples?

Ah, nevermind! I jumped the gun listing them as known limitations, apparently from some poorly-conducted testing on my part. I went back and tested both scenarios much more carefully, including pixel-level measurements comparing rendered PDFs across a bunch of variants, and neither issue actually occurs. I've added regression coverage for both cases to `above-lines-text-alignment` in gregorio-test, so this stays covered going forward.

lbssousa:
> What are the use cases for this? Do you have examples?

Some practical examples for right-aligning ALT:
- labeling _differentias_ at the end of psalm tones (Ia, Ia2, Ig, If, etc.). Randomly-generated example:
image

- adjusting some rubrics written as ALT which happen to occur near the end of line (they may olverflow if kept left-aligned). Randomly-generated example:
image

davidweichiang:
> The syntax chosen for the `<alt>` tag was based on the notation of the `<pr>` one, which already accepts a numeric parameter separated by `:` (e.g. `<pr:.5>` for protrusion amount). 

Oh, okay. I would have said the same thing about `<pr:.5>`, but now that it's done, I agree we should keep the same pattern.

> The syntax chosen for the note-level `[alt:...]` field, in turn, was based on the syntax already supported by other fields of the same nature, which accept additional parameters separated by `;`, such as `[ob:n;size]`, `[oll:left;right]`, `[oslur:shift;width,height]`, etc.

That makes sense. I think I'd mildly prefer `[alt:c;text]` which makes the `c` more visible. But that doesn't solve the issue with `text` containing `;c` or `c;`. It would be ugly, but what about something like `[altc:text]`, `[alt;c:text]`, etc.?

davidweichiang:
> adjusting some rubrics written as ALT which happen to occur near the end of line

I agree right-alignment is a good workaround for this, but I'll also create a new issue about it.

lbssousa:
> That makes sense. I think I'd mildly prefer `[alt:c;text]` which makes the `c` more visible. But that doesn't solve the issue with `text` containing `;c` or `c;`. It would be ugly, but what about something like `[altc:text]`, `[alt;c:text]`, etc.?

I'm considering alternative syntaxes, but my main concern is not breaking current `[alt:text]` syntax, which should produce an ALT with the default alignment. Assuming that `text` is a mandatory parameter, but alignment control is an optional parameter for `[alt]`, I'm not fully convinced that `[alt:c;text]` makes more sense than `[alt:text;c]`. A proposal like `[alt;c:text]`, in turn, breaks the pattern followed by other note-level tags (`:` should always come first).

Maybe a better approach is not introducing alignment control as an optional parameter for `[alt]`, but instead introduce _new tags_ `[alt-l:text]`, `[alt-c:text]` and `[alt-r:text]` (which are a cleaner alternative to your `[altc:text]` proposal). We could even extend this idea to the lyric-level tag for symmetry (`<alt-l></alt-l>`, `<alt-c></alt-c>`, and `<alt-r></alt-r>`).

lbssousa:
> Maybe a better approach is not introducing alignment control as an optional parameter for `[alt]`, but instead introduce _new tags_ `[alt-l:text]`, `[alt-c:text]` and `[alt-r:text]` (which are a cleaner alternative to your `[altc:text]` proposal). We could even extend this idea to the lyric-level tag for symmetry (`<alt-l></alt-l>`, `<alt-c></alt-c>`, and `<alt-r></alt-r>`).

Following this direction, A new tag naming like `lalt` / `calt` / `ralt` (for both lyric-level and note-level tags) could be an even better choice.

davidweichiang:
Aesthetically, out of all these, I like `[alt-c]` and `<alt-c>`.

tex/gregoriotex-main.tex
```diff
@@ -692,11 +722,48 @@
...
+    \fi%
+    \raise\gre@dimen@temp@five\hbox to 0pt{%
+      \ifnum\gre@count@temp@one>1\relax%
+        \unhbox\gre@box@abovelinestexttemp%
+      \else% left: nothing was measured above, typeset directly
+        \gre@style@abovelinestext
+          \gre@applyphantomwrapper{\gre@abovelinestext@phantomwrapper}{#2}%
+        \endgre@style@abovelinestext
+      \fi%
     \hss}%
```
davidweichiang, line 740: Would it not have been simpler to generate `\hss` at the beginning and/or end of the box conditioned on the alignment setting (left = at end, right = at beginning, center = at both)?

lbssousa: I've just applied the simplifications you suggested, as well as the new tags `alt-l`, `alt-c`, and `alt-r`. The corresponding tests were also updated.

